PR #69, opened 2015-08-20
Martin

mmaechler:


mmaechler:
My changes to NAMESPACE are correct.. also importing yet another S3method is even more wrong..

Issue #27, opened 2015-08-19
Sparse matrix for design matrices 


fishfollower:
Consider the possibility to transfer sparse matrices for selected design matrices

mmaechler:
As maintainer of package `Matrix`,  I'd definitely like very much to support that

bbolker:
implemented in 9cb23d77af

PR #94, opened 2015-08-21
sparse Z matrices

bbolker:


bbolker:
partially addresses #27 (sparse Z only)

PR #147, opened 2015-11-19
fix tests on basic Gaussian Sleepdata

mebrooks:
I hope it's ok that `sigma.glmmTMB` is not internal anymore. I needed it to run one of the tests.

When I updated the namespace, the line `export(getME)` also showed up. I'm guessing this was due to changes made by someone else without rerunning `make namespace-update`


mebrooks:
I want to check that these changes are ok before I add more changes to the tests.

Issue #160, opened 2016-06-16
summary can't handle RE in ziformula


mebrooks:
Here's an example
```
data(Owls)
m0=glmmTMB(SiblingNegotiation~ 1, ziformula=~ (1|Nest), family=poisson(), Owls)
summary(m0)
```
This is the traceback
```
> traceback()
7: stop(sprintf(ngettext(length(r), "%s is not TRUE", "%s are not all TRUE"), 
       ch), call. = FALSE, domain = NA)
6: stopifnot(length(cnms) == (nc <- length(cor)), nc == length(sd), 
       is.list(cnms), is.list(cor), is.list(sd), is.character(nnms <- names(cnms)), 
       nzchar(nnms))
5: mkVC(cor = xrep$corzi, sd = xrep$sdzi, cnms = cn)
4: VarCorr.glmmTMB(object)
3: VarCorr(object)
2: summary.glmmTMB(m1)
1: summary(m1)
```

mebrooks:
I found the problem and am working on fixing it. `corzi` should have been `corrzi` on line 111 of VarrCorr.R

bbolker:
fixed yet ... ?

PR #162, opened 2016-06-20
fixed VarCor with RE in zi

mebrooks:
also added a test for dispformula


mebrooks:
issue #160

mebrooks:
It's strange that the Travis build is failing on code I didn't change. The fails have to do with having `pscl` installed and the structure of `fm0$modelInfo$reTrms`.

Issue #161, opened 2016-06-20
Weights should be removed


mebrooks:
The way that weights are included are not compatible with the Laplace approximation and are misleading. In glmmADMB, they were done in a separate step after the Laplace. It makes more sense to use the dispersion formula.

bbolker:
In my experience weights are a perennial source of confusion, because there are so many possibilities/ambiguities about how they can be defined, e.g.

- are they automatically rescaled to sum to 1? Or does increasing the sum of the weights increase the reported precision of the estimates?
- there's often a bunch of special machinery needed to deal with observations with weights of zero (if that's allowed)
- there is a distinction between *prior* and *working* weights (in an IRLS framework, observations in a GLM(M) are iteratively weighted based on their implied variances, which changes over the course of the fitting procedure
- survey weights are another can of worms altogether.

Any chance you could write up a short description of what is done/should be done, for future reference?

mebrooks:
@skaug, can you articulate what might go wrong with the Laplace approximation and weights? 

Also, I forgot to mention the caveat that we still need them for the binomial distribution. Maybe we want to throw an error if the family is not binomial and weights are not NULL.

skaug:
Currently the weights are applied "in the template", i.e. before the Laplace approximation (LA), 
but then an error occurs when you apply the LA. To see why,
consider dim=1, and apply LA to weight*logL, the log(-Hessian) becomes 
log(weight) + log(-logLL''), while it should have been weight*log(-logLL'') 
for a correctly weighted marginal likelihood.

I do not think there is an easy fix for this in TMB.

mebrooks:
In the future, reconsider ways to implement weights.

bbolker:
We've had another request for this, so I'd like to re-open it.  Can we think of a way to restructure things in some not-too-painful way that would allow weights to be handled properly in LA?

mebrooks:
I'm hoping a breakout group can discuss this at the TMB developer's meeting in August. If we solve the problem sooner, that's great, but I have my hands pretty full through June at least.

mebrooks:
Sadly, we didn't get to this in the developer's meeting. It's still an open question.

mebrooks:
Weights were added back by #272

PR #188, opened 2016-11-15
Only allow weights for binomial and betabinomial

mebrooks:
I removed weights from multiplying the log-likelihoods in glmmTMB.cpp. Here is Han's explanation for why: #161. I also added a check in glmmTMB.R so that it stops if weights are attempted with an inappropriate family. I followed the examples of other checks done on families (using a function that checks for matches to a vector of family names) even though I wasn't sure why they were written in that way...I'm sure there are special cases.


bbolker:
I think this is fine - it's definitely better to have no weights than incorrect weights - but at some point we might want to consider re-implementing weights for other families, in the right way.  There are, unfortunately, lots of different ways to implement weights corresponding to different meanings of the weights ... e.g. https://github.com/lme4/lme4/issues/155

PR #198, opened 2016-12-01
Printdispersion

mebrooks:
It always seemed weird to me to call the overdispersion parameter "sigma" for all families, so I removed that from the summary output. I found the explanation of the `sigma()` function to be very helpful for interpreting overdispersion parameters. So I point to it in the details of the description of `dispformula` in the glmmTMB helpfile (someone might want to read it and make sure it makes sense). I wonder if it would be clearer (it would for me) if `sigma()` was renamed `dispersion()`. 

Soon, I also want to edit the examples in the helpfile to include nbinom and ZI since those features will be the attractor for many new users. Maybe we could include examples of the RE structures with simulated data, to give an idea of what the RE structures mean.

### DIFF
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -666,9 +666,9 @@ print.summary.glmmTMB <- function(x, digits = max(3, getOption("digits") - 3),
             .prt.grps(x$ngrps[[nn]],nobs=x$nobs)
         }
     }
-
-    printDispersion(x$family,x$sigma)
-
+    if(length(x$sigma)==1) { #only print here if not printing dispersion model
+        printDispersion(x$family,x$sigma)
+    }
     for (nn in names(x$coefficients)) {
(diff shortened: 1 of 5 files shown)


mmaechler:
Well,  `sigma()` has become a generic function in R ...  to be used even in cases such as this where the greek sigma() is not used typically.   OTOH,  most non-users of glm  who may know quite a bit of applied statistics don't know the term  "dispersion".
So it may have been a suboptimal choice to choose the name sigma(), ... and I am the culprit ...
I would be considerably rustrated if you'd not use it here...

mebrooks:
I'm ok with sticking to convention for the function name.  It was just a thought. I still want to stop outputting "sigma" with the overdispersion parameter. 

It turns out that I edited the helpfile in the wrong place (post Roxygen), so I'll have to make that edit again and do another pull request.

mmaechler:
((maybe you understand my roxygen view on producing *.Rd a tiny bit better now ;-))

PR #199, opened 2016-12-01
Covstruct vignette

kaskr:


mebrooks:
Eventually we also want to add small symbolic representations of the matrix structures.

Issue #201, opened 2016-12-02
Convergence diagnostics


kaskr:
We should somehow inform the user when a model has not converged. This raises two questions:

- How to decide if a model has converged. Simply checking whether `sdr$pdHess` is `TRUE` is usually sufficient.
- How to report it. As a warning after running  `glmmTMB` ? As part of the printed output from `print` or `summary`?

bbolker:
`lme4`-based advice: skipping point 1 (which seems sensible, and what we've done in `lme4` has been messy and ugly and not necessarily a bad idea - ask me sometime), I think there should be *both* a warning *and* something included in the output when the model is printed or summarized.  There's a `.prt.warn` function in `lme4` that we probably don't want to copy exactly, but which could be used as an example.

mebrooks:
In addition to a warning at the end of fitting, it would make my life easier if `AICtab` found out about convergence problems somehow and didn't include those. It's happened to me twice already that there are 1 or 2 models at the top of the AIC table that I see are problematic when I look at the `summary` and see `NA`s. Is there a good way to prevent unconverged models from ending up at the top of the AIC table? Just thinking here...if `sdr$pdHess == FALSE`, then will the `logLik` ever contain useful information, or can it return `NA`?

mebrooks:
On the topic of warnings...would it make sense to suppress the warnings like 
```
Warning messages:
1: In nlminb(start = par, objective = fn, gradient = gr) :
  NA/NaN function evaluation
2: In nlminb(start = par, objective = fn, gradient = gr) :
  NA/NaN function evaluation
3: In nlminb(start = par, objective = fn, gradient = gr) :
  NA/NaN function evaluation
```
that usually mean the optimizer wandered into a bad region for a while, but may have gotten back out. As long as it's back in a good region by the end of optimization, then do these warnings serve a purpose?

bbolker:
I generally err on the side of *not* sweeping warnings under the rug.  What we try to do in `lme4` is to catch the warnings and store them somewhere, then report that "warnings occurred" - it's enough to know that at least one NA/NaN evaluation occurred, one doesn't need to see 45 of them.  So, in order, my preferences are:

- try to make the fitting process more robust (e.g. via @kaskr's recent modifications) so that we can stay out of areas where the warnings occur;
- collect warnings and let the user know in a streamlined way;
- spit out all the warnings (current approach);
- convince ourselves that these warnings are definitely ALWAYS false positives and suppress them

[...]

PR #202, opened 2016-12-08
logLik=NA when pdHess=FALSE

mebrooks:


mebrooks:
I think I fixed it now so that it should work with `glmmTMB(..., se=FALSE)`. However, now it's failing on Travis (RVERSION=devel) with 
```
* checking package dependencies ... ERROR
Packages required but not available: ‘TMB’ ‘lme4’ ‘RcppEigen’
Packages suggested but not available:
  ‘knitr’ ‘ggplot2’ ‘mlmRev’ ‘bbmle’ ‘pscl’
```

mebrooks:
@mmaechler, if the Travis check is only failing on `RVERSION=devel`, is it safe to assume the problem is coming from the development version of R?

mmaechler:
Well, "only" is maybe an understatement:  That version usually works for all CRAN packages (that the released version works) and it is *the* plan for the next version of R...

However, I do think it will be a Travis problem where for some reason "R devel" version is partly broken, because I cannot see why  lme4 or RcppEigen  should not be available there... but then I usually don't use Travis but test on my computers .. and there lme4 does (almost always) work with R-devel.

So while I'd like to defend the quality of R-devel (which is usually also the *fastest* version of R), I agree that a failure like the above should not mean a real problem with our package.

mebrooks:
Ok great! Thanks. I'm going to merge this in then. This should take care of #201

bbolker:
I would agree that (in my experience) when the Travis build fails on only one out of three R versions, it's *usually* a Travis glitch rather than a real cross-version incompatibility.

PR #203, opened 2016-12-14
Simulate from random effects

kaskr:
This is a continuation of branch 'simulate_truncated' that now also simulates 'b' and 'bzi' vectors.
Truncated simulators are correct as far as I can tell - still need testing.

### DIFF
--- a/glmmTMB/R/methods.R
+++ b/glmmTMB/R/methods.R
@@ -626,11 +626,10 @@ fitted.glmmTMB <- function(object, ...) {
     predict(object)
 }
 
-.noSimFamilies <- c("beta", "betabinomial", "truncated_poisson", 
-"truncated_nbinom1", "truncated_nbinom2")
+.noSimFamilies <- c("beta", "betabinomial")
 
 noSim <- function(x) {
-	!is.na(match(x, .noSimFamilies))
+    !is.na(match(x, .noSimFamilies))
 }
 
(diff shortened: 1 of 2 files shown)


mebrooks:
I've been testing the truncated simulations and they aren't quite working. There's something strange going on with `Rf_qpois`. (I haven't tried the nbinom versions yet since I couldn't get the Poisson working.) Without a seed, I get all `NaN` values. With a seed, I get 1 integer and the rest `NaN`. I checked the values coming from `runif` and they're fine, so I think it's something weird with `Rf_qpois`.  I've tried writing the `extern "C" {}` code in a few different ways, but it doesn't help. Do you have any ideas? 

This is the code I was using to test it.
```
library(glmmTMB)
library(ggplot2)
mu=3
n=100
y0=rpois(n, mu)
pdat=data.frame(y=y0[y0>0]) 
pdat$type="observed"
m1=glmmTMB(y~1, family=list(family="truncated_poisson",link="log"), pdat)
qpois(runif(n, dpois(0, exp(fixef(m1)$cond)), 1), exp(fixef(m1)$cond))
s1=simulate(m1, seed=1)[[1]]
s1
pdatsim=data.frame(y=s1, type="simulated")
pboth=rbind(pdat, pdatsim)
ggplot(pboth)+ geom_density(aes(x=y, colour=type))
```

kaskr:
Now fixed.

PR #210, opened 2017-01-30
Genpois

mebrooks:
generalized Poisson with parameterization described in Consul & Famoye (1992) Generalized poisson regression model, Communications in Statistics - Theory and Methods, 21:1, 89-109, DOI: 10.1080/03610929208830766

Conway-Maxwel Poisson with new parameterization by @kaskr in TMB that uses the true mean rather than an approximated mean.


mebrooks:
I'm getting the package ready to pass CRAN checks, so I'm updating the `Date` and `Suggests` in the `DESCRIPTION`. @kaskr, should I go ahead and update `Imports: TMB (>= 1.7.6)`

PR #216, opened 2017-02-13
Mollie

mebrooks:
- patch increment version number
- hurdle example
- simplify dispersion model example that didn't converge


bbolker:
It would be a good idea (IMO) to save the non-converging version of the dispersion model example somewhere we can find it.  It's *extremely* useful for future diagnostics/robustification etc. to have well-documented examples of things that *don't* work ...

PR #219, opened 2017-02-15
Mollie

mebrooks:
- initial troubleshooting vignette (needs work)
- warning refers to troubleshooting vignette
- warn when eigen values are small or NA
- stop for zeros in trunc dist without zi

Passes `check --as-cran`. 
Only tested new warning on troubleshooting vignette examples; more examples needed.


mebrooks:
I'm making more edits to the vignette.

PR #230, opened 2017-03-13
Families

### LINKED ISSUE #225: remove option of family as a list?
Related to #224 and Pearson residuals, I wonder if it would be easiest if we always specify family as a function or character referring to a function. Then the variance and anything necessary for residuals and initialization can be connected to the family function.  We would need to clean up these lines of code in the `glmmTMB` function. 
```
    if (is.character(family))
      family <- get(family, mode = "function", envir = parent.frame())
    if (is.function(family))
      family <- family()
    if (is.null(family$family)) {
      print(family)
      stop("'family' not recognized")
    }
    if (!all(c("family","link") %in% names(family)))
        stop("'family' must contain at least 'family' and 'link' components")
```
I was pleasantly surprised to find that when the family is specified as a function, then the identity link works with the Poisson and nbinom distributions (code below). I had been trying to specify it in the list format before and it didn't work.

```
library(glmmTMB)
library(MASS) #for glm.nb
x=1:100
m=2
b=100
y=m*x+b
#nbinom2
dat1=data.frame(obs=rnbinom(length(y), mu=y, size=5), x=x)
with(dat1, plot(x, obs))
summary(glm.nb(obs~x, link="identity", dat1)) #this works
summary(glmmTMB(obs~x, family=nbinom2(link="identity"), dat1)) 

#Poisson
dat=data.frame(obs=rpois(length(y), y), x=x)
with(dat, plot(x, obs))
summary(glm(obs~x, family=poisson(link="identity"), dat)) #this works
summary(glmmTMB(obs~x, family=poisson(link="identity"), dat))
```

### LINKED ISSUE #229: No Pearson residuals for models with dispersion models or zero-inflation
I'm working on improving `families.R` on the `families` branch. I added a check so that we don't produce Pearson residuals in models with zero-inflation. Should I also do this for models with dispersion models? Right now `residuals.glmmTMB` calls `sigma.glmmTMB`, which is only designed to work with a single value in `betad`. It should be possible to calculate Pearson residuals for non-standard models, but it will require overhauling `residuals.glmmTMB`.

mebrooks:
- Failed tests were a Travis problem. 
- I have tested the variance functions using large simulated datasets and checking that the mean of squared pearson residuals is approximately 1. I did not add these tests to unit testing.
- We still need the variance for truncated nbinom distributions. If anyone finds a reference for these, please let me know.
- Fixes #229 
- Fixes #225 by de-emphasizing `family=list`


mebrooks:
To update the namespace in a way that passes CRAN checks, I had to use Roxygen2 version 5.0.1. Something changed with version 6.0 and it exports some of the S3methods that CRAN then expects to be documented (e.g. `print.VarCorr.glmmTMB`). The new version replaces `S3method(print,VarCorr.glmmTMB)` with `export(print.VarCorr.glmmTMB)`. It looks like this function is documented, so I don't know what the problem is. Eventually, we'll need to figure out a better solution, but for now, installing the old version of Roxygen2 works with this code 
```
devtools::install_version("roxygen2", version = "5.0.1", repos = "http://cran.us.r-project.org")
```

Issue #196, opened 2016-11-30
Printing complex/large variance-covariance matrices


kaskr:
Try this:

```
library(glmmTMB)
example(glmmTMB)
sleepstudy$row <- factor(1:180)
fm <- glmmTMB(Reaction ~ ar1(row + 0 | Subject), sleepstudy)
```

Printing the model takes a very long time and prints too much (and for some reason emacs doesn't seem to like scrolling the output) :

```
system.time(print(fm))

   user  system elapsed 
  0.064   0.064  14.796 
```

The problem is `glmmTMB:::print.VarCorr.glmmTMB`. When the corr matrices do not fit the display I'd prefer by default either

1. Printing just the head cols/rows **or**
2. Printing both head and tail cols/rows

Perhaps this is just a matter of passing appropriate arguments to `lme4:::formatVC`...
Any thoughts ?

bbolker:
In the console (linux) it's not so bad -- 0.144 user time, 0.472 elapsed.  (But it took emacs 70 seconds to print on my system.) Regardless of the timing, though, the print format is ridiculous. I think complex variance-covariance matrices should know something about their own structure and how to print themselves -- this was done nicely in nlme, which had a range of matrix types with their own print methods.

mmaechler:
I agree 
- that the current print method is ridiculous as soon as the matrices are non small
- ideally should change this in lme4:::formatVC or another "place" that's common to the two packages.
  IIRC, we don't by default print the VC matrix in summary(<lme4>), do we?

bbolker:
I pushed a preliminary version of this at  cf9237d . I copied `formatVC` from `lme4` and hacked it (but it's backward compatible; if we do want to share it I would propose we eventually put it back in `lme4` and export it from there). I'm not sure about the structure/design: ideally it would make sense for the *individual* components of variance-covariance objects to have classes ... for now, though, I just wanted to get something working.  Right now it looks like this (use the `print_fancy_vcov` branch ...)

```
library(glmmTMB)
data(sleepstudy,package="lme4")
sleepstudy$row <- factor(1:180)
fm <- glmmTMB(Reaction ~ ar1(row + 0 | Subject), sleepstudy)
VarCorr(fm)
## Conditional model:
##   Groups   Name Std.Dev.   Corr       
##   Subject  row1 6.1192e+01 0.853 (ar1)
##  Residual          3.4087e-07     
```

Comments welcome.

bbolker:
[...]

PR #236, opened 2017-04-21
Getting ready to submit new version to CRAN

mebrooks:
- bbmle 1.0.19 is on CRAN and handles unconverged models better than before
- Salamander site names had inconsistent spacing in original version
- passes check as-cran


bbolker:
I would be slightly conservative about submitting new versions to CRAN ... from [CRAN policy](https://cran.r-project.org/web/packages/policies.html)

> Submitting updates should be done responsibly and with respect for the volunteers’ time. Once a package is established (which may take several rounds), “no more than every 1–2 months” seems appropriate. 

Hence there is an opportunity cost to submitting a version, because it blocks us from submitting versions with further corrections for the next while.  I wouldn't mind getting the stuff from the `print_fancy_vcov` branch (see #196) incorporated in the next CRAN version ...

mebrooks:
That makes sense. I mainly want to submit a new version because of all the functions I added to `family.R` so that none of the families need to be specified as a list anymore. I used this feature in the manuscript because it looks simpler, so we should probably get it on CRAN by the time the manuscript goes for review.

PR #239, opened 2017-05-14
Tweedie

kaskr:
TODO: Make print method display the power parameter `plogis(object$fit$par["thetaf"]) + 1`.


bbolker:
this is very exciting!

PR #244, opened 2017-05-31
Kasper

kaskr:
These simple changes enables (on request) automatic generation of OSA residuals and accuracy check of the Laplace approximation.


mebrooks:
Is the purpose of reporting the simulated random effects to let us calculate what the simulated observations would be conditional on the estimated random effects?

PR #245, opened 2017-06-01
Print Tweedie power parameter

kaskr:
When running the tweedie example in `glmmTMB/tests/testthat/test-families.R` the power parameter is now displayed:

```
> twm
Formula:          y ~ 1
      AIC       BIC    logLik  df.resid 
10064.246 10081.049 -5029.123      1997 

Number of obs: 2000

Overdispersion parameter for tweedie family (): 2.04 

Tweedie power parameter:  1.7 

Fixed Effects:

Conditional model:
(Intercept)  
      1.405  
```

Not sure this is the 'right way'. Please check carefully and modify :)

### DIFF
--- a/glmmTMB/R/methods.R
+++ b/glmmTMB/R/methods.R
@@ -398,6 +398,19 @@ printDispersion <- function(ff,s) {
     NULL
 }
 
+## Print family specific parameters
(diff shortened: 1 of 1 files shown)


bbolker:
seems fine to me?

Issue #253, opened 2017-06-30
control for interbally call nlminb from glmmTMB


yukio-takeuchi:
Currently nlminb function called internally from glmmTMB is called with no default control parameters.
Fro example , maximum number of iteration is fixed at default value of 150.  It would be nice, if the function glmmTMB has an additional argument "control" to pass nlminb to change control parameters of  nlminb.

bbolker:
note this is available (since the branch referenced above was merged), via e.g. 

```
control = glmmTMBControl(optCtrl=list(iter.max=1e3,eval.max=1e3))
```

PR #254, opened 2017-07-28
Optimizer

kaskr:
Some comments on this PR:

* A new option to profile the beta parameters as part of the inner
  problem has been added. It is generally faster and more stable than
  the current approach in cases with many fixed effects.
* The argument is currently set to FALSE by default. Eventually we may
  want to enable it based on e.g. the number of beta parameters in the
  model. However, this will require much more testing.
  As a side note, setting profile to TRUE fails a few travis tests
  only because the new method gets the gradient closer to zero.
* The default nlminb control parameters results in "iteration limit
  reached without convergence (10)" too often. A new argument
  optControl is now added and we may want to increase the default
  maximum number of iterations. This also addresses #253.
* A warning message is now thrown if nlminb has non-zero convergence
  code.


kaskr:
Try a motivating example from
 https://github.com/bbolker/mixedmodels-misc/blob/master/notes/cloglogsim.rmd

* Fit with default nlminb control parameters stops too early. we now get a warning:

```r
system.time(fit1 <- glmmTMB(Dead/(Alive+Dead) ~ (Trt + 0)/x + (x | Rep),
                            weights = Alive+Dead,
                            family=binomial(link="cloglog"),data=X))
logLik(fit1)
range(fit1$sdr$gradient)
```

```
   user  system elapsed 
 10.624   0.000  10.621 
Warning message:
In fitTMB(TMBStruc) :
  Model convergence problem; iteration limit reached without convergence (10)
> logLik(fit1)
'log Lik.' -1415.977 (df=51)
> range(fit1$sdr$gradient) ## Not converged
[1] -39.99205  49.44483
```

* We can fix by increasing the number of iterations. However, it becomes slow.

```r
system.time(fit2 <- glmmTMB(Dead/(Alive+Dead) ~ (Trt + 0)/x + (x | Rep),
                            weights = Alive+Dead,
                            family=binomial(link="cloglog"),data=X,
                            optControl=list(iter.max=1e3,eval.max=1e3)))
logLik(fit2)
range(fit2$sdr$gradient)
```

```
   user  system elapsed 
 14.416   0.004  14.417 
> logLik(fit2)
'log Lik.' -1413.115 (df=51)
> range(fit2$sdr$gradient)
[1] -0.006779599  0.008587737
```

* OR we can set profile=TRUE

```r
system.time(fit3 <- glmmTMB(Dead/(Alive+Dead) ~ (Trt + 0)/x + (x | Rep),
                            weights = Alive+Dead,
                            family=binomial(link="cloglog"),data=X,
                            profile=TRUE))
logLik(fit3)
range(fit3$sdr$gradient)
logLik(fit3) > logLik(fit2)
```

```
   user  system elapsed 
  2.360   0.000   2.356 
> logLik(fit3)
'log Lik.' -1413.115 (df=51)
> range(fit3$sdr$gradient)
[1] -7.575120e-08  2.178636e-07
> logLik(fit3) > logLik(fit2)
[1] TRUE
```

* Compare results of the two valid fits

```
> diff <- (summary(fit3)$coefficients$cond - summary(fit2)$coefficients$cond)[,1:2]
> range(diff[,"Estimate"])
[1] -1.457380e-05  1.174259e-05
> range(diff[,"Std. Error"])
[1] -3.401091e-04  1.793294e-05
```

mebrooks:
This sounds like good progress to me, but maybe the lme4 developers will have some input based on experience.

kaskr:
I'm starting to have some second thoughts...

I think it might be better to follow `lme4::glmer` more closely and
add a single `control` argument to glmmTMB along with a function
`glmmTMBControl()`. To begin with the default would be a list
```
list(optCtrl=list(), profile=FALSE)
```
but later on, if we wanted adaptive control parameters, we could
change that to
```
list(optCtrl=list(), profile=quote(length(beta)>2))
```
and evaluate each list element inside `TMBStruc`.

It would be natural to document the optimization strategy under
`?glmmTMBControl`.

bbolker:
(1) I'm not sure you need `quote()` there, as R does lazy evaluation anyway.  Arguments don't get evaluated until they're referred to (however far down the call stack that is). (2) however, I wouldn't use anything fancier than `length(beta)>2` as a default argument. Using lazy-evaluated expressions has the advantage of being self-documenting, but it can also be confusing. (3) I agree that arguments to be passed on to the optimizer should probably be segregated as part of `optCtrl()`.

kaskr:
I agree it's probably better to take advantage of lazy evaluation
although that would prevent the *user* from experimenting with
adaptive parameters:

```
> con <- glmmTMBControl( profile = (length(parameters$beta)>2) )
Error in namedList(optCtrl, profile) : object 'parameters' not found
```

Also note that we will no longer be able to set adaptive defaults in
`glmmTMBControl`. They will have to be defined in `glmmTMB` using
something like
`glmmTMB(control=glmTMBControl(profile=(parameters$beta>2)))`.

In any case, I don't think `profile` should be enabled by default yet.

@bbolker what's your opinion on changing the default max number of
iterations ?

Current (implicit) default is
```
glmmTMBControl( optCtrl = list(iter.max=150, eval.max=200) )
```
I'd suggest changing that to
```
glmmTMBControl( optCtrl = list(iter.max=300, eval.max=400) )
```
based on the cloglog example.

bbolker:
I think there's little downside to increasing the max number of evaluations - it will only be used if it's needed ...

PR #260, opened 2017-10-23
Binomial

### LINKED ISSUE #259: Parameterization of binomial model.
Specifying binomial models with (observed count x out of N) in glmmTMB is currently done by setting the response variable to the fraction x/N and including N as a weight argument. This should be reconsidered for two reasons. 

The simulate() method returns integer counts, which seems a bit inconsistent. 

The one-obs-ahead residual calculation is difficult because the discrete support of the response variable is different for each different N.   

Should we make it such that the response is specified with a cbind(x,N-x), and update the simulate method? Most importantly the internal part should work such that the binomial density is called with x as it argument).

kaskr:
Fix #259

### DIFF
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -389,7 +389,7 @@ okWeights <- function(x) {
 ##' @details
 ##' \itemize{
 ##' \item binomial models with more than one trial (i.e., not binary/Bernoulli)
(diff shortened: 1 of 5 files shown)


mebrooks:
Should we disable/hide the weights argument completely (in a way that's easy to reverse) since it was only used to fill this role in the binomial family? It could prevent users from expecting the weights argument to fill its more general role.

PR #262, opened 2017-10-24
Betabinomial

kaskr:
Continuing #260 and #259 with similar changes for the betabinomial family.


mebrooks:
`weights` comment above still to be discussed and possibly addressed in another commit/pull request

PR #272, opened 2017-11-09
Weights

mebrooks:
I added a test that makes sure a model with aggregated data and weights is identical to its non-aggregated version. Tests of binomial models that had proportions as the response were switched to the `cbind(success, failure)` specification even though the proportion version still works.


mebrooks:
More tests may need to be added in the future (e.g., models with zero-inflation and/or dispersion components).

glmmTMB/tests/testthat/test-families.R
```diff
@@ -33,13 +33,6 @@ test_that("binomial", {
     expect_equal( logLik(mod4)     , logLik(mod5) )
     expect_equal( fixef(mod4)$cond , fixef(mod5)$cond )
 
-    ## Now with extra weights
-    dd$w <- 2
-    mod6 <- glmmTMB(cbind(success,failure)~1,family=binomial,data=dd,weights=w)
-    mod7 <- glmmTMB(prop~1,weights=size*w,family=binomial,data=dd)
```
kaskr: @mebrooks  Could we keep this test ? We still allow prop/weights right ?

glmmTMB/R/glmmTMB.R
```diff
@@ -78,12 +80,36 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
...
+        size <- rep(1, nobs)
+      } else { #proportions
+          yobs <- weights * yobs
+          size <- weights
+          weights <- rep(1, nobs)
+        }
+      }
+    }
   }
+  if (is.null(size)) size <- rep(1, nobs)
```
kaskr: Better to pass `numeric(0)` as default `size`. That'll ensure it's not used somewhere by accident.

glmmTMB/tests/testthat/test-families.R
kaskr: In that case I'd suggest modifying the test to check the likelihood of each of the two fits against `glm`.

mebrooks: @kaskr I put back the check of the `fixef` but removed the `logLik` check.

mebrooks: @kaskr Sorry, I didn't see the suggestion to compare to `glm`. I just made *another* commit that does this.

PR #277, opened 2017-12-08
Bolker ar1fix

bbolker:


mebrooks:
Thanks for fixing that Ben. I replied to Jarrod Hadfield this morning to tell him it was a bug that is fixed now.

PR #296, opened 2018-02-05
Bolker offsets

bbolker:
I'm slightly hesitant to merge this, as I think it could probably use more thorough documentation and testing. On the other hand, it passes all of the original tests, plus a simple test of offsets on dispersion parameters. I'm having a bit of a hard time thinking of sensible use cases for offsets on the zero-inflation term, but it's allowed (not tested yet though).  On balance I think it's better to get it into the master branch where it won't get forgotten about ...


bbolker:
oops, I have noticed one way in which I have made things more fragile: while trying to extract info potentially needed for evaluating offset terms, I have used all.vars() on the formula. This gets confused if we have something like `dd$var` in the formula (which should be unnecessary, but ought to be allowed). I'll see what I can do to fix this.

bbolker:
OK, I think I'm going to need to redesign this ...

bbolker:
Redesigned; it's a little simpler and should be less fragile now. I do [use `deparse()` in one place](https://github.com/glmmTMB/glmmTMB/blob/bolker_offsets/glmmTMB/R/glmmTMB.R#L204), but it really does feel like the best implementation. @mmaechler , if you have alternative suggestions that can avoid `deparse()` I'm happy to hear them ...)

PR #298, opened 2018-02-20
Update CITATION with official reference

kaskr:


bbolker:
I think this fails because the glmmTMB vignette calls `citation()` in an as-is chunk, and the Rnw file doesn't have a definition for \bold{} ... e.g. see https://github.com/glmmTMB/glmmTMB/commit/42c3785cab2ed6d57ee75191ed8ea68c4148268c

Issue #299, opened 2018-02-23
How to get NA values from predict(glmmTMBmod, newdata = testdata)?


willtudorevans:
Is there a way to get the `predict` function to produce NA values for glmmTMB models when it doesn't have the necessary data? The unequal row lengths are causing me issues in examples such as: 

```
library(dplyr)
library(glmmTMB)

n <- 20000
dat   <- data_frame(x = abs(rnorm(n, 10, 3)),
                    y =  rnorm(n, x, 5))
sample <- sample(nrow(dat), floor(.6*nrow(dat)), replace = F)
traindat <- dat[sample,]
testdat <- dat[-sample,]
mod <- glmmTMB(y ~ x, data = traindat)

#Enter one row of incomplete Data
testdat[nrow(testdat) + 1,] <- NA
testdat$predict <- predict(mod, newdata = testdat)
#Error due to row lengths not matching
```

As far as I can tell, `predict` with other model types seem to produce an NA row for rows where it can't make a prediction. 
I appreciate I could do something like:

`testdat <- testdat[complete.cases(testdat),]
`
before I then run `predict`, but in this situation pruning the dataset beforehand is not practical. 

Many thanks for any help.

bbolker:
There's a FIXME in the code:

>   ## FIXME: deal with napredict stuff ...

which I'll try to get to soon now that it's been requested ...

bbolker:
this should be resolved by cea1684fc8e5e376a5fc (if you specify `na.action=na.exclude` when running your model) ...

bbolker:
Wait a second, I didn't look at your test case carefully enough. I fixed a different case.

bbolker:
this should work (in the development version) once the last pull request has been merged ...

PR #301, opened 2018-03-02
Napredict

bbolker:


bbolker:
I don't have a precise use case in mind at the moment, but I believe all of this was done to address [this issue](https://github.com/glmmTMB/glmmTMB/issues/299).  I would vote for throwing an error only if there are NAs **and** se.fit=TRUE (and in the long run fixing the problem so that we don't crash `sdreport` ...)

mebrooks:
That seems reasonable. I made the change so it only catches cases when `se.fit=TRUE`.

PR #305, opened 2018-03-14
Fix predict

### LINKED ISSUE #304: Bug in predict method
Prediction with `dispformula=~0` and `family=gaussian()` doesn't work and I'm not really sure what's the right way to fix it.

```r
y <- 1:10
f <- glmmTMB(y ~ 1, disp=~0)
predict(f) ## Error
```

Problem is that `mkTMBStruc` handles this case by changing the dispformula to `~1` and setting the `map` argument.
The `predict` method should construct its `TMBStruc` using the *original* dispersion formula - not the altered version:

```r
f$modelInfo$allForm$dispformula <- ~0 ## Fix
predict(f) ## Works
```

I guess we need to decide whether `allForm` should contain *original* or *altered* formulas.

kaskr:
Fix #304

### DIFF
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -38,6 +38,7 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
 
     ## Handle ~0 dispersion for gaussian family.
     mapArg <- NULL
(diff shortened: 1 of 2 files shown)


bbolker:
seems harmless (famous last words ...)

PR #311, opened 2018-03-20
Profile

bbolker:


bbolker:
oops, will look at target levels more carefully.

Also: at some point in the future I can imagine `tmbroot` moving into `TMB`? It shares a lot of code with `tmbprofile`/could probably be refactored ...

bbolker:
this was good - I had a bug in the earlier profile conf int levels (I was using levels appropriate for sqrt(-2*L), but was computing sqrt(-L) (!!!)

kaskr:
It seems to be working now.

A few minor TODOs we should not forget:

Output consistency between methods E.g. look at order and rownames here:

```
example(glmmTMB)
confint(m1,method="wald")
confint(m1,method="uniroot")
```

PR #317, opened 2018-04-05
export tmbroot

bbolker:
I found myself wanting to use `tmbroot` in another package (I started to write some [tidying methods for TMB objects](https://github.com/bbolker/broom.mixed/blob/master/R/TMB_tidiers.R), so I'd like to export it from *somewhere*.  If you think you'd like it in TMB instead, I could submit a pull request there ...


kaskr:
I think it would fit well into TMB - feel free to send a pull request! My only concern is the version issues it may cause: If you move `tmbroot` from glmmTMB to TMB our travis tests will fail until a new TMB is on CRAN... Which will have to wait for a while since I recently submitted to CRAN...  It's up to you where you want to put it.

bbolker:
Good point. I don't think there's any harm to having it in both places temporarily.

The other issue with TMB incorporation is that it shares a lot of code with `tmbprofile` -- should probably be refactored to remove duplication ...

PR #319, opened 2018-04-13
Truncated compois and genpois

mebrooks:
- added `truncated_compois` and `truncated_genpois` family capabilities and tests
- cpp random number generators `rtweedie`, `rgenpois`, `rtruncated_genpois`, and `rtruncated_compois2` avoid converting parameters to `double` for brevity and consistency 
- added a test for `rtweedie` used by simulate


mebrooks:
I just realized that I should have tested the two `rtruncated_` methods with low means. I will do that soon, but I'm too tired now.

Issue #324, opened 2018-04-28
Standard Errors in predict()


strengejacke:
I'm referring to the example on p.7 of this paper: https://www.biorxiv.org/content/biorxiv/suppl/2017/05/01/132753.DC1/132753-2.pdf

The confidence intervals for the predictions are fit +/- 1.98 * se.fit.

However, I had some cases where the interval exceeds the plausible range, e.g. I had a model with binary outcome: `mod <- glmmTMB(Risk ~ Landscape + veg + (1 | Site) , data=data, family=binomial)`. What `predict()` returns are the predicted probabilities. One of the CI's, however, had a value of 103%. Is this due to wrong computation of `predict()`, or due to low precision when I use `fit - stats::qnorm(.975) * se.fit`?

I'm afraid I can't provide the data for a reproducible example (this issue was reported by a user of one of my packages), but maybe I can ask. Here's a figure of the plot created with my *ggeffects* package, where you see that one CI is above 100%.

![rplots](https://user-images.githubusercontent.com/26301769/39392770-2de5d0a8-4abc-11e8-80d3-5da4b4c63baa.png)

Here's the plot from the same data, fitted with `lme4::glmer()`

![rplot](https://user-images.githubusercontent.com/26301769/39392853-d7ae1fa4-4abd-11e8-8133-11b7d16b3395.png)

The confidence intervals are smaller, because I don't take the uncertainty of random effects into account (I'm not sure how to achieve this when I use the `newdata`-argument). May this be a reason for the larger - and even out of bounds - confidence intervals in _glmmTMB_, because it "adds" the uncertainty from random effects?

bbolker:
Could be a problem with SEs (I'll look into it), but more generally it could be that for binomial models  `p +/- 1.96*se.fit` is not guaranteed to be within [0,1] anyway; the more precise way to do the calculation is `plogis(logit.p +/- 1.96*se.fit.logit)` (i.e. compute predictions on the logit scale, find the CIs, and back-transform). (Beyond this one could think about likelihood profile confidence intervals on predictions, but people rarely bother ...)

strengejacke:
The other example was a negative binomial model, where the CI were partly negative for the count variable.

I'm not sure, but I think that `predict()` for `glmmTMB`-models always returns the "transformed" predictions (probabilities, incidents...), not the untransformed linear scale (log-odds...)? The documentation is very sparse, as fas as I have seen, and so I relied on the code from your paper I linked to in my post above.

[...]

PR #327, opened 2018-04-28
New se fix

bbolker:
This branch was written while investigating #324. In the process of investigating, I (1) added a `type` argument to `predict.glmmTMB`; (2) **removed the zitype argument** (it's now handled as part of `type`); (3) **changed the default predict() to return type="link"** for consistency with `glm()`; (4) added link-function info to all (?) of the `glmmTMB` family functions. Changes 2 and 3 are not backward-compatible, but I think they're a good idea nonetheless.


strengejacke:
The API-change makes sense to me. Looks like you can now easily handle glmmTMB objects in the same way as glm-objects, plus the zero-inflation options.

kaskr:
Not sure what are the consequences of changing the predict defaults. Here are some more to consider:

* Code snippets of the paper https://journal.r-project.org/archive/2017/RJ-2017-066/RJ-2017-066.pdf could be affected.
* Vignette `covstruct.rmd` is probably affected.

bbolker:
I think we do now have `$linkinv`, etc. components in the families ... so `family(fitted_model)$linkinv` should get the inverse link function.  Let me know if I missed something.

bbolker:
fixed covstruct vignette.  Should we add a warning (for now) whenever `predict()` is called with `type` not explicitly specified?

strengejacke:
You could also deprecate `zitype` first, so it would not radically break backwards compatibility, and simply pass`zitype` over to `type`.

strengejacke:
(or did you mean exactly that?)

bbolker:
No, I agree that deprecating (rather than eliminating) `zitype` is a better idea.

bbolker:
Bump (for those watching this PR). I have in mind some logic like:

```
if (missing(type)) {
   warning("default predict type is now 'link'; use type='response' for back-compatible behaviour or type='link' to avoid this warning")
}
```

We could then leave that warning in place for one or two releases ... ? Alternatively/in addition we could set up an `options(glmmTMB_response_type)` to allow users to adjust global settings ... ?

bbolker:
What did people think about the advisability of always warning users when `type` is not explicitly specified?

kaskr:
It'll get too annoying on the long run with the warning. As long as the new behaviour is documented in vignettes and highlighted in the NEWS I think we're good.

koenvanbenthem:
It'll get annoying, but I think a warning would be good. Now everybody may read the news or vignettes thoroughly when installing a new version.

Would it be possible to create a global option that determines whether the warning is displayed or not? We then simply add a small function that users can run to change the option value and thereby disable the warning message. (e.g. warning will only be displayed as long as options('display_glmmTMB_type_warning') == TRUE)

bbolker:
I added this  in 9c0bdeb (the option is called `glmmTMB.default.response` or something like that), but I'm not sure whether it's worth adding now that version 0.2.1 has already been pushed to CRAN ... ?

PR #335, opened 2018-05-12
Tmb compile help

bbolker:


bbolker:
I just had to deal with this on a MacOS system, so thought I would add details.  Hopefully this isn't too scary for less-technical users ...

PR #336, opened 2018-05-13
enable theta_ and beta_ for profile params

bbolker:


mebrooks:
Eventually, I think this will need more detailed documentation. For a simple RE, theta is just on the log scale, but what will happen with other covariance structures? Are some of the covariance structures parameterized by their Cholesky decomposition and if so what would one do with the profiled parameters?

PR #338, opened 2018-05-13
change na.action setting to mimic lm()

bbolker:


bbolker:
this is done for consistency with `lm()`, and primarily to avoid a warning message from `MuMIn::dredge()`, which isn't expecting a complex default value for the `na.action` argument

PR #340, opened 2018-05-14
bump required TMB version

bbolker:


bbolker:
this one worked.  Don't know where the glitch in the other one came from.

PR #342, opened 2018-05-21
orphaned fix: residuals when binomial response specified as factor

bbolker:
oops.  This was in the wrong branch, never made it into master!


bbolker:
This should probably go to CRAN as soon as we can, but since we're not supposed to submit updates more than "every 1-2  months", we should wait until at least June 10 ...

mebrooks:
I think this is ok to merge. It looks like the Travis tests are only failing on the devel version of R with 
```
'library' or 'require' call not declared from: ‘plyr’checking installed package size ... NOTE
```
I got an email on May 17th pointing out that `plyr` might be an undeclared package dependency for glmmTMB. They recommended adding it to "suggests", so we'll have to do this before the next submission to CRAN, but I don't think this is a problem now. Is it?

bbolker:
this should (?) be fixed now - I had to work a little harder to match the previous results, but it can be done without plyr.  I feel pretty strongly about minimizing dependencies ...

PR #354, opened 2018-06-06
REML in News and help

mebrooks:


bbolker:
The best (only?) description I know of applying REML to GLMMs via integrated likelihood is in Russ Millar's book (he calls it GREML).  There are some subtleties, e.g. re-estimating the regression coefficients after fixing the variances at their initially estimated values ... maybe we want to give a ref for people who are thinking about applying REML to GLMMs?  Do people know of other refs/discussion?

Millar, Russell B. Maximum Likelihood Estimation and Inference: With Examples in R, SAS and ADMB. John Wiley & Sons, 2011.

[this link](https://books.google.ca/books?id=qvzUGYmChxUC&printsec=frontcover&dq=millar+admb+reml&hl=en&sa=X&ved=0ahUKEwjwrKvJrr_bAhUdoYMKHRpTDd8Q6AEIKTAA#v=snippet&q=GREML&f=false) for Millar might work ...

Issue #357, opened 2018-06-15
simulate does not seem to work with beta_family


rdiaz02:
```
## code from https://github.com/glmmTMB/glmmTMB/issues/278
set.seed(101)
ngrp <- 100
eps <- 0.001
nobs <- 4000
ngrp <- 200
nobs <- 800
x <- rnorm(nobs)
f <- factor(rep(1:ngrp,nobs/ngrp))
u <- rnorm(ngrp,sd=1)
eta <- 2+x+u[f]
mu <- plogis(eta)
y <- rbeta(nobs,shape1=mu/0.1,shape2=(1-mu)/0.1)
y <- pmin(1-eps,pmax(eps,y))
dd <- data.frame(x,y,f)

library(glmmTMB)
m1 <- glmmTMB(y~x+(1|f),family=list(family="beta",link="logit"),
        data=dd)
m2 <- update(m1,family=beta_family())
```

```
simulate(m1)
simulate(m2)
```

give 

```
Error in simulate.glmmTMB(m1) : 
  Simulation code has not been implemented for this family
```

(I noticed this trying to use `DHARMa`: see also https://github.com/florianhartig/DHARMa/issues/16 and https://github.com/glmmTMB/glmmTMB/issues/313)

rdiaz02:
Sure, the message is very clear, so this is not a bug per se as you indicate. It would be nice to have simulate work here, though, so that we can use, say, DHARMa for diagnostics. But obviously this falls in the "wishlist" category.

mebrooks:
@mmaechler 's point about the `.Random.seed` is still unresolved.

mebrooks:
I found this code in `lme4::.simulateFun`
```
    if(!is.null(seed)) set.seed(seed)
    if(!exists(".Random.seed", envir = .GlobalEnv))
        runif(1) # initialize the RNG if necessary
    RNGstate <- .Random.seed
```

In `simulate.glmmTMB`, we only have
```
    if(!is.null(seed)) set.seed(seed)
```
@mmaechler, should we add the extra 3 lines from lme4 to resolve part (1) of your comment?

For part (2) of your comment, I'm not sure how this is working in lme4. It returns a structure including `seed = RNGstate`, but according to my naive test (below), this does not give the seed that was input.
```
seed=111
if(!is.null(seed)) set.seed(seed)
if(!exists(".Random.seed", envir = .GlobalEnv))
        runif(1) # initialize the RNG if necessary
[...]

PR #358, opened 2018-06-16
Add beta family simulate #357

kaskr:


bbolker:
that was tough :-)

bbolker:
failed test on r-devel because vignette build took >10 minutes.  We could ignore this for now, but we probably have to fuss with travis settings (either add `travis_wait` specs appropriately, or skip vignette building in tests ...)

bbolker:
should we try to fix the travis issues before or after merging?

mebrooks:
I got an email from CRAN this morning pointing out the same problem: one of the vignettes is building too slowly. The deadline to fix it is July 2nd. I don't have time to look at it at this moment, but will tomorrow...unless someone else gets to it first.

mebrooks:
I've been discussing this with Koen and the other DTU developers. @koenvanbenthem wrote code to get the timing of the vignettes. 
On my computer they are:
```
                    user.self sys.self elapsed
covstruct.rmd         98.386   9.048 107.590 
glmmTMB.Rnw           20.422   0.159  20.743
mcmc.rmd               1.907   0.177   2.208
miscEx.rmd             0.460   0.007   0.468
sim.rmd                2.610   0.045   2.721
troubleshooting.rmd   34.097   0.205  34.470
```
and they are similar on Koen's computer 
```
                    user.self sys.self elapsed
covstruct.rmd         105.826    1.584 108.609
glmmTMB.Rnw            13.129    0.060  13.907
mcmc.rmd                2.727    0.008   3.017
miscEx.rmd              0.381    0.000   0.381
sim.rmd                 2.444    0.000   2.519
troubleshooting.rmd    30.666    0.016  30.735
```

[Here's](https://cran.r-project.org/web/checks/check_results_glmmTMB.html) the latest cran check. The cran maintainer said it shows "very long run times for rebuilding the vignette outputs, often hitting the timeouts used for the check runs".

Is the criteria the *total* time to run all the vignettes? If so, it seems that we should probably keep the essential information in the vignettes and move some of the other information (possibly entire vignettes) to Github. At DTU, we were discussing this solution because, while `covstruct.rmd` is the slowest, it also contains lots of essential information.

Another solution is to avoid running some of the code by storing more objects (e.g. I could probably just use plain text in `troubleshooting.rmd` instead of running the models that don't converge).

mebrooks:
Also, since cran requests that we fix this by July 2, this is an opportunity to get in whatever fixes we want in the next version. Open for discussion...

mebrooks:
To answer @bbolker 's earlier question, I think we can merge and then deal with the vignettes in separate pull requests.

kaskr:
It's not just the vignettes. Our tests also have to be reduced:

```
Check: tests
Result: ERROR
     Running ‘AAAtest-all.R’ [10m/12m]
    Running the tests in ‘tests/AAAtest-all.R’ failed.
```

Judging from this:

```
checking re-building of vignette outputs ... [10m/12m] WARNING
Error in re-building vignettes:
  ...
Killed
```

I'm guessing the time limit for tests and vignettes are the same - and we should keep in mind that the CRAN machines aren't the fastest...

mebrooks:
Yep. It looks like CRAN's machines are more than 3.6 times slower than Koen and my computers. We don't know how far it got in the vignettes before killing them, presumably after 10 minutes, but what does the 12m refer to?

mebrooks:
vignette speed discussion moved to #364

PR #360, opened 2018-06-19
Doc reml

mebrooks:


mebrooks:
The Travis failure seems to be with the vignettes in the devel version of R. So I think this is ok to merge, but double check.

PR #365, opened 2018-07-02
suggest plyr

mebrooks:
This should remove a warning about the undeclared package dependency.


bbolker:
merging even though tests fail, since it seems to be just the absence of MCMCpack on r-devel for some reason ... we should check locally and on win-builder before submitting ...

PR #377, opened 2018-08-18
Remove redundant comment, typo

strengejacke:


strengejacke:
Just wanted to ask: do you recompile the rd-files before submitting to CRAN? As I only changed the R-files, the typos are still in the related rd-files.

PR #388, opened 2018-09-14
too much at once; mcmc, tmbstan, comment out broom.mixed

bbolker:


bbolker:
commented out broom.mixed stuff in the downstream-evaluation vignette because broom.mixed isn't on CRAN yet; hopefully it will be soon ...

PR #389, opened 2018-09-15
Fix covstruct order2

kaskr:
Perhaps best to wait merging until new lme4 is on CRAN.

### DIFF
--- a/glmmTMB/DESCRIPTION
+++ b/glmmTMB/DESCRIPTION
@@ -22,7 +22,7 @@ Depends: R (>= 3.2.0)
 Imports:
     methods,
     TMB (>= 1.7.13),
-    lme4 (>= 1.1-10),
+    lme4 (>= 1.1-18.9000),
     Matrix,
     nlme
 LinkingTo: TMB, RcppEigen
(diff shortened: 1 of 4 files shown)


mebrooks:
The test was failing because the order of the coefficients changed (not the estimates) when we started using the new `lme4:::mkReTrms`.
```
[3] -1.0757 - -0.0425 == -1.0332
[4]  0.0391 -  0.9606 == -0.9215
[5] -0.0425 -  0.0391 == -0.0816
[6]  0.9606 - -1.0757 ==  2.0363
```
I changed the test anyway so that now there are a variety of std.devs instead of all 1s in the variance components. So it's easier to see that the fit is correct when looking at the fit or summary of the fit. Instead of checking the pars on the link scale as I do in the test, it would be better to  translate them back to the scale of the std.devs.

mebrooks:
So we're just waiting to merge this when the new lme4 is on CRAN. @mmaechler @bbolker, any idea when that can happen?

bbolker:
lme4 1.18.1-1 went on CRAN on 17 Aug, so according to CRAN guidelines we have to wait at least 3 more weeks. I think that if lme4 feels like it's slowing glmmTMB progress, you should just make a temporary copy of `mkReTrms` until such time as lme4 catches up (I already included an `lme4_tmp_utils.R` file for another component that's updated in the development/GH version of `lme4` ...)

bbolker:
FWIW lme4 maintainers have been asked to push a new version to CRAN by October 9, after which we can push this version to CRAN without copying methods from lme4 ...

mebrooks:
That's good new!. I guess we should aim to submit to CRAN shortly after lme4 does so. I'm currently working towards another deadline on Nov 7th, but I should have time to work on glmmTMB later in November.

PR #403, opened 2018-10-18
preliminary fix for data-dependent bases

bbolker:
This is (1) not heavily tested and (2) not as elegant as I'd like (see especially `fix_predvars` in `utils.R` ... using `append()` would make it *slightly* nicer, but it still uses `all.vars()` in  a fairly crude way ... but this seems like an important bug.  Comments and more extensive test cases welcome!


bbolker:
if anyone wants to look this over (and/or propose a more extensive set of tests), I'd appreciate it!

bbolker:
well, *that* was harder than I expected, but I think everything is cleaned up now (without huge amounts of technical debt - in fact, I think (hope) `RHSForm()` is improved over its previous version)

PR #408, opened 2018-10-30
added TMB macro DATA_UPDATE on response (yobs)

tiboloic:
response can by changed by changing obj$env$data$yobs.

allows to refit a model on simulated data without having to rebuild the tape.

### DIFF
--- a/glmmTMB/src/glmmTMB.cpp
+++ b/glmmTMB/src/glmmTMB.cpp
@@ -577,6 +577,7 @@ Type objective_function<Type>::operator() ()
   DATA_SPARSE_MATRIX(Zzi);
   DATA_MATRIX(Xd);
   DATA_VECTOR(yobs);
+  DATA_UPDATE(yobs);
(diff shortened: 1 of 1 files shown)


kaskr:
I worry that a lot of glmmTMB models would face a significant performance hit by this change. Optimizations utilizing the actual data values would no longer be possible. For instance, by knowing that an observation is zero, the AD machinery can often reduce the likelihood calculation dramatically. If one were to allow `DATA_UPDATE` there should at least be a flag to disable it.

bbolker:
is there any way to make this selectable at run-time by the user (and make it default to the previous behaviour)?

tiboloic:
Thanks Kasper for the insights, I didn't realize than DATA_UPDATE had a direct impact on the performance. I just ran a simulation study and it appears that, indeed, DATA_UPDATE tends to slow things down rather than speed things up: I could see benefits only for small datasets.
I will delete the pull request.

PR #424, opened 2018-12-18
Fixing MCMC vignette with workaround for knitr function

sadatnfs:
- Wrapping `knitr::read_chunk` in order to force this in current R environment

Reference for fix: https://github.com/yihui/knitr/issues/1647

### DIFF
--- a/glmmTMB/vignettes/mcmc.rmd
+++ b/glmmTMB/vignettes/mcmc.rmd
@@ -20,8 +20,8 @@ One commonly requested feature is to be able to run a *post hoc* Markov chain Mo
 ```{r knitr_setup, include=FALSE, message=FALSE}
 library(knitr)
 opts_chunk$set(echo = TRUE)
-mcscript <- system.file("vignette_data","mcmc.R",package="glmmTMB")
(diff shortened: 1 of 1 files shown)


mebrooks:
Yep, I ran `make test` on my computer and the problems looked minor and unrelated. The only error I got was 
```
test-methods.R:325: error: various binomial response types work
object 'ddb' not found
1: fixef(lme4::refit(f1b, s1[[1]])) at /Users/molliebrooks/Documents/glmmTMB development/glmmTMB/glmmTMB/tests/testthat/test-methods.R:325
2: lme4::refit(f1b, s1[[1]])
3: refit.glmmTMB(f1b, s1[[1]])
4: eval(cc$data) at /Users/molliebrooks/Documents/glmmTMB development/glmmTMB/glmmTMB/R/methods.R:1096
5: eval(cc$data)
```

sadatnfs:
Oh yeah that is weird, because I just prepped all the regression and the data, and running the innards of that binomial test didn't throw me any errors. The error is even weird since it says it couldn't find the data.frame...

PR #430, opened 2019-01-03
Vignette speed

mebrooks:


bbolker:
what was up with the "remove double or ||" stuff?

mebrooks:
I had fixed the same thing in this branch that was fixed in the for_cran branch with [this commit](https://github.com/glmmTMB/glmmTMB/commit/39b91950565c77b3d12a3f7fe70863dbde36ac25) but I kept it as two lines, so they weren't compatible and I had to revert my commit before merging with the master branch that now contains the for_cran changes.

bbolker:
OK, thanks, that makes sense

PR #447, opened 2019-03-12
citation for mean-parameterized compois

### LINKED ISSUE #116: document family options
I believe that we have more distributions available than are documented for the standard glm family in family {stats}. So we may need to document them manually (e.g. beta-binomial).

mebrooks:
Fixes #116


mebrooks:
Even though @kaskr thought of it independently, Huang was the first to publish it.

PR #457, opened 2019-03-27
add 'start' parameter

bbolker:
This allows the user to specify starting values.

I think it's fairly harmless; the only open question I can think of is whether we'd rather make `start` an additional top-level parameter to `glmmTMB`, or whether it should instead be included as part of `glmmTMBControl`  (`lme4` has it as a top-level parameter).

### DIFF
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -185,7 +185,7 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
     
   for (p in names(start)) {
       if ((Lp <- length(parameters[[p]])) !=  (Ls <- length(start[[p]]))) {
-          stop(sprintf("parameter vector length mismatch: length(%s) is %d, should be %d", p, Lp, Ls))
+          stop(sprintf("parameter vector length mismatch: length(%s) is %d, should be %d", Lp, p, Ls))
       }
       parameters[[p]] <- start[[p]]
   }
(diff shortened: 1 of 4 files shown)


mebrooks:
Also, to be totally clear that the problem comes from `start`, I would write 
```
          stop(sprintf("start parameter vector length mismatch: length(%s) in start is %d, should be %d", p, Ls, Lp))

**Comment by bbolker** [comment]
how about now?

**Comment by bbolker** [comment]
for what it's worth, the Travis checks didn't really succeed - they crashed with segfaults. Transient problem? Matrix/TMB binary incompatibility?

**Comment by mebrooks** [comment]
When I tried to install again, I started having some problems that seem unrelated. I'm still looking into it.

**Comment by mebrooks** [comment]
That was so weird. I don't think I changed anything on my computer, but I started getting `fatal error: 'wchar.h' file not found` while trying to compile the revised version of this branch. The solution was to run 
```
open /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg
```
in the terminal.

It looks good now.

mebrooks:
A little more detail...the error was occurring on any version of the package that I tried to compile, so it had nothing to do with this branch. My computer forgot where `'wchar.h'` was living even though it was still in the same place it had recently been. The only thing that was going on with my developer tools was that I was putting off doing an update of Xcode that the installer kept on suggesting.

kaskr:
@bbolker Still segfaults in the travis tests. It seems to me they got introduced in a426a627637cc095af3b87eb7c8aa918b3de39fd

PR #469, opened 2019-04-25
Genpois phi docs retry

mebrooks:


bbolker:
Travis checks are weird (segfaulting in AR1 fit during vignette build, but still claiming to have succeeded, but that's not relevant to these changes.

PR #487, opened 2019-05-29
Fix confint ar1

bbolker:
is there a reason to wait to merge this?

### DIFF
--- a/glmmTMB/R/methods.R
+++ b/glmmTMB/R/methods.R
@@ -746,6 +746,26 @@ confint.glmmTMB <- function (object, parm, level = 0.95,
                                                                     sep="."),
                                                  estimate = estimate)
                 ci <- rbind(ci, ci.sd)
+                ## VarCorr -> correlation
+                getCorrelation <- TRUE ## FIXME: Make 'getCorrelation' a function argument ???
+                if (getCorrelation) {
+                    mat2vec <- function(m) {
+                        ans <- structure(as.vector(m),
+                                         names=do.call(paste,c(expand.grid(dimnames(m)),sep=":")))
+                        ans[lower.tri(m)]
+                    }
+                    reduce <- function(VC) unlist(lapply(VC[[component]],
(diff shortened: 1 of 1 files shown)


bbolker:
OK, I see: Wald case is fixed but others aren't yet.

kaskr:
We have to be careful not to apply this method too generally. Recall that it simply transforms the univariate confidence intervals, i.e. [a,b] => [f(a),f(b)].
This works fine for the `ar1()` parameter, because the internal parameterization is `phi = f(theta)`. However, it wont work for more complex cases such as the correlation matrix elements of a `us()` term, because each such element is a function of multiple thetas.

bbolker:
in the meanwhile there are some issues in here (in `testthat/test-methods.R`): there are both naming changes (which might be OK as long as we warn users in `NEWS`) but some further mismatches  (the last row of the new version is bogus, should be the `sigma` CIs ...) Will have to dig in and see what's going on.

```
## new
> ci.2NB
                                         2.5 %     97.5 %
cond.(Intercept)                    5.48098714 5.58422551
cond.Days                           0.02481639 0.04289932
Subject.cond.Std.Dev.(Intercept)    0.06617725 0.15091785
Subject.cond.Std.Dev.Days           0.01134364 0.02635499
Subject.cond.Cor.Days.(Intercept)  -0.52088384 0.50221168
cond.Corr.Subject.Days:(Intercept) -0.52088384 0.50221168

## old/expectedt
> ci.2NB.expect
                                 2.5 %       97.5 %
cond.(Intercept)            5.48098714   5.58422551
cond.Days                   0.02481639   0.04289932
cond.Std.Dev.(Intercept)    0.06617725   0.15091785
cond.Std.Dev.Days           0.01134364   0.02635499
cond.Cor.Days.(Intercept)  -0.52088384   0.50221168
sigma                     183.81058474 444.73566864
```

mebrooks:
I was just starting to try to resolve the conflicts here. Is that still worth doing for this branch?

PR #490, opened 2019-06-10
add "weights()" method

strengejacke:
I think I have a different save-option in RStudio, that's why more lines appear as "revised" or "changed", although less code was changed, and nothing deleted.

This PR adds a `weights()` method, to retrieve model-weights. If there are no exceptions, it should be just as easy as returning the `"(weights)"` column from the `model.frame()`. Tested on a ZI-poisson model, and added a test.


mebrooks:
What is the motivation for this?

Editing in R studio will lead to a lot of white-space changes. We would prefer to avoid white-space changes because it makes it harder to follow the history of changes for a certain line.

I believe in `aggdat$Freq`, `freq` should be lower case.

strengejacke:
> What is the motivation for this?

You mean for the `weights()`-method? Simply to have less exceptions for different model objects when users call `stats::weights()` to extract weights.

> Editing in R studio will lead to a lot of white-space changes. We would prefer to avoid white-space changes because it makes it harder to follow the history of changes for a certain line.

Yes, I saw this first after submitting the PR, I think my option is set to "Strip trailing horizontal white spaces on saving" or so.

> I believe in aggdat$Freq, freq should be lower case.

In the model code, it's upper case: `glmmTMB(y ~ x+(x|i), data=aggdat, weight=Freq,`.

strengejacke:
Ok, whitespace are restored, so only the "actual" changes are committed.

bbolker:
Quick comments:

- there is a `stats::model.weights()` accessor method which would work (i.e. `weights.glmmTMB <- function(x) model.weights(model.frame(x))`: on the other hand, its entire definition is `x$"(weights)"` (the use of the `$`-accessor seems weird to me, I may have commented on r-devel about that in the past ...) so it's equivalent; it doesn't seem it would add much to use the built-in accessor ...
- given that `weights.glm` and `weights.merMod` have a `type=c("prior","working")` argument, it might be worth adding a check/error if the user specifies `type="working"` (as we don't currently have an easy way to retrieve working weights from a GLMM).

strengejacke:
I know there are many ways to access the weights (as well as there are many ways to access many other information). However, as someone who has many colleagues that work with SPSS and try to use R every now and then, my experience is that there's a lot of confusion when there is a method that works for some models, although it could / should work for other models as well. Newcomers in R are often beaten by its complexity. So, the simple reason for this short method is to make **glmmTMB** behave like many other standard packages, making it most accessible to new R users (even if this case is probably not the most urgent problem).

I tried to address the `type = "working"` issue now, however, I'm not sure which wording you prefer. Feel free to make suggestions or changes.

bbolker:
I wasn't arguing that the model.weights approach is preferable to having an
explicit weights method, just exploring possibilities for implementation.
(I've learned from r-core that a little bit of obsessiveness about
implementation isn't always a bad idea)

bbolker:
Hmm.  I think specifying `type` with a value other than "prior" (perhaps allowing partial matching) should be an error, not a warning (there is no correct use I can think of, and warnings can always be ignored or missed). I'm happy to make this change if everyone agrees, but don't want to step on your toes.

strengejacke:
Feel free to make those changes - I would leave it up to you as pkg developers to decide on such issues, since every team has a different design philosophy, I would say.

I wasn't sure if it makes sense to include a `type`-argument if there are no options for `type`, so I decided to evaluate the `...` in order to find such a possible argument. But then, I was not sure how to do partial matching?

bbolker:
If we implement `type` with a single option and use `match.arg()` (and document that working weights aren't available at present) I think that will take care of everything nicely.  I'll take a crack at this.

PR #495, opened 2019-06-21
Fix anova comp

### LINKED ISSUE #494: argument component in Anova has no effect
I noticed that the component argument to ANOVA has no effect, i.e. irrespective of whether I specify component = 'cond' or argument = 'disp' I always get the ANOVA table of the conditional model, rather than the dispersion model.

This might be related to the code lines

afun(mod, vcov., test = test.statistic, singular.ok = singular.ok)

and 

Anova.II.glmmTMB <- function(mod, vcov., singular.ok=TRUE, test="Chisq",
component="cond", ...){

in Anova.R, which do not pass on the value of component to the downstream functions, rather falling back onto the default component = 'cond'.

Any chance, this could be fixed? Thanks.

bbolker:
I made some moderately large changes in the `terms.glmmTMB()` and `model.matrix.glmmTMB()` methods (and `fitTMB` returns an additional `terms` component in the output), so I wouldn't mind if someone had a look. This should fix #494 ...


bbolker:
FWIW this page says "some checks haven't completed yet", but when you click through to the Travis page it says they're done (and passed)

PR #498, opened 2019-06-24
Troubleshooting adds

bbolker:
modifications of the troubleshooting vignette - a new example and some slight rearranging/additions.


mebrooks:
I'm able to install it using 
```
devtools::install_github("glmmTMB/glmmTMB/glmmTMB", ref="troubleshooting_adds", force=TRUE, build_vignettes = TRUE)
```
and then if I try `make install`, it still gives me the same error.
```
* creating vignettes ... ERROR
Quitting from lines 168-169 (troubleshooting.rmd) 
Error: processing vignette 'troubleshooting.rmd' failed with diagnostics:
object 'mod1' not found
Execution halted
make: *** [glmmTMB_0.2.3.tar.gz] Error 1
```

bbolker:
The problem is the `NOT_CRAN` stuff that was introduced in e339e2c49 . I added an `export  NOT_CRAN=true` to `make vignette-update`, have now added it to `make install` as well.  (1) I'm not sure `export` is going to work for everyone? (I think it's a bash command ...) Is there a way to specify environment variables when invoking R? (2) Is this the best approach? (3) Might be worth adding a comment to the vignette itself specifying that you can use `Sys.setenv(NOT_CRAN="true")` from within an R session ...

bbolker:
Hmm, `make install` still seems to fail at that point?

bbolker:
I don't like it, but I removed all the inline chunks.  It would be better to figure out how to get this to work *with* inline chunks ...

bbolker:
bump ...

mebrooks:
I have no idea why inline chunks aren't working. They're used in another vignette [here](https://github.com/glmmTMB/glmmTMB/blob/f4def225a32f650fb82e6e6f118309ee3ece1491/glmmTMB/vignettes/covstruct.rmd#L171) and it looks like the same syntax as you had.

Issue #438, opened 2019-01-19
confint only prints the first two parameters in version 0.2.3


ShrewMaster:
This function used to work just fine under version 0.2.2. When I updated to version 0.2.3, it stopped printing all the parameters of the model by default, and now only prints the first two. The parameters are still there, it's just not printing them by default. Maybe something changed in another package I updated?

In other words, under version 0.2.2, `confint(model)` printed everything correctly.
Under version 0.2.3, I have to type `confint(model, parm = c(1:[_maximum number of parameters that can be printed_]))` to obtain the same result as before.
If I type `confint(model)` under version 0.2.3, it gives the same result as `confint(model, parm = c(1:2))` under version 0.2.2.

I hope this was clear.

bbolker:
Can we have a reproducible example please?

```
fm0 <- glmmTMB(Murder~Illiteracy+Population+Area+`HS Grad`,data=as.data.frame(state.x77))
> confint(fm0)
                         2.5 %       97.5 %      Estimate
cond.(Intercept) -7.469211e-01 1.451238e+01  6.882728e+00
cond.Illiteracy   1.642143e+00 4.693255e+00  3.167699e+00
cond.Population   7.291709e-05 3.601559e-04  2.165365e-04
cond.Area         2.072511e-06 1.950013e-05  1.078632e-05
cond.`HS Grad`   -2.139798e-01 2.969804e-02 -9.214087e-02
sigma             1.868788e+00 2.765670e+00  2.273423e+00
```

ShrewMaster:
That's interesting… I tried your example and it worked fine. I also tried the "Owls" example you talk about in your glmmTMB vignette, and it too printed correctly.

I investigated the problem a little deeper and eventually figured it out. It only happens when you set REML = TRUE, which is what is happening with my models. I reproduced your example again, but this time with REML = TRUE. The issue resurfaced.

`> fm0 <- glmmTMB(Murder~Illiteracy+Population+Area+`HS Grad`,data=as.data.frame(state.x77), REML = TRUE)
> confint(fm0)
                     2.5 %  97.5 % Estimate
cond.(Intercept) -1.159607 14.9251 6.882749
> confint(fm0, parm = c(1:5))
                         2.5 %       97.5 %      Estimate
cond.(Intercept) -1.159607e+00 1.492510e+01  6.882749e+00
cond.Illiteracy   1.559618e+00 4.775771e+00  3.167694e+00
[...]

PR #502, opened 2019-07-03
Reml c ifix

bbolker:
This is an attempt to address #438, and do some miscellaneous cleanup. There a few potentially non-backward-compatible changes here (which could probably be rolled back if they seem like a bad idea):

- dispersion parameters for trivial-dispersion (`~1`) models are not included in confint/profiles unless `full=TRUE` is specified
- row names for conditional-parameter-only confint objects have `cond.` stripped

I haven't added these to NEWS yet.


mebrooks:
I don't understand what's going wrong with the tests. I reproduce the Travis error on my computer when I run `make install`. 
```
test-methods.R:88: failure: drop1
dd$AIC not equal to c(1763.94, 1785.48).
Lengths differ: 1 is not 2
```
but those lines of code work fine on my computer if I run the individual lines: 
```
load(system.file("test_data", "models.rda", package="glmmTMB",
                 mustWork=TRUE))
dd <- drop1(fm2,test="Chisq")
expect_equal(dd$AIC,c(1763.94,1785.48),tol=1e-4)
```
I get a different error if I try to source the file. 
```
> source("tests/testthat/test-methods.R")
Error: Test failed: 'confint'
* `ci` not equal to structure(...).
Attributes: < Component "dimnames": Component 1: 2 string mismatches >
* dim(cc) not equal to c(5, 3).
1/2 mismatches
[1] 1 - 5 == -4
* rownames(cc) not equal to c("(Intercept)", "Illiteracy", "Population", "Area", "`HS Grad`").
Lengths differ: 1 is not 5
```

Does anyone have a hunch about what is happening?

bbolker:
May have something to do with incompatible changes in the `terms()` method (779b3d54947).  I'll see when I can get a chance to fix it.

bbolker:
I think the issue was with a (somehow?) messed up stored file of testing examples.  If you hadn't re-installed the package (i.e., you were still using an old & OK stored file of testing examples), then you wouldn't have had the problem.

PR #511, opened 2019-08-13
pop-level prediction

bbolker:
This could certainly be extended to try to allow a full `re.form` specification (i.e., use particular REs and ignore others), but to do that we would need a robust way to map from a RE formula to a vector specifying which `theta` values to map to -100 (or whatever).

### DIFF
--- a/glmmTMB/R/predict.R
+++ b/glmmTMB/R/predict.R
@@ -237,7 +237,19 @@ predict.glmmTMB <- function(object,newdata=NULL,
     ) stop("Some variables in newdata needed for predictions contain NAs or NaNs.
            This is currently incompatible with se.fit=TRUE."))
   }
-  
+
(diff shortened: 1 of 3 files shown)


bbolker:
OK, I've tried to implement your first suggestion.  Still not well-tested!

florianhartig:
Hi guys, sorry for my sluggish response - I have tested this in https://github.com/florianhartig/DHARMa/tree/2.4.2-glmmTMBfix and it works like a charm, at least as far as the integration in DHARMa and the problem I had with https://github.com/florianhartig/DHARMa/issues/16 is concerned. As soon as you push this on CRAN, I will do an update to DHARMa!

florianhartig:
Thanks for the work on this, this is really useful for me, and I expect also for some other users!

PR #523, opened 2019-10-21
handle contrasts in predict(): fix #439

### LINKED ISSUE #439: predict() does not work with non-default contrasts
... specifically, contrasts that do not provide column names exactly equal to the levels of the factor in question.

```
> contrasts(iris$Species) <- contr.sum
> mod <- glmmTMB(Sepal.Length ~ Species,iris)
> predict(mod)
Error in checkModelMatrix(data.tmb1$X, data.tmb0$X) : 
  Prediction is not possible for unknown fixed effects: Speciesversicolor, Speciesvirginica
Probably some factor levels in 'newdata' require fitting a new model.
```

I tried digging deeper into `predict.glmmTMB` to see where exactly things went south:

- `TMBStruc$data.tmb$X` (which becomes `data.tmb1$X` in `checkModelMatrix()`) contains the factors laid out side-by-side named and treatment-coded, as you would get from `model.matrix(~Species,iris)` _without having changed the contrasts_
- `object$obj$env$data` (which is `data.tmb0$X`) contains the correctly-coded data
- so obviously, they mismatch and `checkModelMatrix()` errors out

Unfortunately, I don't understand the internals well enough to propose a fix...

bbolker:


bbolker:
ugh, harder than I thought.  Still working on a fix that works in all possible edge cases (`allow.new.levels=TRUE`, etc.)

bbolker:
ugh, still doesn't work (breaks one of the vignette examples)

bbolker:
As anyone can see from the commit history here, this turns out to be a big pain, with lots of edge cases. Some of the (possibly unnecessary) complexity comes from checking for inconsistent contrasts set in the prediction vs the original data.  If someone can make an argument that this doesn't even matter (i.e. contrasts aren't actually used for anything important in the prediction process), we might be able to strip this stuff out.

I would be happy for second opinions on this code.

bbolker:
Can someone take a look at this please?  (It would be best to look at it, but would probably be better to merge it thoughtlessly than to leave #439 unfixed ...)

mebrooks:
I ran all the tests and looked at the file changes. It all seems ok to me, but I don't have much experience with contrasts, so I can't really answer your question about whether or not we can ignore them in the predict code. The tests seem reasonable. @mmaechler, do you have any insight on contrasts?

mebrooks:
OK, I guess we should merge this branch. Any other thoughts first?

Issue #474, opened 2019-05-02
fitting count distributions to non-integer data should give a warning


bbolker:
In my experience these kinds of warnings catch a lot of problems, e.g. users confused about varying exposure/area per observation (i.e. don't know about offsets). I'm pretty sure that both `glm()` and `lme4::glmer()` give warnings in this situation, we can look into how/where the checks are implemented.

```
z <- rlnorm(100,1,1)
library(glmmTMB)
glmmTMB(z~1,family=poisson)
```

mebrooks:
I agree. Someone's been emailing me lately and asking if it's ok to use `family=compois` for non-integer data. I'm trying to explain that it's best to use the raw data if possible. They pointed out something interesting, but I haven't had time to think about it carefully: in the Poisson and CMP, if you multiply the response variable by a constant (e.g. in their case to make it an integer) then the intercept absorbs the multiplication.
```
lambda=exp(1.5)
n=10000
dat=rpois(n, lambda)
dat0.1=dat*0.1  #non-integer data

fixef(glmmTMB(dat~1, family=poisson))
fixef(glmmTMB(dat0.1~1, family=poisson)) #non-integer data
log(lambda)+log(.1)
```

mmaechler:
Of course, the *mean* multiplication factor goes into the intercept with a log() link.  
But the variance and all inference becomes completely wrong... and I agree with Ben that I'm sure many users produce nonsense analysis by using inappropriate count models for non-count but "fudged to look like count" data,  with "multiplication" being just one of several complete nonsense tricks.

I don't have too much time right now, and I'm not sure if it easy to reliably detect such user fudging.
If we give warning for non-integer, people will tell (on stackoverflow, twitter,blog, ..) how to "fix" this, e.g. by rounding their non-integers or similar nonsense.   Honestly I'm appalled how students and data analysts and probably many people who need their research papers to get published .. blog and chat and recommend tricks to each other on what to do to get rid of warnings and errors ... or use a package that does not give the warnings or errors ...  
 Even our stats master students use Google to solve their R modeling problems and cut'n'paste code found "on the internet"  and are clueless what the code does (apart from "it works afterwards")....

mebrooks:
Yep, you're right. I hadn't noticed, but the std errors are completely different, so inference would be wrong.

bbolker:
A little more information: 

[...]

PR #525, opened 2019-11-02
warn on non-integer responses in count families (address #474)

bbolker:


bbolker:
I've refactored the handling of zeros in distributions that shouldn't have zeros (but still want to handle the zero-inflated case correctly).  The changes can be seen in https://github.com/glmmTMB/glmmTMB/commit/8c844f55148951e71ab44f3dad04c0be40834899 . In particular, we now assume a zero in  a truncated count distribution only `yobs(i)<0.001`, not if it is exactly zero. **However**: this apparently makes some of the count distributions fail with non-integer responses (compois, truncated compois, truncated nbinom1/nbinom2), and I'm not sure why.  These are commented out for the moment - they shouldn't necessarily work anyway - but I can't see what in the code is leading to an NaN function evaluation ... see tests [here](https://github.com/glmmTMB/glmmTMB/blob/warn_noninteger/glmmTMB/tests/testthat/test-families.R#L68-L82)

glmmTMB/src/glmmTMB.cpp
```diff
@@ -645,8 +645,10 @@ Type objective_function<Type>::operator() ()
   vector<Type> phi = exp(etad);
 
 // "zero-truncated" likelihood: ignore zeros in positive distributions
-//  (currently only used in gamma_family, beta_family)  
-#define zt_lik(x,loglik_exp) (zi_flag && x==Type(0) ? -INFINITY : loglik_exp)
+// close to zero: use for count data (cf binomial()$initialize)
```
mebrooks: Are the comments on lines 648 and 650 swapped?

PR #526, opened 2019-11-12
swap comments on zt_lik_zero and zt_lik_nearzero

mebrooks:


bbolker:
I don't think you really need to submit pull requests on pull requests -- you could just patch the branch directly ...  thanks.

bbolker:
PS if you include "[skip ci]" in your commit message, it disables Travis checking (which is probably unnecessary if all you're doing is moving comments around).

PR #537, opened 2020-01-09
Conditional openmp

bbolker:
I'm opening this pull request not because I think it should definitely be pulled immediately, but as a venue for asking some questions/improving my understanding of how OpenMP is implemented and controlled.

* the call to `TMB::openmp(n)` (in `glmmTMB.R/fitTMB`, controlled by `control$parallel`) is the most obvious to me: this enables OpenMP machinery in the underlying TMB structure, with `n` threads
* what does `max_parallel_regions = omp_get_max_threads()` (in `glmmTMB.cpp`) do? (the comment says "Set max number of OpenMP threads to help us optimize faster" but I don't know what this means; is this complementary to/redundant with/interfering with calling `TMB::openmp()` ?
* I understand that  `parallel_accumulator<Type> jnll(this);` (l.  632 of `glmmTMB.cpp` specifies that `jnll` should be added up in parallel; as in the previous example, how does this interact/intersect with the TMB OpenMP machinery?
    * Why does the code compile when this line is included even when OpenMP is not available? Is TMB.h doing something clever?
    * When I was trying to strip the OpenMP stuff out I changed this back to `Type jnll=0` and got segfaults, maybe because of other changes to the code ...

I would experiment with all of these, but (1) doing the combinatorial exploration is tedious and (2) I don't have access to a machine where parallelization makes a really big difference, so I would have to run really long/slow jobs to see clearly what was going on ...

Relevant background: [my comment from July 18](https://github.com/glmmTMB/glmmTMB/issues/481#issuecomment-511086730) and [Nafis Sadat's response](https://github.com/glmmTMB/glmmTMB/issues/481#issuecomment-512644073)


kaskr:
`TMB::openmp()` sets or gets the value of `omp_get_max_threads()`.

[max_parallel_regions](https://github.com/kaskr/adcomp/blob/e669048ea54b6d8d7a67514282027a94525230d3/TMB/inst/include/tmb_core.hpp#L690) is a member of the `objective_function` class. It denotes the number of 'buckets' used for the parallel split of the computational graph.

It follows that the statement `max_parallel_regions = omp_get_max_threads()` sets the number of 'buckets' used for the parallel accumulation to the maximum number of threads available.

The [parallel_accumulator](https://github.com/kaskr/adcomp/blob/e669048ea54b6d8d7a67514282027a94525230d3/TMB/inst/include/tmb_core.hpp#L953) works by overloading the `+=` and `-=` operators. When increment/decrement is called it modifies the current bucket and changes to a new bucket. It's OK if there's only one bucket, i.e. `parallel_accumulator` works in serial. However, it is **not** OK to enable `-fopenmp` unless `parallel_accumulator` is in use. The openmp flag makes TMB assume there is a parallel accumulator in play. If not, it will crash.

glmmTMB/src/glmmTMB.cpp
```diff
@@ -577,9 +579,10 @@ template<class Type>
 Type objective_function<Type>::operator() ()
 {
 
+#ifdef _OPENMP
   // Set max number of OpenMP threads to help us optimize faster
   max_parallel_regions = omp_get_max_threads();
```
kaskr: This seems redundant. Already done by `parallel_accumulator` CTOR.

glmmTMB/src/glmmTMB.cpp
```diff
@@ -1,6 +1,8 @@
 #include <TMB.hpp>
 #include "init.h"
+#ifdef _OPENMP
 #include <omp.h>
+#endif
```
kaskr: Redundant: Already done by `TMB.hpp`.

mebrooks:
Do you think this is ready to merge now?

bbolker:
Does it run faster with serial than with parallel ... ?

I think this is *probably* OK but maybe wait a day (just posted a request for feedback to r-sig-mixed-models ...) in case anything else pops up.

mebrooks:
The tests were faster in parallel than in serial on my computer. 

I guess now we need Edgar Gonzalez or @idworkin to see if the latest commit fixes their error. Right?

bbolker:
I heard from Edgar by email: OpenMP stuff was fine, he ran into another problem caused by our recent `emmeans`-related changes. I think I've fixed this now (see recent rash of commits).  Haven't heard back from @idworkin (is there anyone else commenting in #481 who had the same problems who we could check back with?)

bbolker:
Just had another confirmation that this branch solves the `-fopenmp: not found` problem, so I think we should go ahead and merge it.  A number of mysteries remain (why does serial==parallel performance on some platforms? Is this because the number of threads is never getting set >1, *or* never getting reduced below the maximum?  What leads to performance dropping, sometimes significantly, for greater numbers of threads? How does one profile openmp-within-C++-within-R code?) but I think this release won't be any *worse* for anyone than previous releases; as far as I can tell we've solved the obvious installation problems.

mebrooks:
Today I'm getting an error on this branch (see below). I'll try to sort out what's going on.
```
* installing *source* package ‘glmmTMB’ ...
** using staged installation
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... configure: error: in `/private/var/folders/8t/zsp25_b17td7j1y5d_nmfqt00000gq/T/Rtmp1UJ40I/Rbuild3a4b240ad809/glmmTMB':
configure: error: cannot run C++ compiled programs.
If you meant to cross compile, use `--host'.
See `config.log' for more details
ERROR: configuration failed for package ‘glmmTMB’
* removing ‘/private/var/folders/8t/zsp25_b17td7j1y5d_nmfqt00000gq/T/Rtmp1UJ40I/Rinst3a4b43ed495b/glmmTMB’
      -----------------------------------
ERROR: package installation failed
make: *** [glmmTMB_1.0.0.tar.gz] Error 1
```

mebrooks:
Well, this is interesting! The errors above were in the Terminal with `make install`. Then I tried in the R console (see input below) and installing worked. I thought maybe updating RcppEigen and Rcpp in the R console could have been the trick, so I went back to the Terminal and tried again and got the same errors. When I try the point and click installation on this branch in Rstudio, I get the same errors as in the Terminal.
  
```
> remotes::install_github("glmmTMB/glmmTMB/glmmTMB@conditional_openmp")
Downloading GitHub repo glmmTMB/glmmTMB@conditional_openmp
These packages have more recent versions available.
Which would you like to update?

1: All                                      
2: CRAN packages only                       
3: None                                     
4: RcppEigen (0.3.3.5.0 -> 0.3.3.7.0) [CRAN]
5: Rcpp      (1.0.2     -> 1.0.3    ) [CRAN]

Enter one or more numbers, or an empty line to skip updates:
1
RcppEigen (0.3.3.5.0 -> 0.3.3.7.0) [CRAN]
Rcpp      (1.0.2     -> 1.0.3    ) [CRAN]
```

bbolker:
Based on `x86_64-apple-darwin19.2.0`, this looks like [MacOS Catalina](https://en.wikipedia.org/wiki/Darwin_(operating_system); I strongly suspect that's related to the problem (there always seem to be hiccups with new MacOS versions ...). It's certainly surprising that it works differently through different channels, but not **that** surprising (default search paths could be different). Are either of these useful?

* https://stackoverflow.com/questions/tagged/r+macos-catalina
* https://stackoverflow.com/questions/58278260/cant-compile-a-c-program-on-a-mac-after-upgrading-to-catalina-10-15/58349403#58349403

PR #546, opened 2020-01-23
downstream stuff conditional on car/Anova versions

bbolker:
an attempt to pass tests on windows-oldversion

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,7 @@ TAGS
 
 glmmTMB/build
 glmmTMB/revdep
+glmmTMB/config.log
+glmmTMB/config.status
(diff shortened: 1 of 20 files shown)


bbolker:
@mebrooks , `make test` works for me (except for OpenMP warnings because OpenMP isn't supported on the machine I'm running it on). This looks like another testthat scoping issue (i.e., this code has been in the master branch for a while and doesn't seem to cause any problems under `R CMD check` or `devtools::check()` from within an R session, at least not on Travis or on any of my systems). 

However, the problem looks like it's proximally caused by some leftover junk, which I've just removed ... try again?

Issue #493, opened 2019-06-19
version 0.2.4 checklist


bbolker:
I was thinking that it might be time to push a new version to CRAN (it's been about six months and there are some improvements); in any case, there always seems to be a lag between starting to think about a new CRAN version and actually doing it ...)  To be honest the initial motivation was because we now have a TMB/Matrix inversion (i.e., version of Matrix on CRAN is more recent than the TMB version, leading to mismatch warnings etc.), but that doesn't actually have anything to do with updating TMB ...

bbolker:
bump. I think we should include `re.form` (#505) ...

mebrooks:
Yep. I think this is a good idea and I'll try to start moving towards that goal, now that I finished teaching 2 days of teaching fisheries stats for the first time :) 

While checking the latest pull request, I went ahead and did a "check as cran" and found one little thing...

In `man/Owls.Rd`, this link is dead http://www.sciencedirect.com/science/article/B6W9W-4PK8B6H-8/2/e43cfbaad4dc0bb2207adfc54a460c89

Other than that it's looking good so far. I'll keep looking at the pull request.

bbolker:
fixed Owls link: f84f3e953

@florianhartig said he would test the `re.form` stuff when he gets a chance. It'd be nice to make it more feature-ful (i.e., be able to specify population-level effects for some but not all REs) but I'd rather have a robust/well-tested release than a feature-ful one.

We still have only 11 reverse dependencies so rev-dep-checking shouldn't be too onerous.

bbolker:
bump ... CRAN is back open now ...

mebrooks:
I checked the `reform` branch as CRAN and it was fine. As for the reverse dependencies though, I ran `tools::check_packages_in_dir(reverse = "all", dir ="/Users/molliebrooks/Documents/glmmTMB development/glmmTMB" )` and it only checked DHARMa and sjPlot, but you say there are 11. Am I missing something?
```
> tools::package_dependencies(reverse = TRUE)$glmmTMB
[1] "DHARMa" "sjPlot"
```

I think it will be ready to submit after @florianhartig has a chance to test the `re.form`. I did a little testing of it, but did not check for exact expected numbers.

bbolker:
[...]

PR #547, opened 2020-01-27
work around variation in downstream methods exports

bbolker:
attempt to condition tests based on what downstream methods are actually available ...

### DIFF
--- a/glmmTMB/R/emmeans.R
+++ b/glmmTMB/R/emmeans.R
@@ -31,8 +31,14 @@
 ##' if (require(emmeans)) {
 ##'     emmeans (warp.lm, poly ~ tension | wool)
 ##' }
-##' if (require(car) && packageVersion("car")>="3.0.6") {
-##'     Anova(warp.lm,type="III")
+##' if (require(car)) {
+##'     if ("Anova.glmmTMB" %in% methods("Anova")) {
+##'        ## should be available in later versions
+##'        Anova(warp.lm,type="III")
+##'     } else if (length(find("Anova.glmmTMB"))>0) {
+##'        ## try to test back-compatibly
+##'        Anova.glmmTMB(warp.lm, type="III")
+##'     }
 ##' }
 ##' if (require(effects) && packageVersion("effects")>="4.1.4") {
 ##'     plot(allEffects(warp.lm))
--- a/glmmTMB/R/emmeans.R
+++ b/glmmTMB/R/emmeans.R
@@ -26,22 +26,21 @@
 ##' @param \dots Additional parameters that may be supported by the method.
 ##' @param grid A \code{data.frame} (provided by \code{ref_grid}) containing the
 ##' predictor settings needed in the reference grid
+##' @details While the examples below are disabled for earlier versions of
+##' R, they may still work; it may be necessary to refer to private
+##' versions of methods, e.g. \code{glmmTMB:::Anova.glmmTMB(model, ...)}.
 ##' @examples
 ##' warp.lm <- glmmTMB(breaks ~ wool * tension, data = warpbreaks)
 ##' if (require(emmeans)) {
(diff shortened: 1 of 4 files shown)


mebrooks:
The [same error](https://win-builder.r-project.org/Mwk3Qd4J3VT8) keeps occurring with the old R. I'm sure it's frustrating to debug remotely. I'll try to install an old version of R on my Windows laptop to figure out what's going on.

mebrooks:
Ugh.  I cannot replicate this error in R 3.5.3 on my Windows laptop.

bbolker:
Are you 100% sure you re-built the tarball before sending to winbuilder?

Cross-linking [the "0.2.4 checklist" issue](https://github.com/glmmTMB/glmmTMB/issues/493#issuecomment-578569564) here; this issue is almost certainly related to the conditional/different exporting of the `Anova` and `effects` methods depending on R version.  I don't know whether it interacts with `car`/`effects` versions or not, and/or the `_R_S3_METHOD_LOOKUP_BASEENV_AFTER_GLOBALENV_` environment variable.

bbolker:
Ugh.  I built the tarball myself to check, and got the same results.

I'm kind of stumped.

I don't know whether it's worse to bail out and condition these tests on `if (getRversion() >= 3.6.0)` or to leave them failing on R 3.5.3 ... I keep going round in circles about why the tests are failing. Thought about putting some debugging code in the examples (which looks like it would get printed in the win-builder log, if it was associated with a failing test or example), but I can't quite decide what to do.  It feels like `trace()` is appropriate to see which methods are being used, but I can't even figure out how it works locally, let alone hoping that it does something sensible on the remote platform.

Another option is to cry for help on `r-package-devel@r-project.org` ...

bbolker:
Where are we on this?  If you want I can implement the brute force `if (getRversion() >= "3.6.0"` approach tomorrow ...

mebrooks:
I emailed the [devel list](https://stat.ethz.ch/pipermail/r-package-devel/2020q1/004945.html) but haven't gotten any reply. Did I ask in the wrong way somehow? I'm in favor of the brute force approach you suggest.

bbolker:
I think you may have left out some details (i.e. I think this must have to do with the changes in R controlling which S3 methods are loaded/searched for where, and `_R_S3_METHOD_LOOKUP_BASEENV_AFTER_GLOBALENV_` (enabled in R [here](https://github.com/wch/r-source/commit/753ceaea9854d0b4d7a363041f0176ecc1b036fb): I have so far been able to find very little documentation on this change.

mebrooks:
I built it and tested on my computer and all win-builder versions. They were all fine [except R 3.5.3](https://win-builder.r-project.org/QZwi01oZFYs1/00check.log).
```
== Failed ======================================================================
  -- 1. Error: (unknown) (@test-downstream.R#57)  --------------------------------
  object 'fmd' not found
  Backtrace:
   1. stats::model.frame(fmd) test-downstream.R:57:0
  ```

I think these lines may also need to get protected by the if-statement
```
model.frame(fmd)
model.matrix(fmd)
```

**Comment by bbolker** [comment]
Sorry about that, I'm getting sloppy.

**Comment by mebrooks** [comment]
Now there's only a [warning on R 3.5.3](https://win-builder.r-project.org/rHWX66F4jXvL). I think that should be ok.
```
* checking re-building of vignette outputs ... [160s] WARNING
Error in re-building vignettes:
  ...
Warning in qt((1 - level)/2, df) : NaNs produced
Loading required package: carData
```

**Comment by bbolker** [comment]
Ugh, that WARNING is actually referring to this, from later on:

```
Quitting from lines 170-176 (model_evaluation.Rnw) 
Error: processing vignette 'model_evaluation.Rnw' failed with diagnostics:
subscript out of bounds
Execution halted
```

PR #550, opened 2020-02-17
Small fix for influence_mixed to work w/tibbles

jmobrien:
Errors were produced when function was run on model object whose data was classed as a tibble, as tibble subsetting uses an implicit `drop = FALSE` inside subset brackets, rather than the implicit `drop = TRUE` used for standard data.frames.

### DIFF
--- a/glmmTMB/inst/other_methods/influence_mixed.R
+++ b/glmmTMB/inst/other_methods/influence_mixed.R
@@ -114,7 +114,8 @@ influence_mixed <- function(model, groups=".case", data, maxfun=1000,
         data[, del.var] <- apply(data, 1, paste0, collapse=".")
         groups <- del.var
     }
-    unique.del <- unique(data[, groups])
+    # Explicit drop = TRUE will ensure returned as expected vector when data is classed as a tibble:
+    unique.del <- unique(data[, groups, drop = TRUE])
     data[[".groups"]] <- data[, groups]
     par <- list(theta=getME(model, "theta"))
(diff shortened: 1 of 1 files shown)


bbolker:
This seems like a good idea, thank you.  I was thinking about asking if you could write a test for this fix. I'm hesitating slightly because we would drag the `tibble` package into `Suggests:`, which drags in much of the tidyverse - on the other hand it's probably implied already by e.g. `broom` and `huxtable` ...

jmobrien:
I'm a psychologist, not a computer scientist, so I'm unclear what an adequately composed test would be.  That said, my own use case was working fine following this change.  

(Based on my experience, though, I think a simple enough test would be to compare glmmTMB models run with data sources `Owl` and `as_tibble(Owl)`, then are passed to the `influence_mixed`)

Really, I think tibbles are supposed to in almost all cases function as normal data frames; this thing with `drop` is probably the main low-level exception.  Since it's clear from context that `unique.del <- unique(data[, groups])` is expected to always be a single column that drops down to an atomic vector, perhaps it could just become  
`unique.del <- unique(data[[groups]])`.  Then tibble status or not wouldn't be relevant.

bbolker:
I think switching to `[[groups]]` is better; I'm making a PR that does that (and has some tests in the examples).  Not sure at the moment of the status of `influence.merMod` here vs in the `car` package ...

jmobrien:
That makes sense.  Closing this PR in support of your eventual one.

PR #551, opened 2020-02-17
fix influence_mixed for tibbles (jobrien)

bbolker:


bbolker:
Breaking the rules and merging my own PR

PR #558, opened 2020-03-07
Strings as factors

bbolker:


bbolker:
My fix is *slightly* dirty in that it sets `stringsAsFactors=TRUE` rather than giving specific factor ordering (so is in principle sensitive to locale-ordering issues).  Improvements welcome.

bbolker:
bending the rules to merge this (even though I'm an author)

PR #560, opened 2020-03-08
conditionally execute/store figs for covstruct vignette

bbolker:
second try (from old master, not on top of stringsAsFactors branch ...)


bbolker:
I just merged the `stringsAsFactors` branch here, so that it will pass tests with r-devel.  We should merge the other PR first ...

kaskr:
Wanted to check this PR by building the covstruct vignette.
So, I did
`make glmmTMB/vignettes/covstruct.html`
(which worked a few days ago) and got

```
  |...........................................                           |  61%
label: volcano_data_image_fake (with options) 
List of 1
 $ eval: logi FALSE

  |............................................                          |  62%
  ordinary text without R code

  |............................................                          |  63%
label: volcano_data_image_real (with options) 
List of 2
 $ echo: logi FALSE
 $ eval: logi TRUE

Quitting from lines 413-418 (covstruct.rmd) 
Error in plot.new() : 
  could not open file 'inst/vignette_data/volcano_data.png'
Calls: <Anonymous> ... image -> image.default -> plot -> plot.default -> plot.new

Execution halted
make: *** [glmmTMB/vignettes/covstruct.html] Error 1
```

bbolker:
OK, this is because the image files only exist where they're being looked for if the package has been re-installed.  The `make` rule could re-install the package or ...

kaskr:
Ah OK. Let me try again.

bbolker:
No, my diagnosis is wrong.   The problem is in saving, not reading.  Right now we're trying to save in `inst/vignette_data` relative to the current working directory, and the make rule is being run from one level above the package directory ...

bbolker:
Ugh, `rmarkdown::render()` makes it hard to tell what directory you're in.  Have to go to a meeting now, will keep working on this later unless someone else fixes it first ...

bbolker:
OK, please try again. (`make ...` works for me now)

PR #569, opened 2020-03-28
fix na.exclude bug in residuals

bbolker:


bbolker:
got impatient.  Should remember to add this in NEWS too ...

PR #582, opened 2020-05-02
allow for small im parts in eig(cov)

bbolker:


bbolker:
is it worth adding a "skip eigenvalue check" flag to `glmmTMBControl()`, so we can avoid the computational cost for large problems?  (Much more fundamentally: should/could we allow an option for sparse fixed-effect model matrices? Easy to do on the R side.  I don't know how much of a pain it is on the C++ side to offer that alternative.)

bbolker:
redundant with sparse_X

PR #583, opened 2020-05-03
Sparse x

bbolker:
This might not be done yet, but I think it will potentially be  useful.  

- Have only done very very basic tests, only on sparse-X for the conditional model. The structure for sparse-X is there for cond/zi/disp, in what seems like a sensible way.  If there's a better (more TMB-idiomatic) way to do it, I'd be happy to hear it.
- Is there a general preference for C++/TMB style?  The internal implementation is [here](https://github.com/glmmTMB/glmmTMB/blob/sparse_X/glmmTMB/src/glmmTMB.cpp#L641-L658); I was originally using a trinary operator, could go back to that ...


bbolker:
Travis checks are failing right now with
```
$ tlmgr install chicago
tlmgr: verification of checksum for http://ftp.math.purdue.edu/mirrors/ctan.org/systems/texlive/tlnet failed: cryptographic signature verification of /tmp/Ig7_q63BEi/6aO0GY2ok5 against
http://ftp.math.purdue.edu/mirrors/ctan.org/systems/texlive/tlnet/tlpkg/texlive.tlpdb.sha512.asc
failed. Output was
gpg: Signature made Fri 01 May 2020 11:58:00 PM UTC using RSA key ID 19438C70
gpg: BAD signature from "TeX Live Distribution <tex-live@tug.org>"
Please report to texlive@tug.org
```

bbolker:
Another point here: I've added an `eigval_check` to `glmmTMBControl()`.  If we think we will want to add more fine-scaled control of the post-fitting tests and checks we may want to do something parallel to @mmaechler's system for `lme4` where users can specify whether to skip  the tests completely ("ignore" is what lme4 uses but maybe "skip" would be better?), throw a warning, or throw an error, and what tolerance to use for the tests ...

bbolker:
Leaving out `tlmgr` also breaks, looks like we'll have to wait until that problem resolves itself (or report it ...)

In the meantime, another question (@kaskr?): can we make previously stored objects work OK?  Here's an example of reloading a fit 

```
L <- load(system.file("vignette_data","model_evaluation.rda", package="glmmTMB"))
library(glmmTMB)
VarCorr(owls_nb1)
```
> Error in retape(set.defaults = FALSE) : 
  Error when reading the variable: 'XS'. Please check data and parameters.
In addition: Warning messages:
1: In retape(set.defaults = FALSE) :
  Expected sparse matrix of class 'dgTMatrix'.
2: In retape(set.defaults = FALSE) : Expected object. Got NULL.

It looks like this happens here

```
 if (isNullPointer(ADFun$ptr)) {
    if (silent) 
        beSilent()
    retape(set.defaults = FALSE)
}
```
This is presumably because the object got stored without some piece of machinery, and is now trying to use that machinery from the new code (which has an extra data slot)?

kaskr:
There are two ways to make previously saved objects work with new compiled code (with added data reads). The first most obvious is to load the old object and modify it as:

```r
L$obj$env$data$Xs <- ...
```

The second option would be to read and use the sparse matrices conditionally on something we already have in the object. Perhaps we could use zero sized dense matrices rather than a new flag to signify that sparse matrices should be read and used:

```cpp
if (X.rows()==0 && X.cols()==0 && ...) {
  DATA_SPARSE_MATRIX(Xs);
  mu += Xs * beta;
}
```

This ensures that old saved objects will never enter the sparse matrix code block.

bbolker:
Second solution is great, as I've already got a zero-sized dense matrix there. May not get to it today, but that seems like the right way to go.

bbolker:
This looks OK now (and supersedes the min_eig PR).  The only remaining question I have is:

> is it worth adding a "skip eigenvalue check" flag to glmmTMBControl(), so we can avoid the computational cost for large problems? (Much more fundamentally: should/could we allow an option for sparse fixed-effect model matrices? Easy to do on the R side. I don't know how much of a pain it is on the C++ side to offer that alternative.)

bbolker:
bump.  I'm pretty happy with this (with the possible exception of setting up a better interface for allowing the user to adjust checks); can we discuss and/or merge it?

kaskr:
Alternative to 'skip eigenvalue check flag' could be a 'skip eigenvalue check tolerance' on matrix size, perhaps with a warning when check is skipped. A reasonable tolerance could be 1000.

Issue #596, opened 2020-06-15
Documentation of `emmeans()` support


rvlenth:
Just a comment that some users seem to be confused by the way that `emmeans()` support is provided in **glmmTMB**. See for example the query on Stack Overflow: https://stackoverflow.com/questions/62351158/contrasts-with-zero-inflated-glmmtmb/62392249#62392249. This user seems to think you need to call `emm_basis.glmmTMB()` -- and that is somewhat understandable since it is shown in the "usage" section. 

I don't believe that `emm_basis.glmmTMB` or `recover_data.glmmTMB` even need to be documented, since these functions are not exported (only registered as methods if **emmeans** is loaded). Instead, it would help to make it clearer that users need to to call `emmeans(..., component = "cond")` or `emmeans(..., component = "zi")`. 

I think the Details section is misleading, at least in the case of emmeans support, because it advises users to call those methods when they decidedly should not. And it refers to "examples below" when in fact no examples are given at all.

bbolker:
Does this look better? https://github.com/glmmTMB/glmmTMB/blob/emmeans_doc_fix/glmmTMB/man/downstream_methods.Rd

rvlenth:
Thanks. Looks much better to me!

Russ

On Jun 15, 2020, at 1:24 PM, Ben Bolker <notifications@github.com> wrote:

﻿

Does this look better? https://github.com/glmmTMB/glmmTMB/blob/emmeans_doc_fix/glmmTMB/man/downstream_methods.Rd

Issue #579, opened 2020-04-22
Spatial structure in hurdle glmmTMB


mairekirkland:
I am trying to use the glmmTMB package (version 1.1.0) to run a hurdle model that includes a spatial covariance structure. However, when I try to include spatial information in the zero-inflated part of the model I get an error. 

Here is a reproducible example with the error message that I receive: 

```
library(glmmTMB)

d <- data.frame(count = as.vector(Salamanders$count),
               mined = as.vector(Salamanders$mined),
               x = as.vector(row(Salamanders)),
               y = as.vector(col(Salamanders)))

d$pos <- numFactor(scale(d$x), scale(d$y))
d$ID <- factor(rep(1, nrow(d)))

m <- glmmTMB(count ~ mined + mat(pos + 0 | ID), 
               zi = ~., 
               family = truncated_poisson, data = d
)

```

"Error in Reduce(addForm0, list(...)) : object 'mined' not found”

amynang:
I have the same issue using an ar1 term to account for correlation between repeated samplings. Works fine for the conditional model but gives the error mentioned above when included in the ziformula. Looks like ziformula currently (1.0.1) does not accept correlation structures?

amynang:
@mairekirkland A quick workaround for this would be to run the two components as individual models. Your ziformula is a logistic regression.

mebrooks:
This seems like a bug. @kaskr do you have any idea what's happening?

bbolker:
I don't think this is @kaskr's fault.  It looks like it's in the formula-manipulation code somewhere around [here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/glmmTMB.R#L683-L690).  I don't know exactly what's going on, but `noSpecials(subbars(f),delete=FALSE)` is returning an expression without a leading tilde, which surprises me/looks funny:  `mined + (pos + 0 + ID)`. Maybe it isn't properly designed for one-sided formulas?

bbolker:
[This branch](https://github.com/glmmTMB/glmmTMB/tree/nospecials_fix) fixes the proximal problem (`remotes::install_github("glmmTMB/glmmTMB/glmmTMB@nospecials_fix")` should install this version), but ... the proposed example crashes - I think out of memory (has to process two 644x644 dense matrices?) not sure if the problem is pathological/misspecified in some way or if there's another bug?

bbolker:
[...]

PR #598, opened 2020-06-16
make noSpecials work with one-sided formulas; add tests

bbolker:
based on issue #596


mebrooks:
This looks good, but I think you might mean issue #579  unless I'm missing something.

PR #604, opened 2020-06-30
fix doc links to other packages

mebrooks:


mebrooks:
While I was already changing docs, I made a few small updates to the into vignette.

bbolker:
I'm surprised that these issues never showed up before ... but thanks for fixing them!

PR #605, opened 2020-07-03
Better pdcheck

bbolker:
This isn't quite ready for prime time, but I thought I would create a pull request as a venue for discussion. This PR 

* improves on Hessian/SD computations. If the built-in Hessian check from `TMB::sdreport` fails (i.e. `fitted_model$sdr$pdHess` is `FALSE`), then we re-do the computation by using `numDeriv::jacobian()`. This uses Richardson extrapolation and is a lot slower than the `sdreport` version (which uses `optimHess()` = finite differences on the gradient function), but is more accurate.  We *might* want to add a "never do this" switch, but it's hard to see why someone would want this - it only kicks in if the faster method is flagging a problem
* adds a `diagnose()` function that does a lot of standard checks and tries to provide detailed information, specifically (1) checking for parameters with large magnitude, (2) checking for predictors with small (but non-zero) or large ranges, (3) identifies which eigenvalues of the Hessian are problematic, and which associated eigenvector elements are non-trivial.

The good news/bad news is that because the results are considerably more robust, some parts of the `troubleshooting` vignette are now out of date, e.g.

* models that we had as examples of non-pos-def Hessians are no longer problematic;
* an example that previously gave a gradient-length error now returns a more sensible "some responses of a NB model are negative" error
* in general, the document can probably be updated and streamlined using `diagnose()`.

(Note to self: I was getting frustrated with building the `troubleshooting` vignette, I had forgotten about the whole `make vignette-update` -> move to `doc/` thing ...)

I'm locally having a lot of trouble with building the `model_evaluation.Rnw` vignette (I don't know what's going on: I'm getting errors that seem to be `huxtable`/LaTeX related, even if I cut all of the huxtable stuff out of the vignette ...), but that doesn't seem to be a problem on Travis.

I would very much like more eyeballs/opinions both on the interface and implementation of `diagnose()`, and on the discussion in `troubleshooting` ...


mebrooks:
I'm starting to look at this, but it will take more time. I can report that `make install` gave the errors/warnings below, while RStudio's "Install and Restart" seemed to work fine. 

```
--- re-building ‘troubleshooting.rmd’ using rmarkdown
Quitting from lines 226-229 (troubleshooting.rmd) 
Error: processing vignette 'troubleshooting.rmd' failed with diagnostics:
object 'zi1' not found
--- failed re-building ‘troubleshooting.rmd’

--- re-building ‘glmmTMB.Rnw’ using knitr
Warning in qt((1 - level)/2, df) : NaNs produced
Warning in max(ids, na.rm = TRUE) :
  no non-missing arguments to max; returning -Inf
--- finished re-building ‘glmmTMB.Rnw’
```

I've looked at the vignette and can't see any reason for the error (e.g. caching complications).

bbolker:
I think the vignette zi1 issue has to do with the fact that on line 226 there's an inline chunk `r round(zi1,1)` that refers to `zi1`; if the environment variable `NOT_CRAN` hasn't been set, then none of the code chunks get evaluated.

I *may* have solved that problem by judicious forcing of `eval=TRUE` in appropriate chunks ... ?

mebrooks:
This seems good to me. @kaskr, what do you think?

kaskr:
I like the idea of a diagnose function. It's obviously already useful, although one can always find edge cases:

```
example(glmmTMB)
diagnose(m0)
```

The diagnosis (wrongly) suggests a potential problem with the beta parameters, but it doesn't see the actual problem (the theta parameter has huge Std. Error and Wald CI is inaccurate):

```
> m0$sdr
sdreport(.) result
        Estimate   Std. Error
beta  200.034308   0.09727646
beta  -99.714909   0.21753251
betad   3.202870   0.02773501
betad   1.386476   0.03922322
theta  -7.508328 349.79842687
Maximum gradient component: 1.611173e-05
```

```
> confint(tmbprofile(m0$obj,"theta"))
      lower      upper
theta    NA -0.9443034
```

bbolker:
We should probably skip the beta-magnitude check when using an identity link!  (It's still not impossible to get large `|beta|` if the predictor variable itself is weirdly scaled; if we have access to `X` (fixed-effect matrix) we can do a better job with this.)

mebrooks:
I lazily checked this by installing, running the standard tests, and running the example code that @kaskr posted above. I think it's fine to merge now, and we can keep thinking about ways to improve diagnostics of edge cases.

bbolker:
@kaskr, `diagnose` correctly diagnoses your example now (turned off magnitude check for identity-link problems; added a check for large Z-statistics).  See what you think of the warning message:

> Large Z-statistics (estimate/std err) suggest a failure of the Wald
approximation - often also associated with parameters that are at or
near the edge of their range (e.g. random-effects standard deviations
approaching 0).  While the Wald p-values and standard errors listed in
summary() are unreliable, profile confidence intervals (see
?confint.glmmTMB) and likelihood ratio test p-values derived by
comparing models (e.g. ?drop1) may still be OK.  (Note that the LRT is
conservative when the null value is on the boundary, e.g. a variance or
zero-inflation value of 0 (Self and Liang 1987; Stram and Lee 1994;
Goldman and Whelan 2000); in simple cases the p-value is approximately
twice as large as it should be.)

bbolker:
Other than incorporating `diagnose()`  in the "troubleshooting" vignette, maybe this is ready to merge?

PR #609, opened 2020-07-07
fix summary for models with mapping

bbolker:
Main question about this one: should covariances and variances of a mapped parameter be set to 0 or `NA`?

* zero: makes sense, will probably propagate through downstream computations in a sensible way
* `NA`: do we want to force users to make more explicit decisions about what they do with uncertainties of a mapped parameter?

There could be an option or  a switch as well, but probably not worth the trouble ...

At present it returns 0, if no-one comments then that's how it will go in ...


mebrooks:
Zero seems fine to me. I don't have an opinion. Anybody else?

bbolker:
Let's go ahead and merge?

Issue #612, opened 2020-07-10
Offsets break glmmTMB when called via do.call


cvoeten:
```
library(glmmTMB)
ss <- lme4::sleepstudy
off <- rnorm(nrow(ss),10,20)
glmmTMB(Reaction ~ Days,ss,offset=off)
do.call(glmmTMB,list(Reaction ~ Days,ss,offset=off))
```
The first call works completely fine, the second gives:
```
Error in vapply(attr(tt, "variables"), deparse, character(1)) : 
  values must be length 1,
 but FUN(X[[3]]) result is length 46
```
R 4.0.2 on Arch Linux.

From the traceback, it seems that the problem may be related to unsafe use of `deparse` in `fix_predvars`. Replacing `tt_vars <- vapply(attr(tt, "variables"), deparse, character(1))[-1]` with `tt_vars <- vapply(attr(tt, "variables"), function(x) as.character(list(x)), character(1))[-1]` gets it a little further, but then the next problem is `Error in getXReTrms(formula, mf, fr, contrasts = contrasts, sparse = sparseX[["cond"]]) : trouble reconstructing offset name`, at which point I gave up.

bbolker:
Thanks, I'll take a look.  I wish that `lm` hadn't set the original convention that offsets could be specified *either* as elements of the formula *or* as separate terms; it seems to cause so much complexity and fragility ...

bbolker:
I think this should be fixed now: if you want to try it out from the branch, `remotes::install_github("glmmTMB/glmmTMB/glmmTMB@do_call_ofset")` should install it (yes, I did spell offset with only one f ...)

cvoeten:
That indeed fixes the original issue, but a small change brings to light a second problem:
```
library(glmmTMB)
ss <- lme4::sleepstudy
ss <- rbind(ss,ss,ss,ss,ss,ss,ss,ss,ss,ss)
off <- rnorm(nrow(ss),10,20)
glmmTMB(Reaction ~ Days,ss,offset=off) #works
do.call(glmmTMB,list(Reaction ~ Days,ss,offset=off)) #breaks
```
```
Error in terms.formula(object, data = data) : 
  variable names are limited to 10000 bytes
```
Sorry...

bbolker:
This one is harder.  The proximal problem is that (in order to avoid downstream problems not being able to find the variable passed as an offset) we evaluate the `offset` argument before proceeding. In the case of `do.call()

[...]

PR #613, opened 2020-07-10
use safe_deparse for offsets

bbolker:
addresses #612


mmaechler:
hmm.. within glmmTMB this was saveDeparse(), right?

but have you seen that someone (you know well) has introduced a slightly generalization of saveDeparse() into R 4.0.0,
called  deparse1() ?
I strongly recommend using that (and provide it for earlier R versions).

Issue #571, opened 2020-04-01
confint


monicachaudhari:
Issue: confint() does not return CIs when  dispformula is included in the model and throws an error shown below:

```
glmmformula <- formula(Y ~  sex + agestd + ethnicity + batch + disease*time + (1|donor)) 
 modl <- glmmTMB::glmmTMB(glmmformula
                              , dispformula=~ sex + agestd +  batch + disease 
                              , family=glmmTMB::beta_family(link = "logit")
                              , data = tmpdat)
confint(modl,full=TRUE )
```

> Error in dimnames(x) <- dn : 
>  length of 'dimnames' [2] not equal to array extent

mebrooks:
I can't work on this at this moment, but I agree and here is a repeatable example
```
## Dispersion model
sim1 <- function(nfac=40, nt=100, facsd=0.1, tsd=0.15, mu=0, residsd=1)
{
  dat <- expand.grid(fac=factor(letters[1:nfac]), t=1:nt)
  n <- nrow(dat)
  dat$REfac <- rnorm(nfac, sd=facsd)[dat$fac]
  dat$REt <- rnorm(nt, sd=tsd)[dat$t]
  dat$x <- rnorm(n, mean=mu, sd=residsd) + dat$REfac + dat$REt
  dat
}
set.seed(101)
d1 <- sim1(mu=100, residsd=10)
d2 <- sim1(mu=200, residsd=5)
d1$sd <- "ten"
d2$sd <- "five"
dat <- rbind(d1, d2)
m0 <- glmmTMB(x ~ sd + (1|t), dispformula=~sd, data=dat)
confint(m0)
```

monicachaudhari:
Surprisingly, it does not recognize the function when called by  glmmTMB::confint(). I suspect, it is calling stats::confint() and hence the error.

bbolker:
No, it's really our problem. `glmmTMB:::confint.glmmTMB(m0)` replicates what's actually happening (S3 methods like `confint.glmmTMB` don't have to be exported in order to be found by R's method dispatch machinery.

ash-pathak:
Hello folks. Any updates on this issue? Thanks.

bbolker:
not yet.  Spent some time poking around today.  May have localized it to the behaviour of the internal `.CI_univariate_monotone` function, but still trying to figure out what's going on.

bbolker:
[...]

PR #622, opened 2020-08-30
Confint fix

bbolker:
Responding to #571

### DIFF
--- a/glmmTMB/R/methods.R
+++ b/glmmTMB/R/methods.R
@@ -827,6 +827,7 @@ confint.glmmTMB <- function (object, parm = NULL, level = 0.95,
         cf <- cf[nn]
         ss <- diag(vv)
         ## using [[-extraction; need to add component name explicitly
+        ci.tmp <- NULL
         if (length(cf)>0) {
             names(cf) <- names(ss) <-
                 paste(component, names(cf), sep=".")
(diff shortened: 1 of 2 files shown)


mebrooks:
Thanks for working on this. I only had time to look at it a little and found that `uniroot` and `profile` don't work for intercept only models.  I never tried this with the old version of `confint`, so I don't know if this is different. 

```
library(glmmTMB)

mu=3
n=10000
y0=rpois(n, mu)
pdat=data.frame(y=y0[y0>0]) 
m1=glmmTMB(y~1, family=truncated_poisson(), pdat)

confint(m1)
confint(m1, method="uniroot") #NA for 2.5 % and 97.5 %
confint(m1, method="profile") 
#Error in split.default(x = seq_len(nrow(x)), f = f, drop = drop, ...) : 
#  group length is 0 but data length > 0
#In addition: Warning message:
#In min(dd0$value, na.rm = TRUE) :
#  no non-missing arguments to min; returning Inf

pdat$x=runif(n=nrow(pdat))
m2=glmmTMB(y~x, family=truncated_poisson(), pdat)
confint(m2)
confint(m2, method="uniroot") #ok
confint(m2, method="profile") #ok
```

bbolker:
@mebrooks , the problems you show above with single-parameter models are (1) pre-existing and (2) go back to `TMB` functions (see https://github.com/kaskr/adcomp/pull/324 )

bbolker:
I think this should be ready to merge? The issue highlighted by @mebrooks above has been fixed in TMB in b96566332fc368

Issue #623, opened 2020-09-01
AR1 models fail to converge


maxcastorani:
I am trying to utilize the first-order autocorrelation [AR(1)] covariance structure abilities of the glmmTMB package to model experimental time series data collected from multiple locations. However, the models consistently fail to converge, with the warning message: In fitTMB(TMBStruc) :  Model convergence problem; non-positive-definite Hessian matrix. See vignette('troubleshooting').

To diagnose the problem, I created some simulated AR1 data (see code below). However, the failure-to-converge problem persists even when varying the time series length, data distribution and link function, dispersion, zero inflation, and model random effects structure.

Is this a problem with my syntax, a data issue, or a glmmTMB bug? Many thanks for the assistance.

```
# Simulation test of AR(1) time series using glmmTMB

# Install or load packages

#install.packages('bbmle')
#install.packages('dplyr')
#install.packages('ggplot2')
#install.packages('patchwork')
#install.packages('Matrix')
#install.packages('devtools')
#library('devtools')
#devtools::install_github("glmmTMB/glmmTMB/glmmTMB")

library(bbmle)
library(dplyr)
library(ggplot2)
library(glmmTMB)
library(patchwork)

# Set parameters
t = 40       # Length of time series
phi = 0.8    # Amount of autocorrelation
stdev = 0.1  # Standard deviation
eff <- 3    # Effect of a treatment relative to a control

# Simulate data
sim.dat <- expand.grid(time = 1:t,
                      treatment = c("control", "manipulate"),
                      location = c("site1", "site2", "site3", "site4", "site5"))
sim.dat$plot <- paste(sim.dat$location, sim.dat$treatment, sep = "_")

# List description for AR(1) model
ar.sim <- list(order = c(1, 0, 0), ar = phi, sd = stdev)

# Simulate AR(1) data for each plot (= unique combination of location and treatment)
set.seed(1);  ar1.sim_site1.control      <- arima.sim(n = t, model = ar.sim)
set.seed(2);  ar1.sim_site2.control      <- arima.sim(n = t, model = ar.sim)
set.seed(3);  ar1.sim_site3.control      <- arima.sim(n = t, model = ar.sim)
set.seed(4);  ar1.sim_site4.control      <- arima.sim(n = t, model = ar.sim)
set.seed(5);  ar1.sim_site5.control      <- arima.sim(n = t, model = ar.sim)
[...]

PR #624, opened 2020-09-04
store terms information, check that ar1() time is a factor

bbolker:
addresses #623

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ glmmTMB/vignettes/figure
 glmmTMB/vignettes/cache
 *~
 *#*
+glmmTMB*.tar.gz
+glmmTMB.Rcheck
 
 glmmTMB/vignettes/*_cache
 glmmTMB/vignettes/*_files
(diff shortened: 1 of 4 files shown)


kaskr (commented):
A few comments:

I think it makes sense to thow warning in cases `ar1(t|group)` and `ar1(t+0|group)` etc when `t` is numeric.

However, we should perhaps check that 'random regression with correlated slope', i.e `ar1(f:t+0|group)` works without warning where `f` is a factor and `t` is numeric?

Test for factor is useful. Another issue is whether levels are equally spaced which is hard to test for. But at least we could make it part of the warning message that `levels(f)` has requirements?

bbolker:
I haven't thought about it too carefully: does `ar1(f:t+0|group)` do something sensible?

The current check is 

```
if (length(.getXlevels(reXterms[[i]],fr))!=1) {
```

i.e., it's looking for exactly one factor variable involved. `ar1(f:t+0|group)` does pass this check.

At the moment any number of factors != throws an error. Should it be a warning instead, or user-controllable?

bbolker:
should be ready to merge, unless we want to change the error to a warning or something else?

Issue #620, opened 2020-08-19
parallel issues


bbolker:
I think `glmmTMB` by default uses as many cores are available to run OpenMP threads

- this is a dangerous default on large, multi-user machines (e.g. see [this twitter thread about the cellranger package](https://twitter.com/henrikbengtsson/status/1295862941682511874)) {I'm not sure why this hasn't caused us problems on CRAN yet?}
- I'm not even sure this is efficient; I see CPU usage go up to 750% (I have an 8-core machine) and I don't think 

```
data("InstEval",package=lme4)
library(glmmTMB)
system.time(m1 <- glmmTMB(y ~ service + lectage + studage + (1|d) + (1|s), data=InstEval))
system.time(m2 <- update(m1, control=glmmTMBControl(parallel=1))
```

In this case `m1` seems to use 2 cores, but the elapsed time is basically the same! (Maybe due to memory constraints?)

We should check the default behaviour, and see when/how much OpenMP parallelism actually helps.  This might be a good student project ...

strengejacke:
I also have a question related to the parallel options: I have fit a model to a larger data set. Using `control=glmmTMBControl(parallel=4)` took the model ~ 69 seconds to fit, and using `control=glmmTMBControl(parallel=1)` took ~70 seconds, so literally the same time. Are there specific "circumstances" when the use of parallel threads gives no or large benefits?

bbolker:
I'm really not sure.  This is why I want to get feedback from someone (@kaskr? @skaug?) who knows more about the nuts and bolts of the parallelization.  As far as I know OpenMP is shared-memory, and I wouldn't expect a lot of interference between threads (since the log-likelihood of each observation really is independent within a single likelihood evaluation step).  I don't know exactly how we would go about reasoning about this **or** profiling the TMB code to see what's going on ...

Schnickgit:
I am not sure if the parallel function is working as intended. The vignette shows is very slight improvement in the computation time that I can replicate on my Linux system using the same model. However, when monitoring the cpu usage only one core is actually under load. I tried different settings with up to `control=glmmTMBControl(parallel=16)`. Memory usage is not a limiting factor.

@bbolker suggested to use `TMB::openmp() ` to test if OpenMP is working. This does not seem to be a reliable method as the command detects all 4 cores of my windows system. The vignette does not give a lot of information in how to set up you system properly. Maybe the author (Nafis Sadat) could shed some light on this?

[...]

PR #652, opened 2020-12-11
Parallel: reduce peak memory + improve speed

kaskr:
The `parallel_accumulator` construct is useful to quickly implement parallelization in TMB. However, there's a downside in that memory peak scales linearly with number of cores. This may be OK for a few cores but for something like 16 cores it's very bad.
An alternative is to manually mark terms to be summed in parallel (as in this PR). I tested the example in the parallel vignette. Memory peak was reduced with ca `x4` using 8 cores. Run time with 8 cores was improved slightly (20%).


mebrooks:
When you say "run time with 8 cores was improved slightly (20%)", is that compared to the old way of parallelizing?

When you say "mark terms to be summed in parallel" is that done by the user or is that magically done by `PARALLEL_REGION`?

bbolker:
I have accidentally merged the master branch back into this (which I think is not terrible, although it may make the history more confusing).

In addition, I have tried to use this platform to clean up some of the parallel configuration questions (i.e. making the default number of cores 1, unless specifically requested or unless `mc.cores` option is set.

Two questions:

- if someone knowledgeable (@kaskr?) can take a look at my `omp_check()` function (in `src/utils.c` and `R/utils.R`), which is supposed to simply check whether OpenMP is available and return a logical value **without** throwing a warning, and help me figure out why it isn't available from within the package?
- the current config is `parallel=getOption('mc.cores',1)`, which is fairly standard. Should we use `mc.cores` as the option to look for, or should we use a package-specific `glmmTMB.cores` option?
-

bbolker:
PS the check failure here is due to TMB/Matrix version mismatches (fixing it in another branch now)

bbolker:
bump; if some folks could review this (@sadatnfs, @mebrooks ?) to see if it seems to work across a range of platforms?

glmmTMB/src/glmmTMB.cpp
```diff
@@ -579,6 +579,11 @@ Type allterms_nll(vector<Type> &u, vector<Type> theta,
 template<class Type>
 Type objective_function<Type>::operator() ()
 {
+#ifdef _OPENMP
```
sadatnfs, line 582: I will assume that omp_get_max_threads() will always pull in the specified threads instead of every thread in the system right? (It's been a while since I've written R / C++)

sadatnfs (approved):
Everything looks good to me. Would we ever want to have a check on whether the user is overthreading beyond what their system allows, or should we leave that to the user? 

The option here could be to check the available number of threads from the system (from somewhere beyond the OMP_NUM_THREADS environment variable) and use that as the threshold. More of a safety thing really; I feel like that will involve grepping over `lscpu` and parsing stuff out from there (https://unix.stackexchange.com/questions/218074/how-to-know-number-of-cores-of-a-system-in-linux) but not sure about Win/Mac...

glmmTMB/src/glmmTMB.cpp
bbolker: not sure ...

mebrooks:
I'm getting an error when I run `make install`.

```
E>                              ^
E> glmmTMB.cpp:587:55: error: expected ';' after expression
E>   std::cout << "no OpenMP (max_parallel_regions=1)\n;"
E>                                                       ^
E>                                                       ;
E> 14 warnings and 1 error generated.
E> make[1]: *** [glmmTMB.o] Error 1
E> ERROR: compilation failed for package ‘glmmTMB’
```

mebrooks:
I'm trying to test this, but it seems that I don't have OpenMP set up correctly. Can someone point me to directions on how to set this up?

I tried 
```
library(OpenMPController)
omp_set_num_threads(3)
```
but when I test the timing I always get a message `no OpenMP`
```
> system.time(m3 <- glmmTMB(y ~ service + lectage + studage + (1|d) + (1|s), data=InstEval, control = glmmTMBControl(parallel=3) ))
   user  system elapsed 
 86.600   2.278  95.201 
Warning message:
In fitTMB(TMBStruc) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting')
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
```

sadatnfs:
@mebrooks I'm not sure about Windows, but for Mac/Linux, I believe you'd have needed to compile R with OMP flags, something like this (on top of any other C flags). Alternatively, I think you can also edit the Makeconf file and add the `-fopenmp` flags to these variables below, and then re-build glmmTMB (the logs should show that glmmTMB.cpp is getting compiled with the flag if the Makeconf has been edited to include so).

```
## -- compiling for OpenMP 
PKG_CXXFLAGS=-fopenmp
##
## -- linking for OpenMP
PKG_LIBS= -fopenmp -lgomp 
```

bbolker:
@mebrooks as I may have mentioned elsewhere it would be worth trying `TMB::runExample("linreg_parallel")` and see if it seems to be working for you ...

mebrooks:
I broke everything by deleting XCode because I read somewhere that it isn't necessary and I'm low on memory. I'm tired now, so I'll try again tomorrow.

bbolker:
@mebrooks FWIW I believe you can install a subset of Xcode, something like https://stackoverflow.com/questions/66432178/do-you-need-xcode-to-run-the-rstan-or-brms-packages-on-rstudio-macos

mebrooks:
I've been trying all day, but I can't install R-3.6.3 from source. I'm guessing that's the way to get R to use the flags in my `etc/Makeconf` file.  I can't get past `./configure` because it keeps saying 
`configure: error: C++ preprocessor "/lib/cpp" fails sanity check`. I don't have enough space available to update my OS nor put `XCode` back after deleting it. 

I gave up and went back to installing the easy way on a Mac using `R-3.6.3.nn.pkg` from CRAN. Then I installed `TMB` and `glmmTMB` from GitHub. Now, when I run 
```
TMB::runExample("linreg_parallel")
``` 
in the end it says 
```
4 regions found.
Using 4 threads
```

If I run 
```
system.time(m3 <- glmmTMB(y ~ service + lectage + studage + (1|d) + (1|s), data=InstEval, control = glmmTMBControl(parallel=3) ))
```
in the end it says 
```
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
no OpenMP (max_parallel_regions=1)
```
Does that mean `TMB` is able to run in parallel, but `glmmTMB` is not?

bbolker:
Curious why you're testing with 3.6.3?  (4.0.x was released about a year ago ...) Not that this *shouldn't* all work on 3.6.x, but ... ?

mebrooks:
I don't normally call `g++` myself to compile. I just run the make file. If I go into the `src` directory and past in what you have, then I get an error about `fopenmp`.
```
~/Documents/glmmTMB development/glmmTMB/glmmTMB/src g++ -std=gnu++11 -I"/usr/local/lib/R/include" -DNDEBUG  -I'/usr/local/lib/R/library/TMB/include' -I'/usr/local/lib/R/library/RcppEigen/include' -I/usr/local/include  -fopenmp -DARMA_64BIT_WORD -Wno-ignored-attributes -fpic  -g -O2  -c glmmTMB.cpp -o glmmTMB.o
clang: error: unsupported option '-fopenmp'
```

I don't have any real reason for not updating to R 4.0.x. I just felt like I needed to avoid potential catastrophes more than usual. I guess now that it's more established I could try, but I'm worried about not having room for XCode on my computer. CRAN sort of makes it sound like XCode is necessary "CRAN R 4.0.0 builds and higher no longer use any custom compilers and thus this directory is no longer relevant. We now use Apple Xcode 10.1 and GNU Fortran 8.2 from https://github.com/fxcoudert/gfortran-for-macOS/releases. For more details on compiling R, please see also https://mac.R-project.org/tools/"

bbolker:
I don't call g++ myself either: what I pasted came from the output of `make` or `R CMD INSTALL` or some such.
I don't know if it's worth your while wrestling with this stuff (maybe we could recruit @fishfollower @arni-magnusson @casperwberg @mmaechler  to give it a spin?). If you do
- [this SO question](https://stackoverflow.com/questions/43555410/enable-openmp-support-in-clang-in-mac-os-x-sierra-mojave) appears to have some useful info for installing OpenMP support (`brew install libomp; brew install llvm` ?)
- you may not need full Xcode, but I think you still need [command line tools via ](https://stackoverflow.com/questions/66432178/do-you-need-xcode-to-run-the-rstan-or-brms-packages-on-rstudio-macos) `xcode-select --install` ?

sadatnfs:
Hmmm... I've been working only on Linux and I compile R from scratch where I pass the OMP flags at build time. By doing so, the flags stick on in my Makeconf, and therefore any package I install in the R session picks up the OMP flags at build time. 

That being said, I know that Macs have their own favorite compiler in Clang shipped in with xcode (default R download uses clang as the default compiler). And so if someone does change the default compiler to GNU C++ from Clang++.

Now, [I just followed this tutorial](https://ryanhomer.github.io/posts/build-openmp-macos-catalina) and I was able to link a data.table and glmmTMB installation to OMP very cleanly. I tried this on a clean R 4.0 installation (binary pkg, NOT custom build) on Big Sur and it showed the OMP flags at compile time and also at model run time. 

One thing that's not mentioned **is that any ~/.R/Makevars files you may have around will overwrite R's own Makeconf file**. Once I erased that, I got everything going with the Clang compilers and OMP. ~/.R/Makevars is supposed to override R's own etc/Makeconf, and it turns out that I monkeyed around with it a while ago, and so I was getting conflicts.

sadatnfs:
I would paste my Makeconf file now, but I don't see a file attachment option, so here's my Makeconf that includes the changes from that tutorial above (you should be able to put this into your R's Makeconf which should be located in `/Library/Frameworks/R.framework/Versions/Current/Resources/etc` (where Current symlinks to 4.0)):

```Makefile
# etc/Makeconf.  Generated from Makeconf.in by configure.
#
# ${R_HOME}/etc/Makeconf
#
# R was configured using the following call
# (not including env. vars and site configuration)
# configure  'CC=clang -mmacosx-version-min=10.13' 'CXX=clang++ -mmacosx-version-min=10.13' 'OBJC=clang -mmacosx-version-min=10.13' 'FC=gfortran -mmacosx-version-min=10.13' 'F77=gfortran -mmacosx-version-min=10.13' 'CFLAGS=-Wall -g -O3' 'CXXFLAGS=-Wall -g -O3' 'OBJCFLAGS=-Wall -g -O3' 'FCFLAGS=-Wall -g -O3' 'F77FLAGS=-Wall -g -O3' '--enable-memory-profiling' '--x-libraries=/opt/X11/lib' '--x-includes=/opt/X11/include' '--enable-R-framework' '--build=x86_64-apple-darwin17.0' 'build_alias=x86_64-apple-darwin17.0' 'PKG_CONFIG_PATH=/usr/lib/pkgconfig:/usr/local/lib/pkgconfig:/opt/X11/lib/pkgconfig'

## This fails if it contains spaces, or if it is quoted
include $(R_SHARE_DIR)/make/vars.mk

### Courtesy of: https://ryanhomer.github.io/posts/build-openmp-macos-catalina
### These are needed to allow Clang++ and OMP to work
### MAKE SURE THERE'S NO ~/.R/Makevars that are conflicting

XCBASE:=$(shell xcrun --show-sdk-path)
LLVMBASE:=$(shell brew --prefix llvm)
GCCBASE:=$(shell brew --prefix gcc)
GETTEXT:=$(shell brew --prefix gettext)

CC=$(LLVMBASE)/bin/clang 
CXX=$(LLVMBASE)/bin/clang++ 
CXX11=$(LLVMBASE)/bin/clang++
CXX14=$(LLVMBASE)/bin/clang++
CXX17=$(LLVMBASE)/bin/clang++
CXX1X=$(LLVMBASE)/bin/clang++
CXX20=$(LLVMBASE)/bin/clang++

AR = ar
BLAS_LIBS = -L"$(R_HOME)/lib$(R_ARCH)" -lRblas
C_VISIBILITY = 
# CC = clang -mmacosx-version-min=10.13
CFLAGS = -Wall -g -O3 $(LTO) -g -O3 -Wall -pedantic -std=gnu99 -mtune=native -pipe
CPICFLAGS = -fPIC
# CPPFLAGS = -I/usr/local/include
CPPFLAGS=-isystem "$(LLVMBASE)/include" -isysroot "$(XCBASE)"
# CXX = clang++ -mmacosx-version-min=10.13 -std=gnu++11
## Not used by anything in R, in particular not for the .cc.d rule
## but used via R CMD config by several packages
CXXCPP = $(CXX) -E
CXXFLAGS = -Wall -g -O3 $(LTO) -Wall -pedantic -std=c++11 -mtune=native -pipe
CXXPICFLAGS = -fPIC
# CXX11 = clang++ -mmacosx-version-min=10.13
CXX11FLAGS = -Wall -g -O3 $(LTO)
CXX11PICFLAGS = -fPIC
CXX11STD = -std=gnu++11
# CXX14 = clang++ -mmacosx-version-min=10.13
CXX14FLAGS = -Wall -g -O3 $(LTO)
CXX14PICFLAGS = -fPIC
CXX14STD = -std=gnu++14
# CXX17 = clang++ -mmacosx-version-min=10.13
CXX17FLAGS = -Wall -g -O3 $(LTO)
CXX17PICFLAGS = -fPIC
CXX17STD = -std=gnu++17
# CXX20 = clang++ -mmacosx-version-min=10.13
CXX20FLAGS = -Wall -g -O3 $(LTO)
CXX20PICFLAGS = -fPIC
CXX20STD = -std=gnu++2a
CXX_VISIBILITY = 
DYLIB_EXT = .dylib
DYLIB_LD = $(CC)
DYLIB_LDFLAGS = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(CFLAGS) $(CPICFLAGS) 
## should be used by packges for dylibs, but at the time of writing was not
DYLIB_LINK = $(DYLIB_LD) $(DYLIB_LDFLAGS) $(LDFLAGS)
ECHO = echo
ECHO_C = \c
ECHO_N = 
ECHO_T = 
F_VISIBILITY = 
## FC is the compiler used for all Fortran as from R 3.6.0
# FC = gfortran -mmacosx-version-min=10.13
FC=$(GCCBASE)/bin/gfortran
F77=$(GCCBASE)/bin/gfortran
FCFLAGS = -Wall -g -O3 $(LTO_FC)
## additional libs needed when linking with $(FC), e.g. on some Oracle compilers
FCLIBS_XTRA = 
FFLAGS = -Wall -g -O3 $(LTO_FC)
#FLIBS =  -L/usr/local/gfortran/lib/gcc/x86_64-apple-darwin18/8.2.0 -L/usr/local/gfortran/lib -lgfortran -lquadmath -lm
FLIBS=-L$(GCCBASE)/lib/gcc/9/ -lm

FPICFLAGS = -fPIC
FOUNDATION_CPPFLAGS = 
FOUNDATION_LIBS = -framework Foundation
JAR = /usr/bin/jar
JAVA = /usr/bin/java
JAVAC = /usr/bin/javac
JAVAH = /usr/bin/javah
## JAVA_HOME might be used in the next three.  
## They are for packages 'JavaGD' and 'rJava'
JAVA_HOME = /Library/Java/JavaVirtualMachines/jdk1.8.0_241.jdk/Contents/Home/jre
JAVA_CPPFLAGS = -I$(JAVA_HOME)/../include -I$(JAVA_HOME)/../include/darwin
JAVA_LIBS = -L$(JAVA_HOME)/lib/server -ljvm
JAVA_LD_LIBRARY_PATH = $(JAVA_HOME)/lib/server
LAPACK_LIBS = -L"$(R_HOME)/lib$(R_ARCH)" -lRlapack
# LDFLAGS = -L/usr/local/lib
LDFLAGS=-L"$(LLVMBASE)/lib" -L"$(GETTEXT)/lib" --sysroot="$(XCBASE)"

## we only need this is if it is external, as otherwise link to R
LIBINTL=   -Wl,-framework -Wl,CoreFoundation
LIBM = 
LIBR0 = -L"$(R_HOME)/lib$(R_ARCH)"
LIBR1 = -lR
LIBR = -F/Library/Frameworks/R.framework/.. -framework R
LIBS = -L/usr/local/lib -lpcre2-8 -llzma -lbz2 -lz -licucore -ldl -lm -liconv
## needed by R CMD config
LIBnn = lib
LIBTOOL = $(SHELL) "$(R_HOME)/bin/libtool"
## set for --enable-lto, --enable-lto=check
LTO = 
LTO_FC = 
LTO_LD = 
## needed to build applications linking to static libR
MAIN_LD = $(CC)
MAIN_LDFLAGS = 
MAIN_LINK = $(MAIN_LD) $(MAIN_LDFLAGS) $(LDFLAGS)
MKINSTALLDIRS = "$(R_HOME)/bin/mkinstalldirs"
NM = /usr/bin/nm -B
OBJC = clang -mmacosx-version-min=10.13
OBJCFLAGS = -Wall -g -O3 -fobjc-exceptions $(LTO)
OBJC_LIBS = -lobjc 
OBJCXX = clang++ -mmacosx-version-min=10.13
R_ARCH = 
RANLIB = ranlib
SAFE_FFLAGS = -Wall -g -O3 -msse2 -mfpmath=sse
SED = /usr/bin/sed
SHELL = /bin/sh
SHLIB_CFLAGS = 
SHLIB_CXXFLAGS = 
SHLIB_CXXLD = $(CXX)
SHLIB_CXXLDFLAGS = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(CXXFLAGS) $(CXXPICFLAGS) $(LTO_LD)
SHLIB_CXX11LD = $(CXX11) $(CXX11STD)
SHLIB_CXX11LDFLAGS = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(CXX11FLAGS) $(CXX11PICFLAGS) $(LTO_LD)
SHLIB_CXX14LD = $(CXX14) $(CXX14STD)
SHLIB_CXX14LDFLAGS = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(CXX14FLAGS) $(CXX14PICFLAGS) $(LTO_LD)
SHLIB_CXX17LD = $(CXX17) $(CXX17STD)
SHLIB_CXX17LDFLAGS = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(CXX17FLAGS) $(CXX17PICFLAGS) $(LTO_LD)
SHLIB_CXX20LD = $(CXX20) $(CXX20STD)
SHLIB_CXX20LDFLAGS = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(CXX20FLAGS) $(CXX20PICFLAGS) $(LTO_LD)
SHLIB_EXT = .so
SHLIB_FCLD = $(FC)
SHLIB_FCLDFLAGS = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(FCFLAGS) $(FPICFLAGS)
SHLIB_FFLAGS = 
SHLIB_LD = $(CC)
## used for add-on packages linked by C
SHLIB_LDFLAGS = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(CFLAGS) $(CPICFLAGS) $(LTO_LD)
## used for standard packages
SHLIB_LDFLAGS_R = -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress# $(CFLAGS) $(CPICFLAGS) $(LTO_LD)
SHLIB_LIBADD = 
## We want to ensure libR is picked up from $(R_HOME)/lib
## before e.g. /usr/local/lib if a version is already installed.
SHLIB_LINK = $(SHLIB_LD) $(SHLIB_LDFLAGS) $(LIBR0) $(LDFLAGS)
SHLIB_OPENMP_CFLAGS = 
SHLIB_OPENMP_CXXFLAGS = 
SHLIB_OPENMP_FFLAGS = 
STRIP_STATIC_LIB = strip -S
STRIP_SHARED_LIB = strip -x
TCLTK_CPPFLAGS = -I/usr/local/include -I/usr/local/include -I/opt/X11/include
TCLTK_LIBS = -L/usr/local/lib -ltcl8.6 -L/usr/local/lib -ltk8.6 -L/opt/X11/lib -lX11 -Wl,-weak-lXss -lXext
YACC = bison -y

## for linking to libR.a
STATIC_LIBR = #  "$(R_HOME)/lib$(R_ARCH)/libR.a"  $(BLAS_LIBS) $(FLIBS)  $(LIBINTL) -lreadline -lncurses  $(LIBS)

## These are recorded as macros for legacy use in packages
## set on AIX, formerly for old glibc (-D__NO_MATH_INLINES)
R_XTRA_CFLAGS = 
##  was formerly set on HP-UX
R_XTRA_CPPFLAGS =  -I"$(R_INCLUDE_DIR)" -DNDEBUG
## currently unset
R_XTRA_CXXFLAGS = 
## used for gfortran in R > 3.6.0
R_XTRA_FFLAGS = -fno-optimize-sibling-calls

## CXX98 is no longer supported, but packages may use it.
SHLIB_CXX98LD = @SHLIB_CXX98LD@
SHLIB_CXX98LDFLAGS = @SHLIB_CXX98LDFLAGS@

## SHLIB_CFLAGS SHLIB_CXXFLAGS SHLIB_FFLAGS are apparently currently unused
## SHLIB_CXXFLAGS is undocumented, there is no SHLIB_FCFLAGS
ALL_CFLAGS =  $(PKG_CFLAGS) $(CPICFLAGS) $(SHLIB_CFLAGS) $(CFLAGS)
ALL_CPPFLAGS =  -I"$(R_INCLUDE_DIR)" -DNDEBUG $(PKG_CPPFLAGS) $(CLINK_CPPFLAGS) $(CPPFLAGS) 
ALL_CXXFLAGS =  $(PKG_CXXFLAGS) $(CXXPICFLAGS) $(SHLIB_CXXFLAGS) $(CXXFLAGS)
ALL_OBJCFLAGS = $(PKG_OBJCFLAGS) $(CPICFLAGS) $(SHLIB_CFLAGS) $(OBJCFLAGS)
ALL_OBJCXXFLAGS = $(PKG_OBJCXXFLAGS) $(CXXPICFLAGS) $(SHLIB_CXXFLAGS) $(OBJCXXFLAGS)
ALL_FFLAGS = -fno-optimize-sibling-calls $(PKG_FFLAGS) $(FPICFLAGS) $(SHLIB_FFLAGS) $(FFLAGS)
## can be overridden by R CMD SHLIB
P_FCFLAGS = $(PKG_FFLAGS)
ALL_FCFLAGS = -fno-optimize-sibling-calls $(P_FCFLAGS) $(FPICFLAGS) $(SHLIB_FFLAGS) $(FCFLAGS)
## LIBR here as a couple of packages use this without SHLIB_LINK
ALL_LIBS = $(PKG_LIBS) $(SHLIB_LIBADD) $(LIBR) $(LIBINTL)

.SUFFIXES:
.SUFFIXES: .c .cc .cpp .d .f .f90 .f95 .m .mm .M .o

.c.o:
	$(CC) $(ALL_CPPFLAGS) $(ALL_CFLAGS) -c $< -o $@
.c.d:
	@echo "making $@ from $<"
	@$(CC) -MM $(ALL_CPPFLAGS) $< > $@
.cc.o:
	$(CXX) $(ALL_CPPFLAGS) $(ALL_CXXFLAGS) -c $< -o $@
.cpp.o:
	$(CXX) $(ALL_CPPFLAGS) $(ALL_CXXFLAGS) -c $< -o $@
.cc.d:
	@echo "making $@ from $<"
	@$(CXX) -M $(ALL_CPPFLAGS) $< > $@
.cpp.d:
	@echo "making $@ from $<"
	@$(CXX) -M $(ALL_CPPFLAGS) $< > $@
.m.o:
	$(OBJC) $(ALL_CPPFLAGS) $(ALL_OBJCFLAGS) -c $< -o $@
.m.d:
	@echo "making $@ from $<"
	@$(OBJC) -MM $(ALL_CPPFLAGS) $< > $@
.mm.o:
	$(OBJCXX) $(ALL_CPPFLAGS) $(ALL_OBJCXXFLAGS) -c $< -o $@
.M.o:
	$(OBJCXX) $(ALL_CPPFLAGS) $(ALL_OBJCXXFLAGS) -c $< -o $@
.f.o:
	$(FC) $(ALL_FFLAGS) -c $< -o $@
## @FCFLAGS_f9x@ are flags needed to recognise the extensions
.f95.o:
	$(FC) $(ALL_FCFLAGS) -c  $< -o $@
.f90.o:
	$(FC) $(ALL_FCFLAGS) -c  $< -o $@
```

glmmTMB/src/glmmTMB.cpp
kaskr: Yes, `omp_get_max_threads()` pulls the specified number of threads - *not* the system max.

kaskr:
Works fine for me.... Time to get rid of the `std::cout` stuff?

sadatnfs:
> Works fine for me.... Time to get rid of the `std::cout` stuff?

I think the only reason I'd want to keep the `cout` checks in would be to help a user debug possible threading issues, but I'm fine either way.

bbolker:
I'd like to keep that stuff in until we finish testing. Testing now on my machine ... I'm doing a bit of overkill, but for now I think I only see improvement in elapsed time to fit when I use 10 copies of the `InstEval` data set and up to about 5 cores ... everywhere else, parallelization just *increases* elapsed time.  Is that as expected/what other people are seeing? (Will keep checking, make sure I have OpenMP enabled properly, etc.)

kaskr:
A few caveats to consider when testing OpenMP speedup of TMB models:

- Be sure to benchmark examples *targeted by the OpenMP parallelization*. Use `TMB:::benchmark` for a quick diagnostics. Avoid examples with high cholesky factorization timings.
- Note that parallelization of AD is *memory bound* (not *compute bound*). Memory bandwidth is thus a significant factor to consider and don't expect perfect scalability.
- Note that TMB does [subexpression elimination](https://en.wikipedia.org/wiki/Common_subexpression_elimination) under the hood. So the data replication trick used by some examples might not increase the computational workload at all.

For example the 'InstEval' example is not great (there's a speedup of the template stuff but the cholesky dominates):

```
TMB:::benchmark(obj, cores=1:4)


                         ncores
                              1     2     3     4
  template.likelihood     0.020 0.022 0.015 0.012
  template.gradient       0.048 0.050 0.035 0.028
  template.sparse.hessian 0.008 0.009 0.009 0.009
  cholesky                0.196 0.234 0.217 0.213
```

Compare with `model3$obj` from glmmTMB parallel vignette (much better):

```
TMB:::benchmark(model3$obj, cores=1:4)

                         ncores
                              1     2     3     4
  template.likelihood     0.105 0.090 0.068 0.055
  template.gradient       0.261 0.187 0.133 0.114
  template.sparse.hessian 0.000 0.000 0.000 0.000
  cholesky                0.001 0.001 0.001 0.000
```

For this example I get:

```
system.time(model3 <- glmmTMB(formula = ydata ~ 1 + xdata + (1 | groupvar), data = datause, control = glmmTMBControl(parallel = 1)))
   user  system elapsed
 16.051   0.052  16.107

system.time(model3 <- glmmTMB(formula = ydata ~ 1 + xdata + (1 | groupvar), data = datause, control = glmmTMBControl(parallel = 8)))
   user  system elapsed
 24.128   0.201   4.083
```

bbolker:
Thanks @kaskr, that's super-useful information, I will try to dig in and make use of it to convince myself that things are OK/good on my end too.

bbolker:
Another dumb question: which steps/stages of the TMB run involve the Cholesky factorization?

glmmTMB/R/utils.R
```diff
@@ -732,6 +732,15 @@ nullSparseMatrix <- function() {
 }
 
 
+##' Check OpenMP status
+##'
+##' Checks whether OpenMP has been successfully enabled for this
+##' installation of the package. (Use the \code{parallel} argument
+##' to \code{\link{glmmTMBControl}}, or set \code{options(mc.core=[value])},
```
kaskr, line 739: `mc.cores` ?

kaskr: >  which steps/stages of the TMB run involve the Cholesky factorization?

Almost all steps: `obj$fn()` `obj$gr()` etc. In particular the newton steps and determinant calculations.

bbolker: Is there an intuition about when/for what kinds of models or data the Cholesky step will dominate?

bbolker:
@mebrooks try again? FWIW I found that different install paths gave me different results (unfortunately I don't remember which ways worked & which didn't - I just know that reinstalling at one point switched me from "working" to "not working"). `omp_check()` should now be a quick way to see if OpenMP is enabled or not.

bbolker:
OK, except for possible additions to the documentation, and testing as widely as possible, I think this is ready. I'll merge if I don't hear any objections ...

kaskr:
Before finally merging this I want to point out a potential issue using `mc.cores` rather than say `glmmTMB.cores`.

Sometimes people fit models in parallel using constructs like:


```r
mclapply(1:100, function(...)glmmTMB(...))
```

It seems to me that our current approach would use `mc.cores * mc.cores` cores to fit this? (I haven't yet tried it...)

sadatnfs:
Unfortunately that is exactly what would happen with the nested parallel contexts...

bbolker:
@sadatnfs  , good point. I'm inclined to make the change (`mc.cores`→ `glmmTMB.cores`). Any other opinions here ... ?

sadatnfs:
I think regardless, the issue would be if a user decides to still use a parallel context (foreach or mclapply) and overrides the parallel cores in the glmmTMB call. Off the top of my head, I do not know if there's a way to look at the execution context within an mclapply/foreach iteration; that would help perhaps in adding some warnings that the user is attempting to use nested parallelization.

That being said, there could obviously be use cases where the user has enough cores to train models in nested parallelism...

glmmTMB/src/Makevars
```diff
@@ -0,0 +1,7 @@
+PKG_LIBS = $(LAPACK_LIBS) $(BLAS_LIBS) $(FLIBS) $(SHLIB_OPENMP_CXXFLAGS)
+## FLIBS unneeded but harmless (?)
+# 1.2.4 Using C++11 code
+CXX_STD = CXX11
+
+# Tell Armadillo to use 64BIT_WORD, which removes the matrix size limit constraint
+PKG_CXXFLAGS=$(SHLIB_OPENMP_CXXFLAGS) -DARMA_64BIT_WORD
```
kaskr, line 7: Most of this is redundant. Can be reduced to:

```
PKG_LIBS = $(SHLIB_OPENMP_CXXFLAGS)
PKG_CXXFLAGS = $(SHLIB_OPENMP_CXXFLAGS)
```

bbolker: I thought I got this from an authoritative source, but now I can't relocate it. (Probably [here](https://www.r-bloggers.com/2017/08/setting-up-optional-openmp-support-with-rcpparmadillo/), which (a) is about using RcppArmadillo, which we don't; and (b) was probably what got me going down the `configure` rabbit hole in the first place. I'm all for simplifying if it works ...

bbolker:
Although there are still things I think we could improve (more advice/documentation on users for when parallelization is likely to be effective), and threads at #620 are not completely resolved (there may still be some weird stuff going on with crosstalk between MKL and OpenMP parallelization?), I'm thinking we should go ahead and merge this. AFAICT it's definitely better than what we've had before ...  Any final thoughts?

sadatnfs:
@bbolker i can guarantee that MKL threading is itself a garbage fire (I haven't used R in almost 2 years now... Oof). But yeah unless users fudge around with even deeper settings under the hood, I personally think we have enough of a production ready state.

PR #654, opened 2020-12-22
Extend emmeans

bbolker:
fairly small changes


bbolker:
bump: can we merge this? (I don't think the Travis test failures are a real problem. If we are concerned about the tests I can try to backport the GitHub Actions stuff and test it.  It passes tests locally on my machine.)

PR #656, opened 2020-12-24
Fix path issues in R-CMD-check action

amoeba:
Changes use of `working-directory` and tweaks arguments to some steps to support checking the package in its subfolder


bbolker:
thank you!!

PR #658, opened 2021-01-05
update null-matrix args for Matrix update

bbolker:
`Matrix::sparseMatrix` args will be changing in an upcoming version, leading to lots of warnings (`giveCSparse` is deprecated in favour of `repr="T"`). This version (which preserves back-compatibility) may be overkill, we could also wait until the new version is out and impose a package version dependency ...


mebrooks:
Back compatibility sounds better to me, expecially since you already implemented it in this branch. I tested this on my computer and think it's ok to merge unless we want to wait for @mmaechler or @kaskr.

bbolker:
I'd say go ahead & merge

Issue #653, opened 2020-12-17
A closer look at dispformula=~0 - never needing the R-side in gaussian mixed models again?


SchmidtPaul:
In the plant sciences and related fields we often deal with linear mixed models with non-iid variance structures: 

- sometimes for random effects (on the G-side) 
- but more often for the residual (R-side). 

Therefore, in the past I was forced to use `nlme` with its `weights =` argument and basically hoped that my model would not get too complex. If it did get too complex, I switched to the [`asreml`](https://asreml.kb.vsni.co.uk/wp-content/uploads/sites/3/2018/02/ASReml-R-Reference-Manual-4.pdf) package or even SAS' `PROC MIXED` and its [`REPEATED`](https://documentation.sas.com/?cdcId=pgmsascdc&cdcVersion=9.4_3.3&docsetId=statug&docsetTarget=statug_mixed_syntax14.htm&locale=en) argument. Both of them are certainly great, but not open source. I was happy to learn about `glmmTMB`'s ability to deal with variance structures while giving me the advantages of `lme4`. I was **really** happy when I found this in the documentation:

> In Gaussian mixed models, `dispformula=~0` fixes the residual variance to be 0 (actually a small non-zero value: at present it is [set to sqrt(.Machine$double.eps))](https://github.com/glmmTMB/glmmTMB/blob/2b14a42bd55cd0cfeebad1f4eb7a3b2313e5d359/glmmTMB/R/glmmTMB.R#L85), forcing variance into the random effects.

Fixing the residual variance to be 0 and then *mimicing* it with an additional random effect should theoretically lead to identical results. In addition, I may now use all the variance structures with this *pseudo error term*. So all the problems are solved! What is this issue about then? It's basically this:

##  How well does `dispformula=~0` really do the trick? What are the caveats and is there more info/citeable proof somewhere of how well it works?

Here is an example of how the pseudo-error-approach can lead to obtaining essentially identical variance component estimates and AICc values.

``` r
library(agridat)
library(AICcmodavg)
library(glmmTMB)
library(mixedup)
library(tidyverse)

dat <- agridat::mcconway.turnip %>% 
  as_tibble() %>% 
  mutate(unit = 1:n()) %>% 
  mutate_at(vars(density, unit), as.factor)

m1 <- glmmTMB(yield ~ gen*date*density + 
                (1 | block), 
              dispformula = ~ 1, # default i.e. iid error variance
              REML = TRUE,
              data = dat)

[...]

PR #659, opened 2021-01-06
Dispzero flex

bbolker:
add a parameter to `glmmTMBControl` to allow setting of the minimal constant for `dispformula=~0` case


bbolker:
Bump.  If we merge this we can close #653 (although I should also add to the `NEWS` file, although I might wait and do that in the master branch, as `NEWS` updates in different branches tend to lead to minor-but-annoying merge conflicts.

bbolker:
thanks!

PR #667, opened 2021-02-13
Tmb check

bbolker:
This adds a test to see if the current installed version of TMB is more recent than the one that glmmTMB was built with. (The test could perhaps be back-ported to `TMB`, it is a generalization of the tests that TMB uses to determine whether a newer version of `Matrix` has been installed.)

As far as comments go, I'd be most interested in someone checking [the new help page](https://github.com/glmmTMB/glmmTMB/blob/TMB_check/glmmTMB/man/reinstalling.Rd) (`?reinstalling`/) and see if it seems accurate and useful.

I haven't actually tried a 'live' test (install older version of TMB; re-install glmmTMB; install newer version of TMB and make sure that the warning is appropriately thrown) yet ...


bbolker:
Not sure why this is failing with a TMB mismatch check??

kaskr:
```
glmmTMB was built with TMB version 1.7.19
Current TMB version is 1.7.21
```

It's using an old cached glmmTMB build - why? The whole idea of the CI is to rebuild the package every time...

kaskr:
I think you checked the TMB package version into the repository - was that intended?

glmmTMB/inst/TMB-version
```diff
@@ -0,0 +1 @@
+1.7.19
```
kaskr, line 1: I don't think this should be in the repository.

glmmTMB/R/utils.R
```diff
@@ -213,9 +213,69 @@ nullSparseMatrix <- function() {
         do.call(Matrix::sparseMatrix, c(argList, list(repr="T")))
     }
 }
+    
+## Check for version mismatch in dependent binary packages
+#' @importFrom utils packageVersion
+checkDepPackageVersion <- function(dep_pkg="TMB",this_pkg="glmmTMB",write_file=FALSE) {
+    cur_dep_version <- as.character(packageVersion(dep_pkg))
+    fn <- sprintf("%s-version",dep_pkg)
+    if (write_file) {
```
kaskr, line 222: I think the file should be written at install time. It seems to me it never is (`write_file=FALSE`)?

bbolker:
I think you're probably right about all of this. I'll need to get back and check/see what's going on.  It should be stored at the time the package is **built** (which is not necessarily when it is installed, e.g. if someone installs from a binary then this should be the version of TMB/Matrix/etc. that the glmmTMB .so files were compiled against ...)

kaskr:
Yes, I meant to say **built**

bbolker:
Have implemented this in what *might* be the right way (see `src/Makevars` and the appropriate part of *Writing R Extensions* (search for "phony")), but I'm not going to stay up to see if it passes checks or not.

bbolker:
OK, I think this is ready to merge ...

kaskr:
Not a big deal, but personally I had preferred to get the warning *once* when doing `library(glmmTMB)` rather than 6 times after running  `example(glmmTMB)`...
I also don't like the fact that every model fitting now requires a disk access (again, not a big deal - had just expected something closer to zero overhead):

```
> system.time(glmmTMB:::checkDepPackageVersion())
   user  system elapsed 
  0.000   0.002   0.008
```

In any case it would be easy to fix by moving the version check somewhere else - I'll leave it to you to decide.

bbolker:
That makes sense, I'll see what I can do.

bbolker:
Seems easy/sensible to move the `checkDepPackageVersion()` call to `.onLoad` (in `zzz.R`) ?

kaskr:
Right, that would be my preference too - as in https://github.com/kaskr/adcomp/blob/master/TMB/R/zzz.R

Issue #643, opened 2020-11-30
Name choice "overdispersion parameter" potentially confusing for users?


florianhartig:
Hi guys,

I just answered a DHARMa question about overdispersion https://github.com/florianhartig/DHARMa/issues/217 , and I suspect that the reason for the misunderstanding is that dispersion parameters in glmmTMB are called "overdispersion parameters". It seems that people misinterpret this as the output saying that the respective model is overdispersed (I had similar questions before). 

Was there are special reason for deviating from the term "dispersion parameter", that is used in other packages? Personally, I find this much more neutral.

Best,
Florian

mebrooks:
Yes, I agree. People get confused about this pretty often. In most of the documentation (e.g. `?sigma`) we refer to it as a dispersion parameter.  To prevent "overdispersion" from showing up in `print` and `summary`, I think the main change would be in the [print.Dispersion](https://github.com/glmmTMB/glmmTMB/blob/45916c45475cff18e0bf358eec2225f5325beaea/glmmTMB/R/methods.R#L537) function, but also there would be a few small changes in tests and docs.

bbolker:
Agree that we should globally substitute "dispersion" for "overdispersion".  (Arguably "overdispersion" is *correct* for the negative binomial k/theta parameter, which can only be positive and can only represent overdispersion, but it would also be OK to call it "dispersion" IMO.)

I think this is a complete list:

```
$ find . -type f -exec grep -il overdispersion {} \;
./glmmTMB/inst/doc/troubleshooting.html
./glmmTMB/man/sigma.glmmTMB.Rd
./glmmTMB/R/VarCorr.R
./glmmTMB/R/methods.R
./glmmTMB/tests/testthat/test-methods.R
./misc/salamanders.rmd
```

PR #668, opened 2021-02-15
Change "Overdispersion" to "Dispersion"

mebrooks:
I didn't change `./misc/salamanders.rmd` because it's part of the R Journal publication. I changed `./glmmTMB/inst/doc/troubleshooting.html`, but not the files underlying it.


mebrooks:
Fixes #643

bbolker:
I think it would be OK to change `misc/salamanders.rmd`, but I don't think it's a big deal (it's not part of the package anyway ...)

Issue #644, opened 2020-12-02
Memory issues in predict/residuals with large model


skanskan:
Hello.

I'm fitting a logistic regression model with mixed effects using the package glmmTMB. (Because the dataset is very large and lme4 produces out of memory errors).    
And I need help to interpret and report the output.

    Family: binomial  ( logit )
    Formula:      OUTPUT ~ SEX + YEAR + OTHER +  (1|CITY/ID)
    Data: mydata
    
          AIC       BIC    logLik  deviance  df.resid 
     890000  891000 -450000  889000    1000000 
    
    Random effects:
    
    Conditional model:
     Groups Name        Variance Std.Dev.
     ID:CITY (Intercept) 10.0    3.1 
     CITY    (Intercept) 1.5    1.2
    Number of obs: 1000009, groups:  ID:CITY, 200000; CITY, 20
    
    Conditional model:
                 Estimate Std.Error z value Pr(>|z|)    
    (Intercept)  0.79   0.28    2.9   0.005
    SEX1       -0.21   0.017 -12   <2e-16
    YEAR        0.48   0.0048 100   <2e-16 
    OTHER       -0.70   0.005 -130   <2e-16 

Output from sjPlot:

    Random Effects
    σ2	3.29
    τ00 ID:CITY	10.0
    τ00 CITY	1.5
    ICC	0.78

How can I report the effect of CITY?
I think most people would calculate the Confidence Interval of the odds of the Intercept with the variance as      
     
     { exp(Intercept-1.96*1.2) , exp(Intercept+1.96*1.2) }   

but this doesn't take into account the standard error of the Intercept from the conditional model, 0.28. How should I combine them?
(I think this is the estimated deviation for the Intercept excluding the random effects).

Is it acceptable to disregard the Intercept value and just say...?    
"The 95% Confidence Interval for the odds ratio for CITY is {exp(-1.96\*1.2),exp(+1.96\*1.2)}"

or 

"The odds of OUTPUT is multiplied by a number between exp(-1.96\*1.2) and exp(+1.96\*1.2) due to the variability of CITY". I don't know how to include here the 95% argument.

I have also tried to calculate the residuals with residuals(model) but it produces an out of memory error.
Is there any way to do it approximately? or is there already some useful information in the output?

I just need to say what proportion of the total variance (or of the residuals) is explained by the variable CITY.

The σ2 value reported by sjPlot is the variance of the residuals.

[...]

PR #681, opened 2021-04-05
Fastpred

bbolker:
This implements the "fast prediction" method (hack?) in #644


bbolker:
In the process of testing this I discovered & fixed a bug in prediction when variables were mapped (map-values were being set to zero rather than filled in from the original `start` values (!)).

I have made `fast=TRUE` the default when `newdata` and `newparams` aren't specified, but we could adopt a more cautious approach if we wanted and leave it optional for now ... (I wonder about our test coverage of prediction ...)

This prediction helps an enormous amount with the problems documented in #644. It does make me wonder a bit why we *ever* need to append the prediction data to the existing data; if we want to predict with `newdata` where `newdata` has 3 rows and `data` has 1e7 rows, why do we need to construct a data set with (1e7)+3 rows ... ?

kaskr:
Looks correct although the logic in `predict.glmmTMB` became harder to follow (as if it wasn't already...).
I have the feeling this could be simplified but it's OK for now - here some comments for future reference:

- We need to append the prediction data to the existing data **only** in the case when new random effects are predicted (because then the full random effect hessian must be updated).
- In all other cases (the 'common case') a fast predict should be possible, i.e. one that avoids calling MakeADFun by simply updating `obj$env$data` and restoring on exit. I think (but may be wrong) that in this 'common case' the whole observation loop can be skipped when calculating ADREPORT variables (set `yobs=numeric(0)`) and the `rbind(olddata,newdata)` can be replaced by `newdata`.
- The `se.fit=TRUE` case can still be surprisingly slow. I have located the cause in `TMB::sdreport` that might be possible to improve...

PR #687, opened 2021-04-23
change order of on.exit to accommodate bug in R <4.0.2

bbolker:
We could also fix this problem by mandating R >= 4.0.2, but the adjustment to the code needed to avoid this bug is fairly trivial.

https://stackoverflow.com/questions/67229026/failing-to-produce-glmmtmb-diagnostics-plots-with-packages-performance-and-dharm

### DIFF
--- a/glmmTMB/R/predict.R
+++ b/glmmTMB/R/predict.R
@@ -175,15 +175,19 @@ predict.glmmTMB <- function(object,
     assign("data",dd, ee) ## stick this in the appropriate environment
     newObj <- object$obj
 
-    ## restore original value
-    on.exit(add = TRUE,
-    {
-        for (i in names(orig_vals)) {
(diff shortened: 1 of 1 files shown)


mebrooks:
This error was also triggered by the predict function. It should be fixed now. Thanks!
```
> owls_nb1 <- glmmTMB(SiblingNegotiation ~ FoodTreatment*SexParent +
+                          (1|Nest)+offset(log(BroodSize)),
+                          contrasts=list(FoodTreatment="contr.sum",
+                          SexParent="contr.sum"),
+                          family = nbinom1,
+                          zi = ~1,
+                          data=Owls)
> p1 <-  predict(owls_nb1)
Error in on.exit(add = TRUE, { : invalid 'add' argument
```

Issue #696, opened 2021-05-22
problems caused by link/inverse link during prediction


bbolker:
Right now we are always returning predictions on the response scale (`mu`), then transforming to the link scale if requested.

This causes problems if, for example, we have a fairly extreme prediction on the `cloglog` scale (e.g. 3.99). Especially given the relatively naive inverse-link function we are using (`1-exp(-exp(eta))`, in contrast to `make.link("cloglog")$linkinv` [which uses `pmax(pmin(-expm1(-exp(eta)), 1 - .Machine$double.eps), .Machine$double.eps)`, this overflows to 1.000 on the probability scale.

We should probably do one or the both of the following:

- modify the cloglog link to be 'clamped' within (0,1) as in `make.link("cloglog")$linkinv` (this would require using something from `CppAD` and/or putting it inside an atomic function so that we can do the hard truncation?)
- modify `glmmTMB.cpp` to return `eta_predict` as well as `mu_predict` and use it instead? (Is it that easy?)

An example from John Maindonald: using [ffly.csv](https://github.com/glmmTMB/glmmTMB/files/6525594/ffly.csv):

```r
## ffly <- read.csv('https://github.com/jhmaindonald/dataR/files/6521483/ffly.csv')
## write.csv(ffly,"ffly.csv", row.names=FALSE)
ffly <- read.csv("ffly.csv")
ffly$obs <- factor(ffly$obs)

form1 <- cbind(Dead,Live)~0+trtGp/TrtTime+(1|obs)+(1|trtGpRep)

library(glmmTMB)
ObsTMB.cll <- glmmTMB(form1,
                      family=binomial(link="cloglog"), data=ffly)
(p0 <- predict(ObsTMB.cll, re.form=NA))[61:65]
p1 <- predict(ObsTMB.cll, re.form=NA,type="response")
1-p1[61:65]
X <- model.matrix(~0+trtGp/TrtTime, data=ffly)
## compute by hand
pp <- c(X %*% fixef(ObsTMB.cll)$cond)
pp[61:65]
1.0-make.link("cloglog")$linkinv(pp[61:65])
```

jhmaindonald:
> We should probably do one or the both of the following:
> 
>   1)  modify the cloglog link to be 'clamped' within (0,1) as in make.link("cloglog")$linkinv (this would require using something from CppAD and/or putting it inside an atomic function so that we can do the hard truncation?)
[...]

PR #697, opened 2021-05-24
return eta_predict

bbolker:
This is in response to #696 ; it returns `eta_predict` as well as `mu_predict`.  Some loose ends that might want to be discussed/dealt with before merging:

- right now the TMB file returns *both* `eta_predict` and `mu_predict`. Could that be a computational/memory burden with a very large data set? Is it worth adding complexity/doing extra work to add a flag that tells TMB whether the predicted value is returned on the link (`eta`) or response (`mu`) scale? (Except that we will probably never revisit this, it seems reasonably harmless to do it this way for now ...)
- I've implemented "clamping" on the cloglog inverse link, as in R's `make.link("cloglog")$linkinv`. Is there any reason there is *not* a good idea? (e.g. computational cost). We don't need it as much as R does because we use the 'robust' versions of the reponse distributions (i.e. we never back-transform all the way to the probability scale), but it could make some computations more reliable/more comparable with results from R (e.g. if someone requests predictions on the response (probability) scale and then uses them in `dbinom()` calculations)
- I'm not sure I've tested the 'zlink' results thoroughly
-


kaskr:
Returning `eta_predict` is an improvement (I never understood why one would ever want to back transform `mu_predict` from within `predict()` when we already have it available for free without lack of accuracy).

However, I'm very much against the clamping part (sorry!).
When `1-p` evaluates to `1` it signifies lack of accuracy which is useful information that should be taken seriously. Clamping hides the problem rather than solving it.

bbolker:
That's why I asked.  I've reverted the clamping.  (If someone really wants clamped results they can use R's inverse-cloglog link: `make.link("cloglog")$linkinv(predict(., type="link"))`

bbolker:
Right now there is a integer `doPredict` flag that is equal to `as.numeric(se.fit)` (i.e. 1 to enable ADREPORT, 0 otherwise).  What do you think about making it 

- 0 if no prediction
- 2 if prediction on the link scale
- 1 if prediction on the response scale ?

This means that we don't have to change the overall structure of the API, and 1 maintains its earlier meaning [although I'm not 100% sure of backward compatibility in all scenarios .. ?]

bbolker:
I've now implemented my suggestion above.

PR #702, opened 2021-06-01
Make hessian symmetric

kaskr:
I see `make test` failing because of (new?) complex eigenvalue check. Note that `Jacobian(gr, par)` will result in un-symmetric result 'at random' so `e_complex_check()` is sort of pointless IMO.

✖ |  30 1     | fitting exotic families [12.4 s]                                                                                                                                                            
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Error (test-families.R:199:5): dbetabinom
Error: detected complex eigenvalues of covariance matrix (max(abs(Im))=1.13473e-10: try se=FALSE?
Backtrace:
 1. base::suppressWarnings(...) test-families.R:199:4
 3. glmmTMB::glmmTMB(...)
 4. glmmTMB::fitTMB(TMBStruc) /zhome/8a/0/39821/glmmTMB/glmmTMB/R/glmmTMB.R:1085:4
 5. glmmTMB:::e_complex_check(eigs$values) /zhome/8a/0/39821/glmmTMB/glmmTMB/R/glmmTMB.R:1384:10

### DIFF
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -1380,6 +1380,7 @@ fitTMB <- function(TMBStruc) {
               par <- par[-rr]
           }
           h <- numDeriv::jacobian(obj$gr, par)
+          h <- .5 * (h + t(h))
           eigs <- eigen(h)
           ev <- e_complex_check(eigs$values)
           if (min(ev)>.Machine$double.eps) {
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -1241,7 +1241,7 @@ fitTMB <- function(TMBStruc) {
             }
         },
         TMB::openmp(NULL)
-    )    
+    )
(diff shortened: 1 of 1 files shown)


bbolker:
I think this was added here c77fc9eb773f5173f6 (by me, about 13 months ago) because complex values mess up the check for a very small minimum eigenvalue. If it would be better/simpler we could check for `min(abs(eigval))` (absolute value acts as a modulus-calculator for complex values) instead and get rid of the complex-value checks?

kaskr:
Now that the hessian is symmetric it should be safe to assume the eigenvalues cannot be complex.
I would suggest removing the `e_complex_check()` function and setting `ev <- eigs$values`.

kaskr:
I'm changing my mind: It doesn't hurt to keep `e_complex_check()` as a defensive check even if the `is.complex(ev)` branch is never used.
I think this is ready to merge.

Issue #651, opened 2020-12-09
Loading old fit with new version


kaskr:
Some of the glmmTMB tests assume that fitted objects saved with old
versions of the package remain valid 'forever' even when TMB and/or
glmmTMB are updated - e.g. `fm2` [here](https://github.com/glmmTMB/glmmTMB/blob/2b14a42bd55cd0cfeebad1f4eb7a3b2313e5d359/glmmTMB/tests/testthat/test-methods.R#L13).
This is currently too much to ask for. Objects saved with a given
version are only guarantied to work when loaded with the same version.

I noticed this problem while testing a future version of TMB:

```r
devtools::install_github("kaskr/adcomp/TMB", ref="TMB_experimental")
devtools::install_github("glmmTMB/glmmTMB/glmmTMB")
```

followed by `make test` from glmmTMB main folder.
I got a bunch of these:

> test-methods.R:73: error: Predict
> gradient in optim evaluated to length 1 not 6
> Backtrace:
>  1. stats::predict(fm2, se.fit = TRUE) tests/testthat/test-methods.R:73:4
>  2. glmmTMB::predict.glmmTMB(fm2, se.fit = TRUE)
>  4. base::with.default(object, optimHess(oldPar, obj$fn, obj$gr))
>  5. [ base::eval(...) ] with 1 more call
>  7. stats::optimHess(oldPar, obj$fn, obj$gr)

bbolker:
Perhaps I'm missing the point, but I had assumed that we would rebuild the examples every time we needed to. There's a Makefile and a make rule for `models.rda` in `inst/test_data`; does that address the problem?

I'm not sure what kind of machinery/rules we have for testing/ensuring that the versions of `glmmTMB` and `TMB` are properly synchronized, or whether we can add a mechanism for knowing when `models.rda` needs to be re-built ...

kaskr:
Thanks @bbolker 
I tried to

```
rm models.rds
make
```

before re-installing glmmTMB.

Now `make test` almost works - there's only one error left:

> test-sparseX.R:23: error: back-compatibility
> Incorrect number of arguments (3), expecting 4 for 'MakeDoubleFunObject'
> Backtrace:
>  1. testthat::expect_is(VarCorr(x), "VarCorr.glmmTMB") tests/testthat/test-sparseX.R:23:4
>  5. glmmTMB::VarCorr.glmmTMB(x)
[...]

PR #703, opened 2021-06-01
Remove old fits (must be remade for each check) #651

kaskr:
TODO: How do we trigger rebuild of these fits - especially on CRAN? Must do something like:

```
cd glmmTMB/inst/test_data
make
```

Perhaps it works already?


bbolker:
Would you like me to poke around/help with this?

kaskr:
@bbolker yes, that would be great thanks! Just tried something new but I'm not sure it works. I'll stop for now.

bbolker:
This seems to work now.  Do you still want me to look at it?

kaskr:
It's passing the checks now which is good. But when I run `make test` from command line there's something wrong:

> Error (test-Anova.R:5:1): (code run outside of `test_that()`)

 Do you get the same?

bbolker:
I think we need to discuss the goal some more. When do you want to rebuild these files/what do you want it to trigger on?  We can create a top-level make rule that runs `cd glmmTMB/inst/test_data; make`, but when should it happen?  The problem with your check is that the files have to be present in the inst/ directory *before* the test run starts; they can't really be built on the fly.

Presumably you want to do this based on matching an old/new TMB version????

kaskr:
The goal is to make sure that the glmmTMB tests cannot block future CRAN submissions of TMB. The tests are *not* allowed to

- Rely on the TMB C-level interface
- Cache computations in a way that makes a new TMB (**or** glmmTMB) version appear much slower (on CRAN) because it is forced to update the fitted objects.

Therefore the fits must be computed for each new installation *regardless of the TMB version*. Ideally the fits should be stored in a temporary location (i.e. not `inst`) and removed when done testing (we don't really want them to take up space in the binary packages on CRAN although this waste is currently just around 1MB).

bbolker:
My last commit makes the tests run/checks pass when `inst/test_data/models.rda` is not present.  However, I'm still a little unclear on the requirements/logic, and why it's not sufficient to update `models.rda` whenever the TMB version changes (e.g., keep up with the master branch of `adcomp/TMB`, which is presumably what future CRAN releases will be based on).

- if `models.rda` is in sync with the current CRAN version of TMB at the time of submission, that should work ... right?
- @kaskr , as the maintainer of TMB, do you have problems with CRAN maintainers when updating TMB because downstream tests on glmmTMB break? (TMB has been updated more recently on CRAN (2021-04-08) than glmmTMB has (2020-07-02), although perhaps there weren't breaking C API changes in that version?
- if the previous bullet point is a problem, we could merge this version (i.e., run `inst/test_data/make_ex.R` every time we test); this takes only 10 seconds on my machine &mdash; but that means it will probably take 40 seconds on some of the Windows builds (which run almost twice as slowly *and* have to check two different architectures). I don't remember what the time limit on tests is (it's not explicitly stated in the repository policy so I would have to go digging in the R check code to see what the threshold is for a warning), but I do know that we spent a considerable amount of effort to skip a lot of tests on CRAN in a previous version ...

kaskr:
The issue with compiled code (`exit`, `abort`, `cerr`) only occurs when you have installed TMB from github. It's safe to ignore I think...

bbolker:
@kaskr, do you have comments on [my comment above](https://github.com/glmmTMB/glmmTMB/pull/703#issuecomment-853433137) (i.e., clarifying the necessity of this change in terms of robustness to changes in TMB API - why we can't just rebuild the examples when we push new versions to CRAN)? Trying hard to keep the checking machinery (the minimal version to be run on CRAN checks) as fast as possible so we don't have to struggle with CRAN maintainers ...

(I don't object to this PR, I just want to be clear that it's necessary before merging.)

kaskr:
My concern is that there are many factors that may affect how long time it takes to generate `models.rda` (e.g. changes to `r-devel` `Matrix` `glmmTMB` `TMB`). I don't want the TMB version to be the only thing that triggers rebuilding `models.rda` on CRAN because it ultimately means that a TMB submission can be rejected because of problems introduced by other packages.

If my understanding is correct then yes, please merge this PR.

bbolker:
I'm not sure, but let's do it. We can always "revert" by including a copy of `models.rda` in `inst/test_data` - that should automatically short-circuit the rebuilding of the examples during the tests

PR #708, opened 2021-06-08
changed data citation urls to dois

mebrooks:
`check --as-cran` was giving this message
```
Found the following URLs which should use \doi (with the DOI name only):
  File ‘Owls.Rd’:
    https://doi.org/10.1016/j.anbehav.2007.01.027
  File ‘Salamanders.Rd’:
    http://dx.doi.org/10.1111/1365-2664.12585
    http://dx.doi.org/10.5061/dryad.5m8f6
```
So I guess this is what they want.

### DIFF
--- a/glmmTMB/man/Owls.Rd
+++ b/glmmTMB/man/Owls.Rd
@@ -28,7 +28,7 @@
   Nestling barn owls beg more intensely in the presence of their mother
   than in the presence of their father.
   \emph{Animal Behaviour} \bold{74} 1099--1106.
-  \url{https://doi.org/10.1016/j.anbehav.2007.01.027};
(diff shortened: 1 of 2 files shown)


bbolker:
I'm not sure what's going on here, maybe some kind of upstream change to pandoc? I don't see anyone commenting about it recently (google search etc.). I restarted the checks hoping that it was a transient glitch. Not obvious how to fix this if not. Staring at https://github.com/r-lib/actions/tree/master/setup-pandoc doesn't give me any clues. (For what it's worth `pandoc-citeproc` is now deprecated, but the workflow seems to be using an older version of pandoc anyway ...

- open an [issue at r-lib/actions](https://github.com/r-lib/actions/issues) ?

Issue #647, opened 2020-12-02
1.1.0 checklist


bbolker:
I think it would be good to think about a new release.  What shall we try to get done before release?

bbolker:
My wish list:

[ ] update NEWS (#637)
[ ] implement fast prediction for large files (#644, #584)
[ ] finish up `better_pdcheck` branch, diagnostic tools
[ ] improve/sort out parallel handling (#620)
[ ] rename overdispersion -> dispersion (#643)
[ ] implement some sort of check for the `poly`+`predict`+matrix-column problem (#635)
[ ] #634 (probably low-hanging?)
[ ] simulation issues (#572)

bbolker:
Bump. Do we have any plans for pushing out a new release? Seems like a good idea. There may be enough improvements to warrant pushing something out even if we seriously triage the release-critical issues list? (Does anyone actually have any time? A lot of these issues are *relatively* low-hanging fruit (we know more or less what to do, we just have to implement stuff)

bbolker:
Bump.  Does anyone have time & energy to drive this? I don't know that we *need* a new release but there is a lot of good stuff ... (version on CRAN is 8 months old)

mebrooks:
Yes, I think it would be good to do this after #670 and #605. I have time March 20-24, possibly sooner but not guaranteed.

bbolker:
Bump again.  I think #605 is basically done, #670 is nearly there (especially if we're OK marking it as "experimental" in the docs and deferring some of the cosmetic/documentation components). I would **really** like to figure out what's up with parallel computation (PR #652 , issue #620)

bbolker:
Bump.  Nearly there?  I'd really like to deal with #620 / #652

bbolker:
Bump. We've got lots of good stuff, it would be nice to get it out to a wider audience ... @mebrooks/@kaskr, what do you think the release-critical issues are at this point?

mebrooks:
I agree that it's time. I'm wiling to work on #637 and to run a test for #651, but I'll need help fixing it if there's a problem with #651. Any other developers willing to help a little?

mebrooks:
[...]

PR #715, opened 2021-06-23
Correct 'invalid' URLS #647

seananderson:
Fix 'invalid' URLS from NOTEs in last R CMD check. Noted here: https://github.com/glmmTMB/glmmTMB/issues/647#issuecomment-866910497


bbolker:
thanks!

PR #717, opened 2021-06-24
cast in pow() for solaris

bbolker:
resolve float/int ambiguity for Solaris


kaskr:
I'm quite convinced this will work (still not tested).

mebrooks:
I've been checking this, but `rhub` is being very slow.  The [results](https://builder.r-hub.io/status/glmmTMB_1.1.0.tar.gz-9c0d1c3384d242cfa1c55918b5e1e6c1) give errors about LaTeX, not C++.

bbolker:
we could try adding `chicago.bst` and `lineno.sty` to the `vignettes` directory.  I don't know if that would cause problems with redundancy on other platforms ...

PR #718, opened 2021-06-24
Phylo gen

bbolker:
this removes the assumption that the number of random effect blocks is equal to the number of factor levels of the corresponding grouping variable (e.g. for some forms of phylogenetic mixed models ...)


kaskr:
That was too fast. Don't want https://github.com/glmmTMB/glmmTMB/pull/718/commits/2dd2bef54fc867fd3926aed9b83946c6bc838662

PR #727, opened 2021-07-13
Openmp fixes

bbolker:
this has more stuff in it to reset OpenMP threads, not warn unnecessarily when OpenMP is missing, etc..  A little bit risky to try it on CRAN, maybe, but it does resolve a bunch of issues that are found on r-hub platforms ...

(I merged `vignette_cran_skip` with it, so it supersedes the other pull request, although merging them both shouldn't hurt)


bbolker:
here we go, hoping for the best

PR #737, opened 2021-07-27
$ and missing-data warnings

bbolker:
Warnings on use of `$` in a formula (`response ~ df$a`) or missing `data`. Think this will be OK/not too annoying: missing-data can be suppressed by specifying `data = NULL`.  Are there sufficiently important use cases for `$` in formulas that we want to provide a way to suppress these warnings?

Not added to `NEWS` yet because there's another incoming PR with `NEWS` changes, but should go in "user-visible changes" section ...


bbolker:
I think these changes are a good idea, but may be worth discussing whether we want to be this prescriptive or not.

mebrooks:
This seems like a good idea to me. It's one of those things we can put in the code once and then avoid having to correct people and talk about it many times later, even if we would hope that people learned "best practices" in the first place. So it will save our time and energy in the end.

mebrooks:
This leads to some warnings in the test. We might need to fix those. E.g.
```
Warning (test-methods.R:60:5): Fitted and residuals
use of the ‘data’ argument is recommended (to suppress this warning, you can explicitly specify ‘data = NULL’
```

mmaechler:
Hmm,  I don't quite see why would recommend  using `data=NULL`  as better than not specifying `data`  (and have all variables in the `globalenv()` , right?).
What we should really recommend is to have 'data = <the data>'   where that ideally is a data frame, but IIRC could also be an environment or a list.  ... what am I missing?

bbolker:
The current warning is phrased as

> "use of the ", sQuote("data"), " argument is recommended (to suppress this warning, you can explicitly specify ", sQuote("data = NULL")

I didn't mean to suggest that users *should* use `data = NULL`; rather, that if they insisted/had a legitimate use case for taking the data from the environment rather than a data frame, they could use `data = NULL` to avoid the warning.

We could perhaps achieve the same goal by including the hint about using `data=NULL` in the man page rather than directly in the warning message?

bbolker:
Adjusted warning (comment on suppressing the warning via data=NULL is now in glmmTMB man page rather than in the warning). Adjusted tests.

bbolker:
I think this is OK now; can we go ahead and merge it? @mmaechler , does the new wording of the warning/description in the documentation work for you?

mebrooks:
Testing the latest version, there's only one warning which I think is justified. SO maybe we just edit the test to expect a warning?
```
Warning (test-families.R:65:5): binomial
use of the ‘data’ argument is recommended‘data = NULL’
Backtrace:
 1. testthat::expect_error(glmmTMB(x ~ 1, family = binomial()), data = data.frame(x)) test-families.R:65:4
```

bbolker:
How about now?

Issue #748, opened 2021-08-24
confusing docs/default setting for na.action


bbolker:
Although it is technically correct, I find `glmmTMB`'s default `na.action` setup confusing. Naively reading the help page etc. it looks like the default setting of `na.action` is `na.fail`, but that's not how the code actually works.  Compare with `lme4::lmer`

`glmmTMB` documentation:

> how to handle missing values, see ‘na.action’ and ‘model.frame’. From ‘lm’: “The default is set by the ‘na.action’ setting of ‘options’, and is ‘na.fail’ if that isunset. The ‘factory-fresh’ default is ‘na.omit’.”

`lmer` documentation: 

>  a function that indicates what should happen when the data contain ‘NA’s.  The default action (‘na.omit’, inherited from the 'factory fresh' value of ‘getOption("na.action")’) strips any observations with any missing values in any variables.

The main difference is that `glmmTMB` **includes a default value of `na.fail` in the function signature**, which is disregarded but makes it look like that's the default.

I also think the `lmer` wording is clearer. It wouldn't hurt to give a hint about `na.exclude`, either ...

PR #749, opened 2021-08-24
improve na.action documentation, remove bogus na.fail default

bbolker:
addresses #748


bbolker:
Is someone willing to merge this (@kaskr @mebrooks @mmaechler )? I know, I could just do it myself ...

PR #755, opened 2021-09-09
Update stored

bbolker:
I think this is ready to go.

I think we should still think about how this integrates with https://github.com/glmmTMB/glmmTMB/pull/667 .  Should individual glmmTMB fits store the version of TMB they were fitted with, internally?  (This would go along with having glmmTMB cache Matrix/TMB version numbers internally when installed ...)


kaskr:
`make test` and `make check-cran` now work for me!

PR #763, opened 2021-09-24
unload DLL on pkg unload

bbolker:
see https://github.com/kaskr/adcomp/issues/353


jaganmn:
I think there is a typo in this commit (replace `"TMB"` with `"glmmTMB"`)...?

PR #769, opened 2021-10-22
Get cor fix

### LINKED ISSUE #768: `get_cor` fills lower triangular matrix in column-major, not row-major order
In `utils.R`, you have:

```r
get_cor <- function(theta) {
    n <- round((1  + sqrt(1+8*length(theta)))/2) ## dim of cor matrix
    L <- diag(n)
    L[lower.tri(L)] <- theta
    cL <- tcrossprod(L)
    Dh <- diag(1/sqrt(diag(cL)))
    cc <- Dh %*% cL %*% Dh
    return(cc[lower.tri(cc)])
}
```

The line `L[lower.tri(L)] <- theta` assumes that the elements of `theta` are in column-major order, but TMB assumes that they are in row-major order:

https://kaskr.github.io/adcomp/classUNSTRUCTURED__CORR__t.html

Here is a version of `get_cor` that is consistent with TMB. It is also slightly more efficient than the original implementation (it modifies in place and avoids matrix multiplies).

```r
actually_get_cor <- function(theta) {
  n <- as.integer(round(0.5 * (1 + sqrt(1 + 8 * length(theta)))))
  R <- diag(n)
  R[upper.tri(R)] <- theta
  R[] <- crossprod(R) # R <- t(R) %*% R
  scale <- 1 / sqrt(diag(R))
  R[] <- scale * R * rep(scale, each = n) # R <- cov2cor(R)
  R[lower.tri(R)]
}
```

Here is a test:

```r
set.seed(145820L)
n <- 5L
S <- cov2cor(crossprod(matrix(rnorm(n * n), n, n)))
R <- chol(S)
R[] <- R * rep(1 / diag(R), each = n)
theta <- R[upper.tri(R)]

x <- S[lower.tri(S)]
y1 <- get_cor(theta)
y2 <- actually_get_cor(theta)
isTRUE(all.equal(x, y1)) # FALSE
isTRUE(all.equal(x, y2)) # TRUE
```

Note that the test under `@examples` in the `get_cor` header only passes because it only considers the `n <= 2` case, for which the row-major and column-major orderings of `theta` are identical.

Running `grep -rnw R -e "get_cor"` in the root of the package suggests that `get_cor` isn't being called anywhere, so this fix would mainly serve to avoid trouble if it is eventually called somewhere.

bbolker:
Fix #768 (bug in get_cor, from Mikael Jagan)


bbolker:
@mebrooks  ?

PR #771, opened 2021-11-22
add an option to print trivial dispersion coef tables

d-morrison:
It might occasionally be useful to print the dispersion parameter as a fixed effects table even when it is only a single intercept. I added an option to print.fixef.glmmTMB to do so.


d-morrison:
Apologies for the slow reply; I made those changes!

PR #790, opened 2022-03-08
Add TMB_SAFEBOUNDS to compilation

kaskr:
Can sometimes detect bugs in the C++ code...


bbolker:
(1) this already found a bounds error! While testing `REML`

kaskr:
I'm just realizing we can't enable safe bounds by default due to CRAN policy... Too bad

kaskr:
I have never noticed a measurable performance hit due to safe bounds.

kaskr:
FWIW `make quick-install` has safe bounds enabled, so no need for this PR

PR #793, opened 2022-03-11
Pdhess logic

bbolker:
this fixes a minor bug (see `tests/testthat/test-misc.R` for example). I think I fixed the logic correctly but wouldn't mind someone double-checking ...


mebrooks (commented):
With the master branch installed, I took `cygu.rda` from elsewhere and ran the test, but I don't see what the problem was.

mebrooks (commented):
I think the new logic here makes sense. In the old version, `g` might not exist when the line `if (any(is.na(g))) {` is executed.

PR #795, opened 2022-03-12
Use short DOI instead of expanded URL

arni-magnusson:
Hi All,

Among the `R CMD check --as-cran` notes that I was seeing (I'm running R 4.1.3 on Debian 11) were:

```
Found the following (possibly) invalid URLs:
https://doi.org/https://doi.org/10.1111/2041-210X.12236
  URL: https://doi.org/https://doi.org/10.1111/2041-210X.12236
    From: inst/doc/covstruct.html
    Status: 503
    Message: Service Unavailable
  URL: https://doi.org/https://doi.org/10.1111/2041-210X.13303
    From: inst/doc/covstruct.html
    Status: 503
    Message: Service Unavailable
```

This seems like a valid note, since we don't want `https://doi.org` to appear twice in the same URL. To fix this problem, I edited `glmmTMB.bib` to change entries
```
  doi = {https://doi.org/10.1111/2041-210X.13303},
  doi = {https://doi.org/10.1111/2041-210X.12236},
```
to
```
  doi = {10.1111/2041-210X.13303},
  doi = {10.1111/2041-210X.12236},
```

Based on recent CRAN reviews of my submitted packages, they recommend using short `\doi{}` entries in documentation rather than expanded `\url{}` entries. In other words, include just the DOI, not the URL that it expands to. This recommendation is also stated in the `?bibentry` help page: "If the URL is an expanded DOI, we recommend to use the doi field with the unexpanded DOI instead."

To have the glmmTMB package follow CRAN recommendations and to have a consistent style within the package, I have:

1. Replaced `doi` with `\doi{}` in Rd files
2. Replaced `https://doi.org` with `\doi{}` in Rd files
3. Replaced `url` and `eprint` with `doi` in CITATION file
4. Replaced `doi: https://doi.org` with `doi:` in CITATION file
5. Replaced `url` with `doi` in glmmTMB.bib file
6. Replaced `doi = {https://doi.org/` with `doi = {` in glmmTMB.bib file

This results in cleaner references. For example, at the bottom of https://cran.r-project.org/web/packages/glmmTMB/vignettes/covstruct.html, we will now have a single hyperlink for each bibliographic reference, instead of two links that lead to the same webpage.

In summary, mainly a cosmetic change for consistency and to follow CRAN recommendations. Also fixes strange URLs such as `https://doi.org/https://doi.org/10.1111/2041-210X.12236` with `https://doi.org` appearing twice inside the same URL.

### DIFF
--- a/glmmTMB/R/family.R
+++ b/glmmTMB/R/family.R
@@ -82,11 +82,11 @@ make_family <- function(x,link) {
 ##' \item Ferrari SLP, Cribari-Neto F (2004). "Beta Regression for Modelling Rates and Proportions." \emph{J. Appl. Stat.}  31(7), 799-815.
 ##' \item Hardin JW & Hilbe JM (2007). "Generalized linear models and extensions." Stata Press.
 ##' \item Huang A (2017). "Mean-parametrized Conway–Maxwell–Poisson regression models for dispersed counts." \emph{Statistical Modelling} 17(6), 1-22.
-##' \item Joe H, Zhu R (2005). "Generalized Poisson Distribution: The Property of Mixture of Poisson and Comparison with Negative Binomial Distribution." \emph{Biometrical Journal} 47(2): 219–29. doi:10.1002/bimj.200410102.
+##' \item Joe H, Zhu R (2005). "Generalized Poisson Distribution: The Property of Mixture of Poisson and Comparison with Negative Binomial Distribution." \emph{Biometrical Journal} 47(2): 219–29. \doi{10.1002/bimj.200410102}.
 ##' \item Morris  W (1997). "Disentangling Effects of Induced Plant Defenses and Food Quantity on Herbivores by Fitting Nonlinear Models." \emph{American Naturalist} 150:299-327.
 ##' \item Sellers K & Lotze T (2015). "COMPoissonReg: Conway-Maxwell Poisson (COM-Poisson) Regression". R package version 0.3.5. https://CRAN.R-project.org/package=COMPoissonReg
-##' \item Sellers K & Shmueli G (2010) "A Flexible Regression Model for Count Data." \emph{Annals of Applied Statistics} 4(2), 943–61. https://doi.org/10.1214/09-AOAS306.
-##' \item Shonkwiler, J. S. (2016). "Variance of the truncated negative binomial distribution." \emph{Journal of Econometrics} 195(2), 209–210. doi:10.1016/j.jeconom.2016.09.002
+##' \item Sellers K & Shmueli G (2010) "A Flexible Regression Model for Count Data." \emph{Annals of Applied Statistics} 4(2), 943–61. \doi{10.1214/09-AOAS306}.
+##' \item Shonkwiler, J. S. (2016). "Variance of the truncated negative binomial distribution." \emph{Journal of Econometrics} 195(2), 209–210. \doi{10.1016/j.jeconom.2016.09.002}.
 ##' }
 ##' @export
 ##' @importFrom stats make.link
--- a/glmmTMB/inst/CITATION
+++ b/glmmTMB/inst/CITATION
@@ -20,24 +20,23 @@ bibentry(bibtype = "Article",
          title = "{glmmTMB} Balances Speed and Flexibility Among Packages for Zero-inflated Generalized Linear Mixed Modeling",
          year = "2017",
          journal = "The R Journal",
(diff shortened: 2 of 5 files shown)


bbolker:
Hmm, this hit an error.

```
Error: Error: processing vignette 'glmmTMB.Rnw' failed with diagnostics:
Running 'texi2dvi' on 'glmmTMB.tex' failed.
LaTeX errors:
! Undefined control sequence.
l.172 doi:\nobreakspace{}\Rhref
                               {[https://doi.org/10.32614/RJ-2017-066}{10.326...](https://doi.org/10.32614/RJ-2017-066%7D%7B10.326...)
The control sequence at the end of the top line
of your error message was never \def'ed. If you have
```

arni-magnusson:
Hi Ben, thanks for looking into this.

This runs on my machine (R 4.1.3 on Debian 11) without problems:

```
$ R CMD Sweave --pdf glmmTMB.Rnw
Output file:  glmmTMB.pdf
$
```

bbolker:
@arni-magnusson , does the full `R CMD build ...; R CMD check --as-cran ...` pipeline work as well? Maybe the environments/commands used inside `R CMD check` aren't exactly equivalent to `R CMD Sweave --pdf glmmTMB.Rnw` ?  I'm nervous about merging something that breaks on one of our test platforms (although I *don't* particularly looking forward to diagnosing/debugging this).

bbolker:
Sebastian Meyer says [on r-package-devel](https://stat.ethz.ch/pipermail/r-package-devel/2022q1/007829.html):

```
I think this comes from the glmmTMB vignette doing

> print(citation("glmmTMB"),style="latex")

which produces LaTeX code via Rd2latex(). This uses Rd's \Rhref, wrapping 
either \href from hyperref.sty, or, as a fallback, \url from url.sty. In addition, 
the default tools::bibstyle() will also output the volume number in \bold, another LaTeX command from Rd.sty.

I can see the vignette already accounts for \bold via

> %% need this for output of citation() below ...
> \newcommand{\bold}[1]{\textbf{#1}}

To fix the missing \Rhref definition you could similarly do

\NewCommandCopy\Rhref\href

after including hyperref (\NewCommandCopy needs LaTeX >= 2020-10-01).
```

I haven't played around with/tested this yet. It would be nice to avoid backward compatibility issues with old versions of LaTeX, but I guess people with old versions of LaTeX can just disable vignette-building?

bbolker:
@arni-magnusson , it would be nice to incorporate this instead of having it fall through the cracks.  Could you bump something / try incorporating Sebastian Meyer's suggestion above / merge the current master branch back in so that we can see if it passes tests?

bbolker:
@arni-magnusson is this worth following up?  (Heading toward another release sometime soon-ish ...)

mebrooks (commented):
I'm thinking about merging this or part of this. CRAN replied to my recent submission with the following
```
Sorry, the 503 URLs are wrong: please fix. You have

glmmTMB.bib:	doi = {https://doi.org/10.1111/2041-210X.13303},
glmmTMB.bib:	doi = {https://doi.org/10.1111/2041-210X.12236},

but you should really have e.g. doi = {10.1111/2041-210X.13303}
only.
```
So I asked for clarification about whether or not we need to delete urls.

mmaechler:
>>>>> Mollie Brooks 
>>>>>     on Mon, 14 Nov 2022 06:57:40 -0800 writes:

    > @mebrooks commented on this pull request.
    > I'm thinking about merging this or part of this. CRAN replied to my recent submission with the following
    > ```
    > Sorry, the 503 URLs are wrong: please fix. You have

    > glmmTMB.bib:	doi = {https://doi.org/10.1111/2041-210X.13303},
    > glmmTMB.bib:	doi = {https://doi.org/10.1111/2041-210X.12236},

    > but you should really have e.g. doi = {10.1111/2041-210X.13303}
    > only.
    > ```
    > So I asked for clarification about whether or not we need to delete urls. 

?  No.  Instead of

    doi = {https://doi.org/10.1111/2041-210X.13303},

you "want"

    doi = {10.1111/2041-210X.13303},

bbolker:
This URL was changed to a DOI, which is arguably a better long-term way to reference the document -- URLs change when publishers and organizations reorganize their web pages, but the DOIs should (by design) be stable.  Are there URLs that were *not* replaced by DOIs?

mebrooks:
It looks like all the deleted URLs have DOIs, so it should be fine. I didn't know the best-practice on publication web addresses had changed.

Issue #787, opened 2022-02-14
1.1.3 release checklist


mebrooks:
We need to update to solve some errors on CRAN https://cran.r-project.org/web/checks/check_results_glmmTMB.html before before Feb 27. I'm on vacation until Feb 18 with a stack of work already, so I can't do much this week.

I'm uncertain which number is best for this release (we can change the issue name). In #759, @bbolker had changed to 1.1.3, but I was confused and changed it back to 1.1.2.3 before submitting to CRAN.

bbolker:
I suspect the most significant errors *may* be due to changes in `testthat`  (version 3.1.2 was released 2022-01-20); some of the others may possibly be driven by TMB binary stack incompatibilities (e.g. if `Matrix` version is more recent than the most recently released TMB version on CRAN ...) -- however at the moment I'm having trouble reproducing them locally.

bbolker:
Wish list:

[ ] add machinery for less-than-full rank X matrices (i.e. import machinery from [here](https://github.com/bbolker/nlme/blob/main/R/lme.R#L2905-L2960))
[ ] fix `vcov(., full = TRUE)` for multi-to-one parameter mapping (i.e., `map` with repeated values, not just `NA`): #683 
[ ] maybe address #758 (includes diagnosis/fixes)
[ ] bugs? #634, #662

bbolker:
@mebrooks , any thoughts? Minimally, I think we could get away with merging https://github.com/glmmTMB/glmmTMB/pull/788, doing the full round of testing, and submitting to CRAN ...

mebrooks:
I'm doing checks now, just to be prepared for the deadline Sunday. I'm traveling for work until Saturday afternoon, so it's unlikely that I'll have time to tackle anything on the wishlist this time. Sorry. I'll try to do more next time. 

The error was with R Under development on platform: x86_64-w64-mingw32 (64-bit) which I guess might correspond to `windows-x86_64-devel` in R-Hub. I'll check it.

mebrooks:
Windows (release and devel) seems ok and only gives notes about package size.

Solaris gives an error that seems to come from `nloptr` which causes a problem with [`lme4`](https://github.com/lme4/lme4/issues/610).
[`ERROR: compilation failed for package ‘nloptr’`](https://builder.r-hub.io/status/glmmTMB_1.1.2.9000.tar.gz-e47f2107d8944d749ffed9882a2dba80)

I'm trying debian now with both patched and devel versions of R.

[...]

PR #796, opened 2022-03-13
prevent up2date from hogging threads

bbolker:


bbolker:
I think this fixes our win-builder/Windows problems! #787

kaskr:
It might be better to follow the same pattern as `predict`:

```r
n_orig <- openmp()
on.exit(openmp(n = n_orig), add = TRUE)
```

but that's not so important - can do another time.

Issue #797, opened 2022-03-19
Pearson residuals for ZI/variable dispersion models


bwiernik:
I'd like to get Pearson residuals working for zero-inflated and variable dispersion models if possible. The *pscl* package provides Pearson residuals for its parameterizations of ZIP and ZINB(2), and these perform well based on some quick simulations.

From the *pscl* package:

For zero-inflated models:

```r
function (object, type = c("pearson", "response"), ...) 
{
  type <- match.arg(type)
  res <- object$residuals
  switch(type, response = {
    return(res)
  }, pearson = {
    mu <- predict(object, type = "count")
    phi <- predict(object, type = "zero")
    theta1 <- switch(object$dist, poisson = 0, geometric = 1, 
      negbin = 1/object$theta)
    vv <- object$fitted.values * (1 + (phi + theta1) * mu)
    return(res/sqrt(vv))
  })
}
```

For hurdle models:

```r
function (object, type = c("pearson", "response"), ...) 
{
  type <- match.arg(type)
  res <- object$residuals
  switch(type, response = {
    return(res)
  }, pearson = {
    mu <- predict(object, type = "count")
    phi <- predict(object, type = "zero")
    theta1 <- switch(object$dist$count, poisson = 0, geometric = 1, 
      negbin = 1/object$theta["count"])
    vv <- object$fitted.values * (1 + ((1 - phi) + theta1) * 
      mu)
    return(res/sqrt(vv))
  })
}
```

As far as I can see, implementing these and related formulas for other families in *glmmTMB* requires these steps:

1. In `residuals()`, passing `predict(mod, type = "disp") * sigma(mod)` to the second argument of the `variance()` function, rather than just `sigma(mod)`.
3. Adapting the `variance()` arguments to expect a `zprob` argument. (We might also want to add a `shape` argument for Tweedie distributions while we are at it (cf. https://github.com/glmmTMB/glmmTMB/issues/293).3. 
4. In `residuals()`, passing `predict(mod, type = "zprob")` to the new `zprob` argument of the `variance()` function.
5. In `residuals()`, using `predict(mod, type = "conditional")` to extract the conditional `mu` rather than `fitted(mod)` to extract the marginal response.
6. Not necessary, but would make managing the additional arguments easier, maybe it would be a good idea to make the second argument of the `variance()` function consistent across families, rather than varying between phi/theta/alpha.

If these sound correct, I'm happy to get started on them. 

Related issues:
[...]

PR #800, opened 2022-03-19
Support Pearson residuals for zi and variable dispersion models

bwiernik:
This is a work in progress. Currently, I've generalized the call to `variance()` in `residuals.glmmTMB()` so that it can handle variance() functions with a variety of different numbers and names of arguments. 

Next will update various families to add zero inflation and other arguments. 

Closes https://github.com/glmmTMB/glmmTMB/issues/797


bwiernik:
@bbolker This is ready for review.

I'm not exactly sure about https://github.com/glmmTMB/glmmTMB/pull/800/commits/62a151d6c7a0253096dec64cb1a3642feb097120. I don't think the current solution is going to be appropriate for unscaling the variance for all families; to do that might require special handing for each specific family.

I haven't written any additional tests. Can you advise what you'd like there?

bbolker:
I'm sorry if I'm being thick, but can you elaborate on where [this equivalence](https://github.com/glmmTMB/glmmTMB/pull/800/commits/d25e4506979ec997c4c1df4f2c7ada6901198b93#diff-37f414490e4459ba2c7c48930299a558e2ba6f596cb46de0917d6647bc9cdb2aR683-R687) comes from?

I see [this](https://math.stackexchange.com/questions/4197681/what-is-the-variance-of-the-product-of-a-bernoulli-0-1-and-a-normal-random-var); can we assume independence of the zero-inflation and the conditional component? I think so but I'm having trouble completely convincing myself ... (note that this is the variance *conditional on the estimates of the distributional parameters*, not including the uncertainty of the parameters themselves ...)

bwiernik:
It comes from the general formula for the variance of the product of two independent variables (e.g., [1](https://www.ucl.ac.uk/%7Eucfbpve/geotopes/indexch10.html) or [second formula at 2](https://math.stackexchange.com/questions/4271183/variance-of-product-of-two-random-variables-fx-y-xy)); it's a take on the "shortcut" formula Var[x] = E[X^2] - E[X]^2.

As you note, this is the model-implied variance of the residuals *conditional on the estimates of the distributional parameters*. The zero-inflation model assumes that the zero-inflation and conditional components of the model are independent by definition. So the estimated conditional variance itself doesn't need to handle any covariance. (SEs on the estimated conditional variances or Pearson residuals would need to handle the covariance in the uncertainty of the parameter estimates.)

For demonstration, see that the formula yields the same result as the variance function used by the *pscl* package implementing the "classic" variance of a zero-inflated Poisson. Then see, that it yields the same result as observed for a simulated zero-inflated Gaussian model:

``` r
mzip <- pscl::zeroinfl(
  art ~ fem + mar + kid5 + phd + ment | fem + mar + kid5 + phd + ment, 
  dist = "poisson",
  data = pscl::bioChemists
)

mu <- predict(mzip, type = "count")
phi <- predict(mzip, type = "zero")
theta1 <- 0

vv_formula <-  mu * (1 - phi) * (1 + (phi + theta1) * mu)
vv_bmw     <-  mu * (1 - phi) + mu^2 * (1 - phi) * phi

identical(round(vv_formula, 6), round(vv_bmw, 6))
#> [1] TRUE


zi <- rbinom(10000, 1, .8)
x  <- rnorm( 10000, mean = 2, sd = 1)
y  <- data.frame(y = x * zi)


mzig <- glmmTMB::glmmTMB(
  y ~ 1, ziformula = ~ 1, data = y, family = gaussian()
)

mu2   <- predict(mzig, type = "conditional")[1]
phi2  <- predict(mzig, type = "zprob")[1]
disp2 <- predict(mzig, type = "disp")[1]

( vv_sim2 <- mean(y$y^2) - mean(y$y)^2 )
#> [1] 1.444794
( vv_bmw2 <- disp2^2 * (1 - phi2) + mu2^2 * (1 - phi2) * phi2 )
#> [1] 1.444794
```

<sup>Created on 2022-03-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bbolker:
Can you please pull upstream to resolve the glitch in the NEWS file (which I think is not your fault), and remove the `:::` in `glmmTMB:::.tweedie_power` to resolve the NOTE?

bwiernik:
Okay, those are both fixed.

bbolker:
I think this failed because of transient TMB/Matrix incompatibilities. I'm going to merge it and hope for the best/fix it on the other side.

Issue #798, opened 2022-03-19
Bug: predict = "prob" returning "response" values for non-ZI models


bwiernik:
If I run, `predict(object, type = "zprob")` and `predict(object, type = "zlink")` with a non-zero-inflated model, then the values returned are those for `predict(object, type = "response")`, rather than an error as described in the documentation.

> "zprob"
> the probability of a structural zero (gives an error for non-zero-inflated models)
> 
> "zlink"
> predicted zero-inflation probability on the scale of the logit link function

``` r
predict(glmmTMB::glmmTMB(mpg ~ log(disp), data = mtcars), type = "zprob")
#>  [1] 22.03911 22.03911 25.69183 17.59883 14.50276 18.87072 14.50276 22.84563
#>  [9] 23.22712 21.60783 21.60783 16.97880 16.97880 16.97880 11.98540 12.22473
#> [17] 12.63784 28.63310 28.99429 29.57691 24.70493 15.65564 16.07407 14.76457
#> [25] 13.52360 28.59774 24.68947 26.87398 14.73805 22.95396 16.16623 24.63555
predict(glmmTMB::glmmTMB(mpg ~ log(disp), data = mtcars), type = "zlink")
#>  [1] 22.03911 22.03911 25.69183 17.59883 14.50276 18.87072 14.50276 22.84563
#>  [9] 23.22712 21.60783 21.60783 16.97880 16.97880 16.97880 11.98540 12.22473
#> [17] 12.63784 28.63310 28.99429 29.57691 24.70493 15.65564 16.07407 14.76457
#> [25] 13.52360 28.59774 24.68947 26.87398 14.73805 22.95396 16.16623 24.63555
predict(glmmTMB::glmmTMB(mpg ~ log(disp), data = mtcars), type = "response")
#>  [1] 22.03911 22.03911 25.69183 17.59883 14.50276 18.87072 14.50276 22.84563
#>  [9] 23.22712 21.60783 21.60783 16.97880 16.97880 16.97880 11.98540 12.22473
#> [17] 12.63784 28.63310 28.99429 29.57691 24.70493 15.65564 16.07407 14.76457
#> [25] 13.52360 28.59774 24.68947 26.87398 14.73805 22.95396 16.16623 24.63555
```

[...]

PR #801, opened 2022-03-19
Zprob pred fix

bbolker:
addresses #798


bbolker:
(the checkDepPackageVersion was supposed to go straight to the master branch but is probably harmless here)

PR #805, opened 2022-04-07
diagnose() improvements (explanation, handle complex hess eigs)

bbolker:


bbolker:
There is a test case [here](https://stackoverflow.com/questions/71784848/inflated-or-not-inflated-true-zero-dilemma-in-glmm/71789207#71789207) for complex eigs, don't know if I should save data or the fitted model for tests/examples?

PR #819, opened 2022-05-27
Tmb 1.9.0

kaskr:
Passes `devtools::check(vignettes = FALSE)` for me. I didn't get a note about `TMB::openmp` which I would have expected needs to be added to imports?

### DIFF
--- a/glmmTMB/DESCRIPTION
+++ b/glmmTMB/DESCRIPTION
@@ -40,7 +40,7 @@ License: AGPL-3
 Depends: R (>= 3.2.0)
 Imports:
     methods,
-    TMB (>= 1.7.14),
+    TMB (>= 1.9.0),
     lme4 (>= 1.1-18.9000),
     Matrix,
     nlme,
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -15,7 +15,7 @@ openmp <- function (n = NULL) {
       w <- options(warn = -1)
       on.exit(options(warn = w[["warn"]]))
     }
-    .Call("omp_num_threads", n, PACKAGE = "glmmTMB")
+    TMB::openmp(n, DLL="glmmTMB")
 }
 
 ##' Change starting parameters, either by residual method or by user input (start)
--- a/glmmTMB/src/glmmTMB.cpp
+++ b/glmmTMB/src/glmmTMB.cpp
@@ -441,13 +441,6 @@ Type allterms_nll(vector<Type> &u, vector<Type> theta,
 template<class Type>
 Type objective_function<Type>::operator() ()
 {
-#ifdef _OPENMP
-  this -> max_parallel_regions = omp_get_max_threads();
-  // std::cout << "OpenMP max_parallel_regions=" << this -> max_parallel_regions << "\n";
-#else
-  this -> max_parallel_regions = 1;
-  // std::cout << "no OpenMP (max_parallel_regions=1)\n";
-#endif
 
   DATA_MATRIX(X);
   bool sparseX = X.rows()==0 && X.cols()==0;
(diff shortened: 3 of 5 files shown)


bbolker:
Since `glmmTMB` already imports `TMB`, and you refer to `TMB::openmp` explicitly, you don't need to add `@importFrom` in Roxygen, or add it to the NAMESPACE file (which is what Roxygen does when it builds). Fails checks on Github Actions, I'll check and see what's going on ...

bbolker:
in order to make the GHA check work we'll have to add something to the workflow to have it get the development version rather than the CRAN version of TMB ...

kaskr:
Version 1.9.0 is on CRAN now so may just be a matter restarting the GA check? (I don't know how to do that...)

bbolker:
I'm assuming that's some transient weirdness, restarted again ...

bbolker:
There's a lot of other `glmmTMB::openmp()` stuff on the R side. Is that OK or should it be changed ...?

bbolker:
Also, I think it would be better to delete the obsolete stuff rather than commenting it out (that's what revision control is for ...)

kaskr:
Cleaned up my messy commits.

kaskr:
```shell
git grep "openmp("
```

shows me all (?) left-over openmp stuff to consider. I must admit that I don't really understand the logic behind all of it...

Some of it seems redundant but harmless (better keep it, just in case...), e.g.:

```
R/utils.R:  openmp(1)  ## non-parallel/make sure NOT grabbing all the threads!
```

I gather the purpose of this stuff:

```
R/glmmTMB.R:    n_orig <- openmp(NULL)
R/glmmTMB.R:          openmp(n = n_orig)
R/methods.R:          n_orig <- openmp(n = object$modelInfo$parallel)
R/methods.R:          on.exit(openmp(n_orig))
R/predict.R:  n_orig <- openmp(n = object$modelInfo$parallel)
R/predict.R:  on.exit(openmp(n_orig), add = TRUE)
R/predict.R:    n_orig <- openmp(NULL)
R/predict.R:    on.exit(openmp(n = n_orig), add = TRUE)
```

is to restore a global openmp state (?) perhaps to avoid interfering with other packages (?). This shouldn't be necessary anymore...

There's some combination of `mclapply` and openmp going on here, which could potentially fork a multithreaded R process (undefined behavior):

```
R/profile.R:            n_orig <- openmp(NULL)
R/profile.R:            openmp(n = fitted$modelInfo$parallel)
R/profile.R:            on.exit(openmp(n_orig))
```

It would make sense to disable threading before forking, but that's not really what seems to be going on (?).

Overall I feel there's plenty of openmp stuff to tidy but I don't think it's strictly necessary right now. The present PR can't possibly make anything worse than it already is.

bbolker:
For the record, a lot of the 'extraneous' looking stuff was added by me when I discovered that machinery that called the TMB component of the fitted object *without* going through `glmmTMB` (e.g. profiling, prediction, etc.) was likely to skip over setting the OpenMP thread values to values that were specified (either when fitting the model or in an `options()` variable).  I don't have great test cases for these, but the standard way to check them is for me to run tests on my 32-core machine while keeping an idea on the system monitor and see if anything goes crazy ...

In any case I'm inclined to accept this but then open up another "clean up OpenMP" issue so we don't lose track.

mebrooks:
This passes checks on my computer and Winbuilder. It gets the same note as the master branch on R-release, so I think it's fine. The note does not appear on R-devel.
```
* checking HTML version of manual ... NOTE
Found the following problems:
Owls.Rd:17:1: Warning: <table> attribute "width" not allowed for HTML5

Salamanders.Rd:17:1: Warning: <table> attribute "width" not allowed for HTML5

epil2.Rd:17:1: Warning: <table> attribute "width" not allowed for HTML5
```

bbolker:
OK, go ahead and merge (and let's open a "clean up OpenMP cruft" issue?)

Issue #817, opened 2022-05-25
confint for random effects in more complex cases


strengejacke:
I'd like to calculate confidence intervals for the random effect variances. This works in "simple" cases, but `confint()` fails for more complex RE structures (might also be due the warnings during model fitting, I don't know). Do you think these issues can be solved?

``` r
library(glmmTMB)
library(lme4)
#> Loading required package: Matrix

data(sleepstudy)
data(cake)
set.seed(123)
sleepstudy$Months <- sample(1:4, nrow(sleepstudy), TRUE)

m1 <- glmmTMB(Reaction ~ Days + (Days | Subject), data = sleepstudy)
m2 <- glmmTMB(angle ~ temperature + (temperature | recipe) + (temperature | replicate), data = cake)
#> Warning in fitTMB(TMBStruc): Model convergence problem; non-positive-definite
#> Hessian matrix. See vignette('troubleshooting')
#> Warning in fitTMB(TMBStruc): Model convergence problem; iteration limit reached
#> without convergence (10). See vignette('troubleshooting')
m3 <- glmmTMB(angle ~ temperature + (temperature | recipe), data = cake)
#> Warning in fitTMB(TMBStruc): Model convergence problem; non-positive-definite
#> Hessian matrix. See vignette('troubleshooting')
m4 <- glmmTMB(angle ~ temperature + (temperature | replicate), data = cake)
#> Warning in fitTMB(TMBStruc): Model convergence problem; non-positive-definite
#> Hessian matrix. See vignette('troubleshooting')
#> Warning in fitTMB(TMBStruc): Model convergence problem; false convergence (8).
#> See vignette('troubleshooting')
m5 <- glmmTMB(Reaction ~ Days + (Days + Months | Subject), data = sleepstudy)
#> Warning in fitTMB(TMBStruc): Model convergence problem; non-positive-definite
#> Hessian matrix. See vignette('troubleshooting')

confint(m1, parm = "theta_", method = "wald", level = .9)
#>                             5 %       95 %    Estimate
#> Std.Dev.(Intercept)  16.1689591 34.9753951 23.78057467
#> Std.Dev.Days          4.0627802  8.0442555  5.71682093
#> Cor.Days.(Intercept) -0.4147801  0.5263468  0.08132635

confint(m2, parm = "theta_", method = "wald", level = .9)
#> Warning in sqrt(diag(vv)): NaNs produced
#> Warning in sqrt(diag(object$cov.fixed)): NaNs produced
[...]

PR #824, opened 2022-06-03
Ci tweaks

bbolker:
This fixes what seems to be a fairly serious oversight in computing confidence intervals for random effects parameters - just didn't work in cases with multiple random effects terms! https://github.com/glmmTMB/glmmTMB/issues/817  However, I think we should be careful about merging this, may want to wait until after pushing the next release ... ? Or at least make sure we have tested downstream carefully ...

### DIFF
--- a/glmmTMB/R/methods.R
+++ b/glmmTMB/R/methods.R
@@ -955,7 +955,11 @@ confint.glmmTMB <- function (object, parm = NULL, level = 0.95,
         ## Take subset
 
         ## drop mapped values (where lower == upper)
-        ci <- ci[ci[,2]!=ci[,1], , drop=FALSE]
+        ## can get confused by failed stderr calculation
+        ##  (both lwr and upr CIs are NA/NaN)
+        mapped <- !(is.na(ci[,1] & is.na(ci[,2]))) &
+            (ci[,2] == ci[,1])
+        ci <- ci[!mapped, , drop=FALSE]
 
         ## now get selected parameters
         if (!is.null(parm)) {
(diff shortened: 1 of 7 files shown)


bbolker:
Since CRAN isn't yelling at us right now, so we're not in an **big** hurry, I think we should merge this and see what breaks downstream ... (important to make sure this is clearly described in NEWS)

bbolker:
Further evidence that this should be merged: https://stackoverflow.com/questions/72552991/ggcoef-model-error-when-two-random-intercepts/72553582#72553582

bbolker:
Should be working reasonably consistently for mapped parameters now **for Wald CIs**.  Need to extend this to work with `uniroot` and `profile` CIs, which will (I think) mean refactoring/extracting the code for computing parameter names so that it can be used outside of its current narrow context ...

bbolker:
This is working at least for some simple cases for profile/uniroot CIs.  Definitely still have some failing cases ... (current version is still much better than existing version, though, I think ...)

bbolker:
Factorial tests at https://github.com/glmmTMB/glmmTMB/blob/ci_tweaks/misc/test_confint.R

method = {wald, profile, uniroot} × include_mapped = {TRUE, FALSE} × 4 test models × parm = {NULL (all), beta_, theta_}

- wald always OK
- beta_ OK except for uniroot + include_mapped (all models)
- {NULL, theta_} + profile works only for model 4/no-mapped
   - fails with 4/mapped (subscript out of bounds)
   - fails for other reasons (infinite profile problem) for models 1, 2, 3
- {NULL, theta_} + uniroot fails ('length of dimnames [2] not equal to array extent') as long as include_mapped is TRUE (all models)

PR #825, opened 2022-06-07
Add cov.fit argument to predict() method

bwiernik:
This would permit computation of custom contrasts of predicted values (e.g., each treatment vs control).


bwiernik:
(l. 422 was unintentional)

PR #829, opened 2022-06-17
Smart refit

bbolker:
this may or may not be finished, but I think it might be sufficiently useful to merge?


mebrooks:
Is the new function meant to be exported? 

I suggest adding an example of how it should be used. I tried to make one (expecting it would work the same as `refit` and made it break in two different ways (details below). The error from the salamander model was `Error in EvalDoubleFunObject(Fun, theta, do_simulate = do_simulate) : 
  Wrong parameter length.` and the error from the dispersion model was `caught segfault`.
```
library(glmmTMB)
 
(m1 <- glmmTMB(count ~ mined + (1|site),
   family=poisson, data=Salamanders))

Salamanders2 <- transform(Salamanders, count2=2*count)

refit(m1, Salamanders2$count2)

fast_refit(m1, Salamanders2$count2)

glmmTMB:::fast_refit.glmmTMB(m1, Salamanders2$count2)


## Dispersion model
sim1 <- function(nfac=40, nt=100, facsd=0.1, tsd=0.15, mu=0, residsd=1)
{
  dat <- expand.grid(fac=factor(letters[1:nfac]), t=1:nt)
  n <- nrow(dat)
  dat$REfac <- rnorm(nfac, sd=facsd)[dat$fac]
  dat$REt <- rnorm(nt, sd=tsd)[dat$t]
  dat$x <- rnorm(n, mean=mu, sd=residsd) + dat$REfac + dat$REt
  dat
}
set.seed(101)
d1 <- sim1(mu=100, residsd=10)
d2 <- sim1(mu=200, residsd=5)
d1$sd <- "ten"
d2$sd <- "five"
dat <- rbind(d1, d2)
m0 <- glmmTMB(x ~ sd + (1|t), dispformula=~sd, data=dat)

dat2 <- transform(dat, x2=2*x)
refit(m0, dat2$x2)

glmmTMB:::fast_refit.glmmTMB(m0, dat2$x2)
```

bbolker:
If it can be debugged and made robust (and if benchmarking suggests it's worth it), I'd like to roll it into `refit.glmmTMB` as an option.

bbolker:
I've integrated this into `refit` now. https://github.com/glmmTMB/glmmTMB/blob/smart_refit/misc/refit_test.R tests it; for 100 parametric bootstrap samples, on my machine, I get 73 seconds ('fast') vs 125 seconds (not-fast) - seems worthwhile!

bbolker:
@mebrooks, this is getting better.   Re questions from [above](https://github.com/glmmTMB/glmmTMB/pull/829#issuecomment-1160469993):

- the new function `fast_refit()` isn't exported, it's meant to be used via `refit(..., fast = TRUE)` (eventually we could switch the default to `TRUE`)
 - Your first example works (at least doesn't fail); your second example fails because there are `NA` values in the original response variable. I'm not sure how we should handle that - I guess we should filter using the `na.action` attribute of the model frame ... ?

bbolker:
do I need to add `DATA_UPDATE()` to the C++ code to guarantee that this works properly ... ???

PR #837, opened 2022-07-08
Enable new AD library

kaskr:
Trigger GA tests (?)


bbolker:
Are there any risks to merging this? If so, I would vote for merging *after* we send the next release to CRAN, then merge and let this live in the development version to find out if there are any surprises ...

PR #842, opened 2022-07-19
Rank deficiency checking for fixed effects model matrices

stouffer:
This code provides the user with the option to check for rank deficiency of their fixed-effects model structures.

Currently the code allows a user to fit model with linearly-dependent predictors. This will often, but not always, lead to convergence issues and non-positive-definite variance-covariance matrices. Though some users may wish to continue to fit models of this sort, by providing the option it can (i) provide valuable clarity for why they receive a "Model convergence problem" and (ii) allow users to avoid fitting deficient/over-parameterized models.


bbolker:
Should I merge this or would you like to keep working on it here?

stouffer:
I'm happy for it to be merged. We can then add the need for better unit tests and downstream modifications of `fixef()` etc as further items in the existing issue or just as a new issue. Whatever works best for you.

Issue #853, opened 2022-09-05
predict(newdata =,)) issue when using multiple transformations of same variable 


willtudorevans:
I seem to be getting very strange predictions when using predict(newdata =)) with a glmmTMB mode. Predictions that don't align with fitted(mod) values if I rebuild the predictions on trained data. The issue seems to arise if I use a second transformation of the same variable, and particularly if that transformation involves a '+' sign.

Apologies for the esoteric reproducible example but I hope it demonstrates the point ok. I'm using glmmTMB v1.1.4.

```
rm(list = ls())
library(glmmTMB)
set.seed(100)
N <- 100
x1 <- rnorm(N)
Y <- rpois(N, lambda = exp(x1))
df <- data.frame(Y=Y, x1=x1)

#Base model
mod <- glmmTMB(Y ~ x1 + I((x1+10)^2) + I((x1+10)^3), data = df, family = "poisson")
all(round(fitted(mod),2) == round(predict(mod, newdata = df, type = "response"),2))
#not ok

#Rearrange ordering of second transformation
mod <- glmmTMB(Y ~ x1 + I((x1+10)^2) + I((10+x1)^3), data = df, family = "poisson")
all(round(fitted(mod),2) == round(predict(mod, newdata = df, type = "response"),2))
#ok

#Remove second transformation
mod <- glmmTMB(Y ~ x1 + I((x1+10)^2), data = df, family = "poisson")
all(round(fitted(mod),2) == round(predict(mod, newdata = df, type = "response"),2))
#ok

#Try glm of base model
mod <- glm(Y ~ x1 + I((x1+10)^2) + I((x1+10)^3), data = df, family = "poisson")
all(round(fitted(mod),2) == round(predict(mod, newdata = df, type = "response"),2))
#ok

#Add fourth transformation
mod <- glmmTMB(Y ~ x1 + I((x1+10)^2) + I((10+x1)^3) + I((10+x1)^4), data = df, family = "poisson")
all(round(fitted(mod),2) == round(predict(mod, newdata = df, type = "response"),2))
#not ok

#Add fourth transformation with different starting point
mod <- glmmTMB(Y ~ x1 + I((x1+10)^2) + I((10+x1)^3) + I((5+x1)^4), data = df, family = "poisson")
all(round(fitted(mod),2) == round(predict(mod, newdata = df, type = "response"),2))
#ok

#Remove additions from transformations
[...]

Issue #845, opened 2022-08-03
predict() fails with ns() in conditional and dispersion formula.


tjmahr:
Overlaps with #70 and #632.

Here's a minimal example. I use `ns()` in both the conditional formula and dispersion formula with different degrees of freedom. `predict()` doesn't work unless the two `ns()` have different function calls (e.g., ns() versus splines::ns()).

``` r
library(glmmTMB)
library(splines)

x <- glmmTMB(
  am ~ ns(wt, df = 3), 
  dispformula = ~ ns(wt, df = 2), 
  data = mtcars
)
newdata <- data.frame(
  wt = seq(min(mtcars$wt), max(mtcars$wt), length.out = 50)
)
predict(x, newdata = newdata)
#> Error in value[[jj]][ri, ] <- xij: number of items to replace is not a multiple of replacement length

# Don't use the same literal function in the dispformula
x <- glmmTMB(
  am ~ ns(wt, df = 3), 
  dispformula = ~ splines::ns(wt, df = 2), 
  data = mtcars
)
newdata <- data.frame(
  wt = seq(min(mtcars$wt), max(mtcars$wt), length.out = 50)
)
predict(x, newdata = newdata)
#>  [1]  1.001491e+00  9.990882e-01  9.964032e-01  9.931546e-01  9.890604e-01
#>  [6]  9.838390e-01  9.772084e-01  9.688868e-01  9.585924e-01  9.460434e-01
#> [11]  9.309579e-01  9.130541e-01  8.920501e-01  8.676643e-01  8.396146e-01
#> [16]  8.076193e-01  7.713966e-01  7.307445e-01  6.862062e-01  6.387312e-01
#> [21]  5.892730e-01  5.387849e-01  4.882205e-01  4.385332e-01  3.906765e-01
#> [26]  3.456036e-01  3.040200e-01  2.659067e-01  2.311128e-01  1.994874e-01
#> [31]  1.708796e-01  1.451386e-01  1.221134e-01  1.016533e-01  8.360722e-02
#> [36]  6.782437e-02  5.415387e-02  4.244482e-02  3.254636e-02  2.430758e-02
#> [41]  1.757763e-02  1.220560e-02  8.040627e-03  4.931824e-03  2.728309e-03
#> [46]  1.279202e-03  4.336202e-04  4.068282e-05 -5.049172e-05  9.215170e-06

```

[...]

Issue #632, opened 2020-10-16
error for predict with newdata if splines/poly with different df/degree in fixed and random effects


rdiaz02:
If we fit models with splines (say, `ns`) with different df for a term in the fixed and random effects, `predict` gives an error if we pass `newdata`. Same for polynomials with different degree for a term in fixed and random effects. For example:

```
library(glmmTMB)
library(splines)
data(sleepstudy,package="lme4")
m4 <- glmmTMB(Reaction ~ ns(Days, df = 3) + (ns(Days, df = 2)|Subject), 
              data = sleepstudy)
predict(m4, newdata = data.frame(Days = 4, Subject = "372"), re.form = NULL, 
          type = "response")
```

gives error

```
Error in value[[jj]][ri, ] <- xij : 
  number of items to replace is not a multiple of replacement length
```

These two also fail
```
p4 <- glmmTMB(Reaction ~ poly(Days, degree = 3) + (poly(Days, degree = 2)|Subject), 
              data = sleepstudy)
predict(p4, newdata = data.frame(Days = 4, Subject = "372"), re.form = NULL, type = "response")

p5 <- glmmTMB(Reaction ~ poly(Days, degree = 2) + (poly(Days, degree = 1)|Subject), 
              data = sleepstudy)
predict(p5, newdata = data.frame(Days = 4, Subject = "372"), re.form = NULL, type = "response")
```

Note that this (and also for the other fitted models above)
```
predict(m4,  re.form = NULL, type = "response")
```
is OK. 

And this, equivalent to `p5`, but without using `poly` in the random effect specification, also works:
```
p6 <- glmmTMB(Reaction ~ poly(Days, degree = 2) + (Days|Subject), 
              data = sleepstudy)
predict(p6, newdata = data.frame(Days = 4, Subject = "372"), re.form = NULL, type = "response")
```

It is also OK to obtain predictions using `newdata` from models such as

```
m1 <- glmmTMB(Reaction ~ ns(Days, df = 3) + (Days|Subject), 
              data = sleepstudy)
m2 <- glmmTMB(Reaction ~ Days + (ns(Days, df = 3)|Subject), 
              data = sleepstudy)
m3 <- glmmTMB(Reaction ~ ns(Days, df = 2) + (ns(Days, df = 2)|Subject), 
              data = sleepstudy)
predict(m1, newdata = data.frame(Days = 4, Subject = "372"), re.form = NULL, type = "response")
[...]

PR #854, opened 2022-09-11
More predvars

### LINKED ISSUE #70: make sure complex basis prediction works correctly
Complex bases (`poly`, `ns`, `bs`) have to be handled carefully.  This information is captured in the model frame

```
attr(attr(fixedfr,"terms"), "predvars")
```

We should
- detect whether the user is doing something crazy like using _different_ computed bases on the same covariate in the fixed vs random effects
- make sure that we test this (there are test cases in lme4's predict tests)
- for now, defensively warn if predvars are used

bbolker:
I think this is OK, may want to wait for confirmation from issues authors (#632, #845, #853).  More test cases would be good. If we feel sufficiently confident we could also consider this to resolve #70  ...

### DIFF
--- a/glmmTMB/R/predict.R
+++ b/glmmTMB/R/predict.R
@@ -236,10 +236,10 @@ predict.glmmTMB <- function(object,
   tt <- terms(object$modelInfo$allForm$combForm)
   pv <- attr(terms(model.frame(object)),"predvars")
   ## get rid of response variable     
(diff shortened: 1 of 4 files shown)


bbolker:
@mebrooks Still waiting on confirmation from various issue commenters, but I think this is OK to merge ... ?

PR #871, opened 2022-10-18
Rename thetaf

### LINKED ISSUE #870: ranef applied to tweedie or t_family models
If we use `ranef` on a model without random effects, with `family =tweedie` or `t_family`, then it gives this error:
```
Error in ss[rownames(ss) == "b", "Std. Error"] : 
  no 'dimnames' attribute for array
In addition: Warning message:
In summary.sdreport(object$sdr, "random") :
  no or empty summary selected via 'select = %s'"random"
```

I'm not sure if this is a recent change. I came across this while trying `confint` with these types of models. It happens there because `confint` calls `ranef` on every model, even if it doesn't have random effects.

```
set.seed(101)
dd <- data.frame(y = 3 + 5*rt(1000, df = 10))
m1 <- glmmTMB(y ~ 1, family = t_family, data = dd)
ranef(m1)
```

```
library(glmmTMB)
library(tweedie)
nobs = 100; phi = 2; p = 1.7
set.seed(101)
x = exp(runif(nobs, 1, 2))
m = -1.5; b = 10
mu = exp(m*log(x)+b)
y = rtweedie(nobs, mu=mu, phi=phi, power=p)
dat=data.frame(x=x, y=y)
twm = glmmTMB(y ~ log(x), family=tweedie, data=dat)
ranef(twm)
```

bbolker:
This should fix #870 . There is some minor back-compatibility breakage (discussed in the issue and documented in [NEWS](https://github.com/glmmTMB/glmmTMB/blob/rename_thetaf/glmmTMB/inst/NEWS.Rd#L26-L32))


bbolker:
bump ...

Issue #830, opened 2022-06-19
Identifiability check for fixed predictors


stouffer:
In contrast to many other model fitting approaches in R (e.g., `lm` or `glm`), I have noticed that `glmmTMB` does not object to the user specifying a model whose fixed effects create detectable identifiability issues during the model fitting process. Sometimes model fitting appears to complete "successfully" (though parameter SEs are understandably very large). In other instances, the user is only alerted to the possibility of these issues when warnings are raised about convergence issues or non-positive definite Hessian matrices.

The troubleshooting guide says "In general models with non-positive definite Hessian matrices should be excluded from further consideration." I agree with this statement 100%, but feel like the package could go a step further given that some of these issues across fixed effects can be determined prior to model fitting via QR decomposition---as always gets implemented when someone uses `lm` or `glm` to fit their model. Wouldn't it be better to warn the user or to not even attempt to fit the model?

As a simple example:
```
library(glmmTMB)

# create some fake data with two random predictors
n <- 1000
df <- data.frame(
  y = runif(n),
  x1 = runif(n),
  x2 = runif(n)
)
# add a third predictor that is collinear to x1 and x2
df$x3 <- -1.5 * df$x1 + 2 * df$x2

# fit the model with lm
# note: lm refuses to fit a coefficient for x3
m1 <- lm(
  y ~ x1 + x2 + x3,
  data = df
)

# fit the same model with glmmTMB
# note: glmmTMB raises no issues and attempts to fit a model with a priori detectable identifiability issues
m2 <- glmmTMB(
  y ~ x1 + x2 + x3,
  data = df
)

```

bbolker:
#522 ...  this is actually a pretty easy fix (we can steal code from `lme4`), the hardest part is deciding on the interface and checking that all of the downstream methods work appropriately.  Do we automatically fill in `NA` values in the coefficient vector, `vcov()`, `confint()`, etc., or do we do that optionally?  Do we always check for rank-deficiency or can the user opt to skip the check for performance reasons?

stouffer:
I figured as much, and sorry for missing the earlier issue (I searched for everything _except_ "rank-deficiency"...). I think users being able to skip the check is handy. As noted above, I would suggest that models with truly non-identifiable parameters (as opposed to weakly-identifiable ones) not even make it to the model fitting stage: the fastest boost to performance might often be to prevent the optimizer from attempting to solve an unsolvable problem.

[...]

PR #872, opened 2022-10-20
Common implementation for adjusting rank-deficient model matrices X

stouffer:
Major changes:
* Following #830, follows two approaches for rank-deficient matrices depending on whether using "dense" or sparse model matrices `X`.
    * For dense matrices, use `base::qr(X)` and column pivoting determined therein
    * For sparse matrices, drop predictors whose diagonal elements of `Matrix::qr(X)@R` are below specified tolerance
* Added tests for rank-deficient models when using sparse model matrices mirroring existing tests for non-sparse model matrices
* Added tests to ensure that fixed effects estimated with `rank_check='skip'` and `rank_check='warn'` are equivalent.

Still to do but left for a later pull request:
* Implement `why_dropped=TRUE` option that tells the user about which predictors are creating issues of collinearity with those that get dropped as discussed in #830


bbolker:
Wish-list for a later pull request: implement an `napredict()`-like mechanism to fill in `NA` values in the estimated coefficient vector (mimicking `lm`/`glm`-type behaviour). Shouldn't be too hard, although I don't know what it might break downstream ...

PR #875, opened 2022-10-25
Adjust X routine was missing an absolute value

stouffer:


bbolker:
Thanks.  Is there a test for this ... ?

stouffer:
In practice, all tests using sparse matrices would have already applied, I just missed the absolute value that appears in `Matrix::qr2rankMatrix()`.

To the best of my reckoning, it's just a fluke that the random data currently generated happened to always have positive values in `diag(Matrix::qr(X)@R`. I had to try very hard to generate a matrix that would produce a matrix that would return a _large_ negative value. Do you want me to create a new fake dataset that should have previously failed without the absolute value?

bbolker:
If possible, yes. If it seems like a big hassle, no.

stouffer:
I'll see what I can come up with and let you know by the end of the week if I'd prefer to give up.

bbolker:
OK. I think there should be some limit to how much harder you have to work to write a test than to fix the bug in the first place ...

stouffer:
I cannot reproduce whatever black magic I did to get a negative value in the diagonal of the `R` matrix to previously appear. I poked around in the source code of the `Matrix` package but wasn't able to figure out if/where diagonal values might be constrained to be positive (with the sign of corresponding columns of `Q` flipped to match).

But I do know that `Matrix::qr2rankMatrix` includes the absolute value and also warns when `R` does have negative values on the diagonal. Moreover, to justify dropped a column the corresponding value should be close to 0, not extremely large and negative. So I think the change is justified---and correct---even if it covers a situation that may never actually arise in practice.

mebrooks:
Should we merge this? Or wait?

Issue #873, opened 2022-10-23
plan for release 1.1.5?


bbolker:
I think we have enough to start working on a new release (it always takes way longer than we think!)
~~Other than pulling https://github.com/glmmTMB/glmmTMB/pull/871, https://github.com/glmmTMB/glmmTMB/pull/869,~~ I don't think there's anything absolutely release-critical. Getting stuff from #831, #832, #847 in would be nice but not critical ... ?

mebrooks:
Yes, I think it's a good time to do it. 

@mmaechler, based on your email on checking `Matrix 1.5-0`, I updated to the latest version of `Matrix`, updated `TMB`,  added `options(Matrix.warnDeprecatedCoerce = 2) ` to the top line of `glmmTMB/tests/AAAtest-all.R`, then ran `R CMD build glmmTMB`, then ran `R CMD check --as-cran glmmTMB_1.1.4.tar.gz`. Everything seemed to run fine. Is that sufficient?

mmaechler:
Yes, that part definitely is fine!

bbolker:
Just be careful not to leave the `options()` setting there in the production version of the package, unless you reset it at the end of the tests.

mebrooks:
If I find time to do this over the weekend (assuming all rev deps are ok), are we ready for CRAN?

bbolker:
I think a bird in the hand is good; there's always more stuff to fix, but I think we should go for it.

mebrooks:
So far, it passes on all 3 winbuilder versions, as well as "solaris-x86-patched", "debian-gcc-patched", and "debian-gcc-devel". 

I [reported](https://github.com/easystats/modelbased/issues/217) the only problem with a reverse dependency. [Some advice ](https://r-pkgs.org/release.html) says "If any packages fail R CMD check, you should give package authors at least two weeks to fix the problem before you submit your package to CRAN"

strengejacke:
This code crashes R, that's why the package _modelbased_ failed during revdep checks:

```r
library(glmmTMB)

model2 <- glmmTMB(
  count ~ cover + mined + (1 | site),
  ziformula = ~ cover + mined,
  family = truncated_poisson(),
  data = Salamanders
)

stats::predict(
  model2,
  newdata = Salamanders,
  type = "link",
  re.form = NULL,
  unconditional = FALSE,
  allow.new.levels = FALSE
)
```

Is this expected?

bbolker:
No! Nice catch.

[...]

PR #878, opened 2022-11-13
fix log_nzprob for non-fast predictions

bbolker:
I think this fixes the last issue highlighted in #873 (it took some rearranging ...)
Also cleans up a few warnings thrown by `test-checkRank` tests

### DIFF
--- a/glmmTMB/DESCRIPTION
+++ b/glmmTMB/DESCRIPTION
@@ -87,4 +87,4 @@ LazyData: TRUE
 BugReports: https://github.com/glmmTMB/glmmTMB/issues
 NeedsCompilation: yes
 Encoding: UTF-8
-RoxygenNote: 7.2.1
+RoxygenNote: 7.2.2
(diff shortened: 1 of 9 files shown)


mebrooks:
It's interesting, two tests fail on my computer (in a good way) but they don't fail on the automatic testing. 
I think we can just remove these tests since they no longer produce warnings. Is that OK, @bbolker?

```
  ══ Failed ══════════════════════════════════════════════════════════════════════
  ── 1. Failure (test-checkRank.R:125:5): equivalence between 'skip' and 'warn' wh
  `m1 <- glmmTMB(y ~ x1 + x2 + x3 + x4, data = dat, control = glmmTMBControl(rank_check = "skip"))` did not produce any warnings.
  
  ── 2. Failure (test-checkRank.R:147:5): equivalence between 'skip' and 'warn' wh
  `... <- NULL` did not produce any warnings.
  
```

bbolker:
I get convergence warnings on my system. I was trying to skip them (without using `suppressWarnings()`, which is too big a hammer). Will try something else.

bbolker:
Try now?  (There are probably other places in the tests we can replace `suppressWarnings()` with `control = glmmTMBControl(conv_check = 'skip')` ...)

Issue #887, opened 2023-01-27
Response-scale emmeans support


rvlenth:
I am experimenting with adding **emmeans** support with a new option `comp = "response"`. I have been partially successful. Here are the results for several vignette examples:
``` r
library(glmmTMB)
library(emmeans)

packageVersion("glmmTMB")
## [1] '1.1.5'
packageVersion("emmeans")
## [1] '1.8.4'

### models from glmmTMB vignette
Owls <- transform(Owls,
                  Nest=reorder(Nest,NegPerChick),
                  NCalls=SiblingNegotiation,
                  FT=FoodTreatment)

fit_zipoisson <- glmmTMB(NCalls~(FT+ArrivalTime)*SexParent+
                             offset(log(BroodSize))+(1|Nest),
                         data=Owls,
                         ziformula=~1,
                         family=poisson)

fit_zinbinom <- update(fit_zipoisson,family=nbinom2)
fit_zinbinom1 <- update(fit_zipoisson,family=nbinom1)
fit_zinbinom1_bs <- update(fit_zinbinom1,
                           . ~ (FT+ArrivalTime)*SexParent+
                               BroodSize+(1|Nest))
fit_hnbinom1 <-  update(fit_zinbinom1_bs,
                        ziformula=~.,
                        data=Owls,
                        family=truncated_nbinom1)

### Load my own custom recover_data and emm_basis methods
source("~/rvlhome/progs/R/devel/EMMEANS/glmmTMB-support.R")

## Tests of emmeans support for simplest model
rg <- ref_grid(fit_zipoisson, comp = "response")
nd <- transform(rg@grid, Nest = NA)  ## newdata for same combinations

#### compare estimates ####
confint(rg)
##  FT       ArrivalTime SexParent BroodSize prediction    SE  df lower.CL
##  Deprived        24.8 Female         4.39       7.66 0.639 591     6.41
##  Satiated        24.8 Female         4.39       5.73 0.509 591     4.73
##  Deprived        24.8 Male           4.39       7.07 0.563 591     5.96
##  Satiated        24.8 Male           4.39       5.86 0.494 591     4.90
##  upper.CL
##      8.92
##      6.73
##      8.17
##      6.83
## 
## Confidence level used: 0.95
predict(fit_zipoisson, newdata = nd, type = "response", re.form = NA, se.fit = TRUE)
## $fit
## mu_predict mu_predict mu_predict mu_predict 
##   7.661758   5.726675   7.066287   5.864715 
[...]

PR #889, opened 2023-01-31
Enhanced emmeans support and vignette addition

rvlenth:
This PR adds my improved version of **emmeans** support discussed in #887 

Instead of the separate `component` and `mode` arguments I presented there, I decided to just add two more options to `component`: `"response"` and `"cmean"`. These correspond to `type = "response"` and `type = "conditional"` in `predict.glmmTMB`.

I expanded the examples in one of the vignettes, in part to illustrate and explain the confusing thing that happens with hurdle models.

Another note: In the examples, I added `withAutoprint()` in a couple of places, because otherwise we see only the *last* set of results when we run the examples.

### DIFF
--- a/glmmTMB/DESCRIPTION
+++ b/glmmTMB/DESCRIPTION
@@ -88,4 +88,4 @@ LazyData: TRUE
 BugReports: https://github.com/glmmTMB/glmmTMB/issues
 NeedsCompilation: yes
 Encoding: UTF-8
-RoxygenNote: 7.2.2
+RoxygenNote: 7.2.3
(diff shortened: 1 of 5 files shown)


rvlenth:
I made a couple of tweaks in the second commit. There were errors in the examples, and the code still had a `glmmTMB:::` namespace reference that was needed for my local version to work but should not be there when the code is in **glmmTMB**'s namespace.

rvlenth:
FWIW, I have no idea why the checks fail. I keep getting email messages about it, though.

bbolker:
For what it's worth, the tests are failing everywhere, IIRC because of a timeout, which is *probably* driven by a failure to control threading (in BLAS or OpenMP); we've had trouble with that before. I've taken one or two shots at fixing that but with no luck so far.

PR #892, opened 2023-02-21
Denovo simulation

bbolker:


bbolker:
This may be a little premature (I'm not sure everything is named correctly, for example), but I wanted to get some feedback on it.  There's a new `simulate_new()` function that allows `glmmTMB` simulation based only on parameters and a data set with the covariates etc. in it, not requiring that you fit a model first.  There's a vignette ...

mebrooks:
This seems like a useful addition to me and I don't see anything wrong with it. I ran tests and the code added to the sim vignette.

bbolker:
Other possible extensions: for compatibility, `simulate_new` should probably take arguments `seed` and `nsim` ... ?

bbolker:
Did some more work on this, think it can be merged @mebrooks ... ?

mebrooks (changes requested):
Would it make more sense if `simulate_new` took arguments `newdata` and `newpars` like `simulate.merMod` does?

bbolker:
Hmm.  Probably.  In hindsight I might have liked to call the `lme4` arguments `data` and `pars`, but at this point I think you're probably right about consistency.

PR #900, opened 2023-04-07
Modular

bbolker:
extends the modularity of TMB fitting: it doesn't quite work in a piped setup, but this example shows how it can work:

```r
m1 <- glmmTMB(count ~ mined + (1|site),
                  family=poisson, data=Salamanders, doFit=FALSE)
m2 <- fitTMB(m1, doOptim = FALSE)
## could modify the components of m1$env$data at this point ...
## rebuild TMB structure (*may* be necessary)
##      m2 <- with(m2$env,
##                  TMB::MakeADFun(data,
##                                    parameters,
##                                    map = map,
##                                   random = random,
##                                    silent = silent,
##                                    DLL = "glmmTMB"))
m3 <- with(m2, nlminb(par, objective = fn, gr = gr))
m4 <- finalizeTMB(m1, m2, m3)
```

gives a result almost (but not *quite* ...) identical to the all-in-one model


mebrooks:
I think this is valuable since it adds [flexibility](https://bbolker.github.io/mixedmodels-misc/notes/multimember.html). I'm open to merging, if you agree with my edits to the helpfile.

bbolker:
Sure, go ahead.

Issue #902, opened 2023-04-11
deviance residuals


bbolker:
I think this was discussed somewhere, but I'm not sure where.

It would be good if `type = "deviance"` were available. AFAICS all we would have to do to enable this is to store the computed values of `tmp_loglik` (deviance per observation); taking the signed square root would give us *conditional* deviance residuals (i.e., ignoring the contribution of random effects).

bbolker:
started [here](https://github.com/glmmTMB/glmmTMB/compare/deviance_resids) but still have to think/test/figure out whether this is correct or not ...

bbolker:
This should be working for 'vanilla' families at https://github.com/glmmTMB/glmmTMB/tree/deviance_resids

- will work for all families that have `$dev.resids` methods, including `nbinom2`
- **to do**
   - tests
   - residuals are generally *not* taking weights into account.  Fix or at least warn?
   - implement setting dispersion parameter in environment for `nbinom1`; add `$dev.resids` to family
   - document (including that it doesn't work at present for exotic families [Tweedie, COM-Poisson, gen-poisson], and that it returns the deviance based only on the conditional component ...)

bbolker:
This is done, except that deviance residuals for non-canonical families except `nbinom*` (at present: Beta, ordbeta, Tweedie, Student-t, COM-Poisson, generalized Poisson) are all undefined/stubs.

PR #905, opened 2023-04-16
Deviance resids

bbolker:
This implements deviance residuals, and deviance (as sum of squared dev resids), using `family(fitted_model)$dev.resids` &mdash; which is a stub for some families at present.  See https://github.com/glmmTMB/glmmTMB/issues/902 for discussion/details.


glmmTMB/R/methods.R
```diff
@@ -637,19 +652,33 @@ residuals.glmmTMB <- function(object, type=c("response", "pearson", "working"),
...
            },
+           deviance = {
+               if (is.null(dr <- fam$dev.resids)) {
+                   warning(
+                       warningCondition(paste0("deviance residuals undefined for family ",
+                                              sQuote(fam$family),": returning NA"),
+                                        class = c("dev_resids_undefined", "glmmTMB_warn")))
+                   return(rep(NA_real_, length(r)))
+               }
+               d.res <- sqrt(pmax(dr(mr, mu, wts), 0))
```
mebrooks, line 673: In the `nbinom2` case, `dr <- fam$dev.resids` expects 4 arguments, not 3; `theta` is missing.
```
fam$dev.resids
function (y, mu, wt, theta) 
{
    get_nbinom_disp(theta, ".Theta", "theta")
    return(2 * wt * (y * log(pmax(1, y)/mu) - (y + theta) * log((y + 
        theta)/(mu + theta))))
}
```

For example, 
```
> m2=glmmTMB(count~spp, family=nbinom2(), data=Salamanders)
> d2=residuals(m2, type="deviance")
Error in get_nbinom_disp(theta, ".Theta", "theta") : 
  argument "theta" is missing, with no default
```

mebrooks (changes requested):

glmmTMB/R/methods.R
bbolker: Fixed, thanks for the heads-up (I needed a default `NULL` in the argument list).  This works now too:

```r
m3 <- MASS::glm.nb(count~spp,data = Salamanders)
d3 = residuals(m3, type = "deviance")
all.equal(d2, d3, tolerance = 1e-5)
```

mebrooks: Thanks for that. Looks good now :)

Issue #376, opened 2018-08-17
Error with summary()


strengejacke:
I'm trying to "translate" some formulas from [this paper about FE and RE models](https://link.springer.com/article/10.1007%2Fs11135-018-0802-x) into R-syntax, using _glmmTMB_. I'm doing this with some own "real world" sample data (file "glmmtmb_data.RData" attached).

The model computes w/o warning:

```r
m <- glmmTMB(
  QoL ~ time + age + x_tv_dm + x_tv_gm + z1_ti + z2_ti + (1 + time | ID) + (1 + x_tv_dm | ID),
  data = d,
  REML = TRUE
)
```

It also prints well, but using `summary()` gives following error:

```r
summary(m)
#> Error in solve.default(as.matrix(Qm)) : 
#>   system is computationally singular: reciprocal condition number = 2.90076e-17
```

Fitting the same model with `lme4::lmer()` works fine (including `summary()`). I just wonder what does the error message exactly mean and what could be the reason behind it, especially since it works with `lme4::lmer()`.

[glmmtmb_data.zip](https://github.com/glmmTMB/glmmTMB/files/2296580/glmmtmb_data.zip)

bbolker:
The problem happens [here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/methods.R#L245), in `vcov.glmmTMB`.

Essentially, `glmmTMB` is failing to invert the *full* variance-covariance matrix (including both the fixed-effect and the random-effect parameters).  I have to admit I'm not 100% clear on how this chunk is doing what it's doing (computing the marginal of the precision matrix), nor why it's failing/what to do about it ... we could obviously wrap this step in `tryCatch()` and do something if it fails (compute marginal precision only of fixed-effect parameters?), but I don't know what the implications would be.
@kaskr ?

```
  if (REML) {
      ## NOTE: This code would also work in non-REML case provided
      ## that jointPrecision is present in the object.
      Q <- sdr$jointPrecision
      whichNotRandom <- which( ! rownames(Q) %in% c("b", "bzi") )
      Qm <- GMRFmarginal(Q, whichNotRandom)
      cov.all.parms <- solve(as.matrix(Qm))
  } else {
      cov.all.parms <- sdr$cov.fixed
  }
  keepTag <- if (full) { "."
             } else if (!trivialDisp(object)) { "beta*"
             } else "beta($|[^d])"
[...]

PR #908, opened 2023-04-25
return NA for bad Qm inversion (GH #376)

bbolker:


bbolker:
bump.  I think this is relatively harmless (as the title says, it will return a matrix of `NA` values instead of failing when the `Qm` matrix can't be inverted for a REML-fitted model)

bbolker:
thanks!

Issue #831, opened 2022-06-20
handle mapped parameters better in congint


bbolker:
Factorial tests at https://github.com/glmmTMB/glmmTMB/blob/ci_tweaks/misc/test_confint.R

method = {wald, profile, uniroot} × include_mapped = {TRUE, FALSE} × 4 test models × parm = {NULL (all), beta_, theta_}

    wald always OK
    beta_ OK except for uniroot + include_mapped (all models)
    {NULL, theta_} + profile works only for model 4/no-mapped
        fails with 4/mapped (subscript out of bounds)
        fails for other reasons (infinite profile problem) for models 1, 2, 3
    {NULL, theta_} + uniroot fails ('length of dimnames [2] not equal to array extent') as long as include_mapped is TRUE (all models)

Issue #522, opened 2019-10-18
automatically handle rank-deficiency in fixed-effect model matrices?


bbolker:
This would be easy enough to implement, if we want - just need to make sure that we have all the appropriate infrastructure for sticking `NA` values in the right places.  (rank-checking can be slow for very large modl matrices ...)

l-jaye:
Just wondering if there is any progress on this, would be very very very very helpful! :)

l-jaye:
Or is there a way to manually pass a custom model matrix prior fitting the model? For me this is mainly due to missing combinations of some factor levels

bbolker:
Here are some examples of how to deal with problems like this.  (The other is to create an interaction with `interaction()` of your original crossed/incomplete factors, `droplevels()`, and fit a one-way model ... I could also show that)

```
dd <- expand.grid(f1=letters[1:3], f2=LETTERS[1:3], rep=1:10)
set.seed(101)
dd$y <- rnorm(nrow(dd))
dd <- subset(dd, ! (f1=="a" & f2=="A"))
coef(m0 <- lm(y~f1*f2,data=dd))
library(glmmTMB)

m1 <- glmmTMB(y~f1*f2,data=dd)
## Warning message:
## In fitTMB(TMBStruc) :
##   Model convergence problem; extreme or very small eigen values detected. See vignette('troubleshooting')
fixef(m1)$cond

## check and reduce model matrix
X <- model.matrix(~f1*f2,data=dd)
Matrix::rankMatrix(X)<ncol(X)  ## uh-oh

X2 <- lme4:::chkRank.drop.cols(X,kind="message+drop.cols")
attr(X2,"col.dropped") ## f1c:f2C

## by hand
newdata <- data.frame(y=dd$y,X2[,-1])  ## -1 to drop "(Intercept)"
## this should work but doesn't
## also tricky if you want to throw in additional terms such as random effects
## m2 <- glmmTMB(y~., data=newdata)
## more flexible: can add additional terms
form <- reformulate(colnames(X2[,-1]),response="y")
m2 <- update(m1, formula=form, data=newdata)
all.equal(fixef(m2)$cond, c(na.omit(coef(m0))), tolerance=1e-6)  ## TRUE

## OR:
## this almost works ...
m3 <- update(m1, doFit=FALSE)
m3$data.tmb$X <- X2
## we should probably export fitTMB(), but it's not exported ATM
res <- glmmTMB:::fitTMB(m3)
fixef(res)
[...]

PR #911, opened 2023-05-07
Rankdef na

bbolker:
These changes attempt to make handling of rank-deficient fixed effect matrices more complete and more consistent with what happens in base R.

- see `NEWS` for some details (changes default behaviour to automatically drop columns for a rank-def matrix, adds `NA` values into the coefficient vector and the covariance matrix). 
- see #522 for more discussion
- this is backward incompatible because some default behaviours, and some argument names, change. However, it should only be noticeable if people were doing stuff with rank-deficient matrices in the first place ...
- I'm not sure that I'm handling every possible combination of mapped (fixed) parameters and rank-deficient matrices correctly ...
- related: https://github.com/glmmTMB/glmmTMB/issues/831
- some remaining questions about how to handle sparse matrices/whether everything is handled correctly in the sparse-matrix case: https://github.com/glmmTMB/glmmTMB/issues/830


bbolker:
bump ...

mebrooks:
Sorry for the delay.

PR #913, opened 2023-05-17
Augment covstruct vignette

bbolker:
Added a section on how we construct cov matrices and why we need to suppress intercepts; added discussion of why resid variance is suppressed in Toeplitz and comp-symm models; added details of parameterization to covstruct table


bbolker:
thanks!

PR #919, opened 2023-07-04
Mgcv smooths

bbolker:
This is **not** ready for prime time - please don't merge!! But ... I wanted to post this to get comments and feedback on what I've done so far.  Details are [here](https://github.com/glmmTMB/glmmTMB/blob/mgcv_smooths/notes/smooths.rmd), but the basic idea is that I have a *minimal* implementation of the front end where

```r
gtmb1 <- glmmTMB(formula = Reaction ~ s(Days) + (1 | Subject), data = sleepstudy, REML = TRUE)
```

works approximately as expected. (Prediction will *not* work at present; that, and the presentation of the results, will be the hardest part on the back end ...)


mebrooks:
This might overlap with a masters thesis that Hans is imminently planning. We should coordinate and possibly hide this for a little while. I’ll talk to Hans in a few hours.———————————————Sent from a phone. Please excuse any terseness.On 4 Jul 2023, at 02:24, Ben Bolker ***@***.***> wrote:﻿
Assigned #919 to @mebrooks.

skaug:
Thanks, Ben. This is a great starting point for master thesis. Don't let the fact that I will do work prevent you in any way from taking it further in the direction you think is best.

What I can do in the not-to-far future is to check that glmmTMB is giving the same result
as mgcv for a larger suite of datasets. mgcv is not complaining when I try method="ML" 
for the sleepstudy data, but I suspect that it is coming to the same conclusion as glmmTMB,
that the variance of the dummy terms are zero. In general I think these models can be fitted
also with ML.

Another thing that I can do is to write a vignette that explains the glmmTMB output
in terms of Simon Wood's book. For instance, the user must understand that
glmmTMB(formula = Reaction ~ s(Days) + (1 | Subject), data = sleepstudy, REML = FALSE)
estimates a very high smoothing penalty. The model is in effect Reaction ~ Days + (1 | Subject).

skaug:
Another thing is that I think the printout from homdiag is a bit misleading:
 dummy    dummy1         0.8547  0.9245                                     
          dummy2         0.8547  0.9245  0.00                               
          dummy3         0.8547  0.9245  0.00 0.00                          
          dummy4         0.8547  0.9245  0.00 0.00 0.00                     
          dummy5         0.8547  0.9245  0.00 0.00 0.00 0.00                
          dummy6         0.8547  0.9245  0.00 0.00 0.00 0.00 0.00           
          dummy7         0.8547  0.9245  0.00 0.00 0.00 0.00 0.00 0.00      
          dummy8         0.8547  0.9245  0.00 0.00 0.00 0.00 0.00 0.00 0.00

We are in effect only estimating a single variance parameter  (0.8547)

bbolker:
Yes, "fix up VarCorr printing" was on the TODO list already ... (it's ugly for most var-corr types other than the default unstructured ...)

mebrooks:
I'm testing and editing the `NEWS` file. How would one look at `notes/smooths.rmd` in the package?

bbolker:
Yes, I think that's a good idea. The smooths *should* work 'out of the box' in one sense: the big remaining problems are

* starting values - we definitely have to worry about how to do this properly, in at least one example the default starting values don't work (and may even not-work silently, in a worst-case scenario)
* printing/interpreting the output (`predict()` does work, but I don't think there's an easy way to separate out the effect of a smooth)
* at present we can't use smooth terms in dispersion models (because we don't allow random effects in the dispersion model, and that's how smooths are implemented ...)

mebrooks:
How about this?
```
      \item an experimental implementation of splines is available, 
      based on \code{mgcv}. See example in \code{?glmmTMB}.  
      Constructive feedback is welcome at GitHub issue #919.
```

mebrooks:
Sorry, I should have done the rebase in a separate commit.

bbolker:
You didn't break the tests, at least ...

PR #921, opened 2023-07-11
Lognorm dist

bbolker:
This is **not** necessarily ready to merge, but I wanted to open it so it didn't get lost again.

(Can someone remind me what the use case is for this distribution?)


bbolker:
Now that I'm thinking about this, I'm not even sure it makes sense.  The idea was (I think???) to parameterize the lognormal via its mean on the data scale, but now I'm not even sure how this makes sense in a non-trivial case.  Please remind me ...

skaug:
I have not followed the previous discussion, so I might be
missing the point, but Wikipedia lists many application areas:
https://en.wikipedia.org/wiki/Log-normal_distribution#Occurrence_and_applications
	
Regarding the question about "which scale" 
my understanding is that the unifying concept of glmmTMB is
that the mean refers to the data scale. For the log normal, 
the mean-variance relationship is: 
var = phi*mean^2, where phi is the dispersion parameter, given
in terms of the underlying SD sigma, as phi = exp(sigma^2)-1

glmmTMB/src/glmmTMB.cpp
```diff
@@ -772,6 +775,19 @@ Type objective_function<Type>::operator() ()
...
+      case lognormal_family:
+	// FIXME: documentation??
+	// parameterized in terms of mean and SD on *data* scale, i.e.
+	// mu = exp(logmu + logsd^2/2)
+	// sd = sqrt((exp(logsd^2)-1)*exp(2*logmu + logsd^2)) = mu*sqrt(exp(logsd^2)-1)
+	// 1+(sd/mu)^2 = exp(logsd^2)
+	// logsd = sqrt(log(1+(sd/mu)^2))
+	// logmu = log(mu)- 
+        s1 = log(1 + pow(phi(i)/mu(i), 2.0));
+        s2 = sqrt(s1);
```
mebrooks, line 787: I think `s2` is the sd of the normal distribution, so it should be the 3rd argument to `dnorm`. 

I think `s3` agrees with what I have for the mean of the normal distribution in [a related application](https://github.com/mebrooks/StockRecruit/blob/b0ffb731c24a1ad6133829ac11194bb1353bb53f/StockRecruitSET/R/simulateMethods.R#L45) that I thought very carefully about, many years ago.

mebrooks:
I simplified the test a lot to make it work. It would be good to come up with one that uses a covariate, but I'm done for today.

glmmTMB/src/glmmTMB.cpp
```diff
@@ -772,6 +775,19 @@ Type objective_function<Type>::operator() ()
...
+        s3 = log(mu(i)*mu(i)) - log(mu(i)*mu(i) + phi(i)*phi(i))/Type(2.0); //from Wikipedia
```
mebrooks, line 788: I don't think I needed to change this. I was just trying everything I could think of. We should consider how to write it to minimize overflow (probably not this way).

bbolker:
I think simplifying the test to make it work was exactly the right thing to do. The thing that was confusing me yesterday is that (IIUIC), the *only* model that we can easily get to correspond between normal-scale and log-scale is the intercept-only model ...

bbolker:
PS except for swapping s2/s3, my expressions were actually correct (and a little simpler, so I reverted to them).

mmaechler (dismissed):
In the case   sd/mu is really small,   log(1 + (sd/mu)^2))   should be replaced by
log1p( (sd/mu)^2 )   in  's1 = ....'   (currently line 786)

bbolker:
@mebrooks, if you had a current use for this, could you try it out and see if anything breaks before we merge? (We should update documentation and `NEWS` in any case ...)

mebrooks:
If we decide to change `phi` for the Gaussian family, I should redo one of the videos that @skaug and I recorded but haven't released yet. In the video, I demonstrate models on the dispersion parameter using an example that has a Gaussian distribution, to keep it simple. I don't mind redoing it (I have access to a studio); we just need to decide in about a week, before the videos are released.
```
library(glmmTMB)
library(ggplot2); theme_set(theme_bw())

#simulate data
sim1 = function(nID=1000, nt=10, IDsd=0.5, tsd=0.25, mu=0, residsd=1)
{
	dat = expand.grid(ID=1:nID, t=1:nt)
	n = nrow(dat)
	dat$REID = rnorm(nID, sd=IDsd)[dat$ID]
	dat$REt = rnorm(nt, sd=tsd)[dat$t]
	dat$y = rnorm(n, mean=mu, sd=residsd) + dat$REID + dat$REt
	dat
}
set.seed(101) 
d1 = sim1(mu=100, residsd=10)
set.seed(101) # to get same ID and t random effects
d2 = sim1(mu=200, residsd=5)
d1$treatment = "ten"
d2$treatment = "five"
dat = rbind(d1, d2)

# plot the data
ggplot(dat)+geom_histogram(aes(y))+ylab(NULL)
ggplot(dat)+geom_histogram(aes(y, fill=treatment))+ylab(NULL)

# model
m0 = glmmTMB(y ~ treatment + (1|t)+(1|ID), 
		dispformula=~treatment, data=dat)

# extract results
fixef(m0)$disp
c(log(5^2), log(10^2)-log(5^2)) # expected dispersion model coefficients

fixef(m0)
```

mebrooks:
In the initial setup, @mmaechler made an effort to maintain consistency with some existing R methods. Does that ring a bell for anyone?

PR #929, opened 2023-07-30
Priors

bbolker:
this is **not tested at all** (except for not breaking current tests) but in principle is a basic implementation of priors/regularization. See [this document](https://github.com/glmmTMB/glmmTMB/blob/master/notes/priors.Rmd) for design thoughts ...


bbolker:
This is now feature-complete, I think (normal, t, Cauchy priors for fixed effects; gamma priors on the SD scale for SDs; LKJ priors for correlations), but largely untested.

The only obvious/known thing that's missing is that the print method doesn't include information about specific elements (i.e., only gives the distribution and the overall component (fixef/ranef), not the particular parts of that component (if selected).

A vignette has been started but could be extended/improved.

strengejacke:
Is there a specific schedule for this PR? I think this would be nice to have in _glmmTMB_, to address some of the issues that _may_ occur with frequentist models.

One question, in the docs you write that the MAP instead of posterior mean or median will be calculated - can this be defined by the user, or is there no way to avoid using the MAP? I guess in most cases, the posterior will roughly be normally distributed (or nnot?), so it doesn't matter much, but for [this example](https://easystats.github.io/bayestestR/#point-estimates), MAP, median and mean are "quite differen".

bbolker:
Huh, I had forgotten that this was so far along, TBH.  If @mebrooks or @kaskr were OK with it I think we could merge it, marked "experimental"? (That's a good way to get people to find out what breaks ...)

As far as MAP goes -- as far as I know that's the only option that is theoretically available if all you have is an optimizer (we could use Laplace approx to integrate over *Gaussian* priors only ...).  If you want other point estimates I think you're stuck feeding the `TMB` object to `tmbstan` ...

bbolker:
PS there are a lot of conflicts, but almost all of them are in derived files -- could probably fix this by resolving the conflicts in the `.bib`/`.R`/`.yml` files (merging main into this branch), then rebuilding stuff.

bbolker:
Note that CI build is failing because of upstream binary incompatibility (see [here](https://github.com/glmmTMB/glmmTMB/issues/965), [here](https://github.com/lme4/lme4/issues/756) ...) ... sigh ...

bbolker:
@strengejacke , I've brought this up to date with master.  It could always be improved (TESTS!), but do you see anything that would be high priority to add/fix before merging into master as an experimental feature ... ?

strengejacke:
I'm installing the latest version from this branch and will check - I think checking 80 revised files might be too confusing. ;-)

strengejacke:
A first quick glance, and things look good to me:

``` r
library(glmmTMB)

priors1 <- data.frame(
  prior = "normal(0,2)",
  class = "fixef"
)
priors2 <- data.frame(
  prior = c("normal(0,2)", "cauchy(0,0.5)"),
  class = c("fixef", "ranef")
)
priors3 <- data.frame(
  prior = "normal(0,2)",
  class = "ranef"
)
m1 <- glmmTMB(count ~ mined + (1 | site),
  zi = ~mined,
  family = poisson,
  data = Salamanders,
  priors = priors1
)
m2 <- glmmTMB(count ~ mined + (1 | site),
  zi = ~mined,
  family = poisson,
  data = Salamanders,
  priors = priors2
)
m3 <- glmmTMB(count ~ mined + (1 | site),
  zi = ~mined,
  family = poisson, data = Salamanders,
  priors = priors3
)
parameters::compare_parameters(m1, m2, m3, effects = "all")
#> # Fixed Effects
#> 
#> Parameter   |                 m1 |                 m2 |                 m3
#> --------------------------------------------------------------------------
#> (Intercept) | 0.10 (-0.34, 0.54) | 0.03 (-0.48, 0.54) | 0.07 (-0.39, 0.54)
#> mined (no)  | 1.13 ( 0.66, 1.59) | 1.19 ( 0.66, 1.73) | 1.16 ( 0.66, 1.65)
#> 
#> # Random Effects
#> 
#> Parameter            |                 m1 |                 m2 |                 m3
#> -----------------------------------------------------------------------------------
#> SD (Intercept: site) | 0.28 ( 0.14, 0.54) | 0.33 ( 0.16, 0.69) | 0.29 ( 0.15, 0.57)
```

<sup>Created on 2023-12-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I would probably format the docs for `?priors` and add examples (maybe the above one, using `summary()`?).

strengejacke:
This example is failing, not sure if this is supposed to not to work?

```r
library(glmmTMB)

priors3 <- data.frame(
  prior = "t(1,1)",
  class = "fixef"
)
m3 <- glmmTMB(count ~ mined + (1 | site),
  zi = ~mined,
  family = poisson, data = Salamanders,
  priors = priors3
)
```

strengejacke:
So some additional points that would be nice to have so far:

- validate data frame input of priors (check whether defined distributions contains all necessary arguments)
- docs could be easier to read, maybe using bullet points/lists for the description of the different distributions
- in the printed summary, make sure that prior information can be easily associated with the related parameters

strengejacke:
It would also be great to have an example that addresses particular issues, like convergence or singular fits. How can I put a prior on particular coefficients only for the fixed and random effects?

This example shows that convergence or singular fits can be addressed using priors:

``` r
library(glmmTMB)
data(cbpp, package = "lme4")
set.seed(1)
cbpp$x <- rnorm(nrow(cbpp))
cbpp$x2 <- runif(nrow(cbpp))

# convergence issue, singular fit
model <- glmmTMB(
  cbind(incidence, size - incidence) ~ period + x + x2 + (1 + x | herd),
  data = cbpp,
  family = binomial()
)
#> Warning in finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old): Model convergence
#> problem; singular convergence (7). See vignette('troubleshooting'),
#> help('diagnose')

priors <- data.frame(
  prior = c("t(0,2.5,3)", "t(0,2.5,3)"),
  class = c("fixef", "ranef_sd")
)
# no warnings
model2 <- glmmTMB(
  cbind(incidence, size - incidence) ~ period + x + x2 + (1 + x | herd),
  data = cbpp,
  priors = priors,
  family = binomial()
)

summary(model)
#>  Family: binomial  ( logit )
#> Formula:          
#> cbind(incidence, size - incidence) ~ period + x + x2 + (1 + x |      herd)
#> Data: cbpp
#> 
#>      AIC      BIC   logLik deviance df.resid 
#>    197.7    215.9    -89.8    179.7       47 
#> 
#> Random effects:
#> 
#> Conditional model:
#>  Groups Name        Variance Std.Dev. Corr 
#>  herd   (Intercept) 0.339584 0.58274       
#>         x           0.009954 0.09977  1.00 
#> Number of obs: 56, groups:  herd, 15
#> 
#> Conditional model:
#>             Estimate Std. Error z value Pr(>|z|)    
#> (Intercept)  -1.6711     0.3279  -5.096 3.47e-07 ***
#> period2      -1.0090     0.3230  -3.123  0.00179 ** 
#> period3      -1.0357     0.3299  -3.140  0.00169 ** 
#> period4      -1.3829     0.4514  -3.063  0.00219 ** 
#> x            -0.3773     0.1937  -1.948  0.05146 .  
#> x2            0.4279     0.5121   0.836  0.40342    
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
summary(model2)
#>  Family: binomial  ( logit )
#> Formula:          
#> cbind(incidence, size - incidence) ~ period + x + x2 + (1 + x |      herd)
#> Data: cbpp
#> 
#>      AIC      BIC   logLik deviance df.resid 
#>    234.4    252.7   -108.2    216.4       47 
#> 
#> Random effects:
#> 
#> Conditional model:
#>  Groups Name        Variance Std.Dev. Corr 
#>  herd   (Intercept) 0.3360   0.5796        
#>         x           0.0468   0.2163   0.30 
#> Number of obs: 56, groups:  herd, 15
#> 
#> Conditional model:
#>             Estimate Std. Error z value Pr(>|z|)    
#> (Intercept)  -1.6453     0.3229  -5.095 3.49e-07 ***
#> period2      -0.9453     0.3291  -2.872  0.00408 ** 
#> period3      -1.0043     0.3349  -2.998  0.00271 ** 
#> period4      -1.3522     0.4469  -3.026  0.00248 ** 
#> x            -0.3462     0.2051  -1.688  0.09139 .  
#> x2            0.3146     0.5108   0.616  0.53789    
#> ---
#> Signif. codes:  0 '***' 0.001 '**' 0.01 '*' 0.05 '.' 0.1 ' ' 1
#> 
#> Priors:
#> fixef ~ t(0, 2.5, 3)
#> ranef_sd ~ t(0, 2.5, 3)
```

<sup>Created on 2023-12-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


However, if I want to have a prior for `x` only, how do I do that? `x` is also specified as random slope, maybe that doesn't count as "RE"?

``` r
library(glmmTMB)
data(cbpp, package = "lme4")
set.seed(1)
cbpp$x <- rnorm(nrow(cbpp))
cbpp$x2 <- runif(nrow(cbpp))

priors <- data.frame(
  prior = c("t(0,2.5,3)", "t(0,2.5,3)"),
  class = c("fixef", "ranef_sd"),
  coef = c("x", "x")
)
# no warnings
model2 <- glmmTMB(
  cbind(incidence, size - incidence) ~ period + x + x2 + (1 + x | herd),
  data = cbpp,
  priors = priors,
  family = binomial()
)
#> Error in proc_priors(priors, info = list(fix = fix_nms, re = re_info)): can't match prior RE component x

priors <- data.frame(
  prior = c("t(0,2.5,3)", "t(0,2.5,3)"),
  class = c("fixef", "ranef"),
  coef = c("x", "herd")
)
# no warnings
model2 <- glmmTMB(
  cbind(incidence, size - incidence) ~ period + x + x2 + (1 + x | herd),
  data = cbpp,
  priors = priors,
  family = binomial()
)
#> Error in proc_priors(priors, info = list(fix = fix_nms, re = re_info)): can't match prior RE component herd
```

<sup>Created on 2023-12-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

bbolker:
FWIW all of @strengejacke's original concerns have been addressed.

[ ] better documentation of scales on which parameters are applied
[ ] implement Beta priors for individual correlation elements?
[ ] applying priors to individual elements of fixed-effect parameter vectors is OK, but how should individual elements of random effects be specified? currently the `coef` component of the prior specification can be either the grouping variable alone or the full RE term (e.g., `Subject` or `Days|Subject`). Should  `<grpvar>` alone throwing an error if there is more than one RE term with the same grouping variable? What about specifying  `individual component|<grpvar>`;  what is `individual component` in this case? if it's a single value then it's a SD, if it's a pair of terms (e.g. `(Intercept).Days`) then it's a correlation coefficient? (In principle we may want to allow imposing a Beta prior on an individual element of the correlation matrix ... ??)
[ ] allow half-t and half-Normal priors on SDs?
[ ] it would be convenient to have a specification for "all fixed effects *except* the intercept"

strengejacke:
Tagging @bwiernik WDYT?

bwiernik:
> it would be convenient to have a specification for "all fixed effects except the intercept"

I think that this should be the default. Having the intercept treated like all other fixed effects sounds like a problem in general. 

This is what brms does:

> In case of the default intercept parameterization (discussed in the 'Details' section of brmsformula), general priors on class "b" will not affect the intercept. Instead, the intercept has its own parameter class named `"Intercept" `and priors can thus be specified via `set_prior("<prior>", class = "Intercept")`. Setting a prior on the intercept will not break vectorization of the other population-level effects. Note that technically, this prior is set on an intercept that results when internally centering all population-level predictors around zero to improve sampling efficiency. On this centered intercept, specifying a prior is actually much easier and intuitive than on the original intercept, since the former represents the expected response value when all predictors are at their means. To treat the intercept as an ordinary population-level effect and avoid the centering parameterization, use `0 + Intercept` on the right-hand side of the model formula.

mebrooks:
This branch passes tests, but there are still conflicts with the master branch (according to GitHub).

bbolker:
@bwiernik, @strengejacke should the specification that "fixef" means "all values except the intercept" apply to all model components (conditional, zero-inflation, dispersion) or just to the conditional distribution? (I can see arguments either way ...) (the former is implemented in [1450413](https://github.com/glmmTMB/glmmTMB/pull/929/commits/145041319f8f9ac7048fde1999bb5c3b8d2eef0f)

strengejacke:
I have no strong opinion (and not much experience with setting priors on different components and in particular which priors make sense for the different intercepts), but I think for reasons of consistency, it seems more plausible that "all values except the intercept" would apply to all model components. You would have classes `intercept`, `zi_intercept`, `disp_intercept` etc.?

bbolker:
Applying to all components makes sense.  I was thinking of implementing the intercept-specific prior as `c(<distribution + params>, "fixef", "(Intercept)")` rather than having a separate "intercept" class (seems equally clear and handled by the current implementation without adding extra stuff). (Second element would be "fixef_zi" or "fixef_disp" to specify other model components.)

Note that we don't do any of the internal centering/scaling stuff that `brms()` does (maybe we should, but we'll cross that bridge later ...)

strengejacke:
> I was thinking of implementing the intercept-specific prior as c(prior = <distribution + params>, "fixef", "(Intercept)") rather than having a separate "intercept" class 

Yes, sounds good!

> Applying to all components makes sense.

Just to clarify, I did not mean that a prior is applied to _all_ components at once, but rather that the "everything but intercept" rule should be used for all components.

strengejacke:
>  What is your opinion about the wisdom of pulling this branch before release

If I'm not mistaken, this PR would not break any existing code, so this is rather an "additional" feature that does not affect previous "user experience". You _could_ merge this PR, flagging this as still experimental in the docs (and maybe a message/warning whenever priors are used), so users are aware that's it's not completely tested - but you would get the chance that this feature is visible/accessible to a wider user base and therefore more potential users would probably test the prior-implementation and give feedback.

But that's something you should decide, what fits best to your "release philosophy".

bwiernik:
I agree with the discussion above.

bbolker:
I think all in all it would be good to put this out there. My main concern isn't breaking existing code (as discussed, these are brand-new features), but rather releasing an incompletely baked API that may need to change in the future....

strengejacke:
> but rather releasing an incompletely baked API that may need to change in the future

Personally, I would be less concerned (and merge this PR). You can flag this as experimental (docs, msg/warn when used) and indicate that the API may change in the future. I don't know to what extent _brms_ can serve as a "blueprint" for the prior-implementation (regarding API) and thus, to which extent you can arrive with a "fully baked" API on release - if you think, this is possible, than you could probably wait. Else, you could use this experimental feature to gain practical experience - and adapt the API or keep it (if it turns out that way).

The question is, which disadvantages do you expect when this PR is included in the next release? Casual users will probably not notice this change and their life is not affected, more advanced/technically skilled users would try this out, knowing that the API might change. And even if it does, to which extent would their work be affected? Are the _results_ unreliable, or do you have to update your scripts because the API changes? If only the latter, that's nothing of scientific concern, I'd say.

mebrooks:
Since CRAN wants the new version a.s.a.p. and this could potentially slow that down, I'm tempted to send the current master branch to CRAN and then merge in this branch, so that it gets some more testing by people who use the GitHub version of the package before it goes out to the bigger user group. However, @bbolker I could be convinced to go ahead and merge if you think it's best. CRAN wrote again and said the date we would get archived is the 31st, so we have a little time.

bbolker:
I'm OK either way, but I think it would be safe  (and I would slightly prefer) to merge this PR immediately. The only issue is how much of a pain/delay it will be to do a final round of testing before submission ... (I'm doing my own revdep run, which will take about 90 minutes.  I could merge locally and try it again ...)

mebrooks:
I guess we save some time in the long run by merging now. I'll do testing now and start writing to maintainers.

bwiernik:
Very excited about this !

PR #935, opened 2023-09-04
Remove old junk

kaskr:


bbolker:
Right now the `DESCRIPTION` file says we need `TMB` version >= 1.9.0.  Does this need to be bumped?

PR #956, opened 2023-10-22
lognormal-hurdle

bbolker:


mebrooks:
Wow thanks! That was fast! I'm going to use it in a demo this morning to show when the Tweedie is better, via `AICctab`. (It's 6am in Japan, where I am now.)

Issue #955, opened 2023-10-12
Gaussian dispformula specification with SD vs variance


bwiernik:
Is there any way to fit the linear model in `dispformula` to `log(sqrt(V))` rather than `log(V)`? The immediate practical interest is in comparing results with other packages that specify Gaussian dispersion in terms of sigma rather than sigma^2.

If not, would you be open to a PR added a `gaussian2()` family that uses an SD dispersion specification rather than a variance dispersion specificiation?

ASKurz:
Yes please.

bbolker:
This was discussed somewhere recently, but I can't find it (@mebrooks, @kaskr, do you remember?).  I think that this would be automatically handled by changing the internal specification, which (IIRC) is *different* for Gaussian than for any of the other distributions, to be more consistent with the other distributions. The only reasons *not* to do this (again, IIRC) are

* breaking backward compatibility (I don't know whether it would be useful to add an option for back-compatibility somewhere?)
* "Chesterton's fence", i.e. trying to understand/remember what the original reason was for parameterizing the Gaussian differently from the other distributions/families ...

I would personally prefer to make this change than to add another family.

mebrooks:
We were discussing it on the `lognorm_dist` branch, but I think the discussion may have disappeared when we merged and deleted the branch (slightly annoying GitHub behavior). The things I know that would be broken are a test and a teaching video that use something similar to [this example](https://github.com/glmmTMB/glmmTMB/blob/bd625cad51392fee7086b6c832c6bfdd8a9114d4/glmmTMB/R/glmmTMB.R#L1037). I think @skaug is currently using the teaching video in a private class now and then we plan to share them openly. So I'd need to re-record the video (at DTU instead of UiB), but I don't mind doing that in early 2024.

bwiernik:
Here is that discussion https://github.com/glmmTMB/glmmTMB/pull/921

bwiernik:
The only thing that comes to mind regarding why the variance might have been chosen is that the base `gaussian()` family specifies `variance` as a unit vector with length equal to `length(mean)`

bbolker:
@kaskr @skaug @mebrooks should I go ahead and make a branch that does this?

mebrooks:
I'm OK with the change if everyone else is. Any other opinions? @mmaechler ?

bbolker:
[...]

Issue #906, opened 2023-04-21
System computationally singular error produced in summary() for a glmmTMB model


JuanWang814:
Here are minimised data I randomly sampled from my full dataset, which allows me to reproduce the issue:
```
data <- structure(list(a = c(181.25, 181.25, 183.5, 184, 180.25, 183.75, 
177, 182.25, 184, 185, 185.25, 187, 178, 181.5, 185.5, 187.25, 
187, 184.25, 181.25, 185.25, 177.5, 178, 186.5, 186.75, 181.5, 
178.25, 186.5, 185, 178.25, 188.75, 185.25, 187.5, 187.5, 180.5, 
178, 183.5, 179.5, 184.25, 186, 184.25, 184, 180.5, 185.5, 180.5, 
178.75, 183.5, 183, 179.75, 185, 185.25, 188, 182, 180, 179.5, 
188.5, 184.25, 181.5, 183.5, 185.25, 181.25, 183.5, 178, 177, 
186.5, 184.75, 177.5, 182.75, 181, 178.75, 180.5, 178.25, 177.75, 
186, 181.75, 181.25, 178.25, 179.5, 178, 181, 181.5, 181.25, 
185, 179.75, 179.25, 188.75, 181.5, 183.25, 186, 185, 183.5, 
180.25, 183.25, 179.5, 180.25, 181.25, 180.75, 181.25, 182.5, 
186, 186.5, 179, 183, 186, 180.5, 178.5, 178.75, 180, 187.25, 
182.75, 178, 180.25, 178, 183.5, 184.75, 184.5, 182.75, 186.75, 
180, 180.75, 179.25, 177.25, 178, 178.25, 179.75, 181.25, 179.5, 
183, 187.5, 183, 179.25, 188, 190, 184, 181.25, 179.25, 183, 
184, 178, 180.75, 185.5, 185.5, 180, 180.25, 184.25, 180, 181, 
187, 185.25, 179.25, 181.5, 177.25, 186.75, 179.25, 180.75, 187, 
181.25, 181, 183.5, 182.25, 178.75, 181.5, 178.5, 181.75, 182.75, 
188.5, 184.25, 188, 185, 185, 184.5, 186.25, 179.5, 186.5, 180.5, 
187, 181.5, 180.25, 185, 186.75, 183, 177.25, 180, 184.25, 184.75, 
182, 186.75, 181, 178.5, 186.25, 180, 178.75, 185.5, 179.75, 
[...]

PR #973, opened 2023-12-25
Reparam gauss family

bbolker:
See https://github.com/glmmTMB/glmmTMB/issues/955


glmmTMB/R/glmmTMB.R
bbolker, line 1: good catch!

glmmTMB/tests/testthat/test-methods.R
```diff
@@ -654,18 +659,18 @@ test_that("weighted residuals", {
...
-        QoL ~ time + age + x_tv_dm + x_tv_gm + z1_ti + z2_ti + (1 + time | ID) + (1 + x_tv_dm | ID),
-        data = d,
-        REML = TRUE
-    )
-    ## only fails on some platforms ... this is sufficient for now ... FIXME
-    if (getRversion() >= "4.3.0") {
-        expect_true(all(is.na(vcov(m)$cond)))
-    }
-})
+# This test started also giving a warning on os "mac".
```
mebrooks, line 662: @bbolker do you think we should keep this? It seems like maybe it's more trouble than it's worth.

bbolker: it's certainly problematic ...  the original issues are [here](https://github.com/glmmTMB/glmmTMB/issues/376) and [here](https://github.com/glmmTMB/glmmTMB/issues/906); the latter seems like it might be a more "robustly bad" example (i.e., likely to fail more broadly ... ??). (The behaviour we're hoping to reproduce is a model that runs *without* warnings an errors, but where inverting the Hessian fails [which now gives NA values rather than an error] ...)  I just added an example in `misc/glmmTMB_GH906.R` -- but I think it would be OK, now that you've commented this test out here, to add something to the general issues list, and add this back once we've tested to see if that example is reliable across platforms/versions ...

Issue #1009, opened 2024-03-31
Is there an option to constrain two fixed effects to be equal when estimated?


odemodelselection:
Let's say I have the following model:
```
y = b0 + b1*x1 + b2*x2 + b3*x3 + b4*x4
```

and I assume that -b1=b3, b2=-b4.
Is there an option using "map=" to introduce such constraints, aka:
```
model <- glmmTMB(y ~ x1+ x2 + x3+ x4 + (1 + x1+ x2 + x3+ x4|subject),
                                  map=(....),
                                  data=data)
```
and variances of b1,b3 and b2,b4 should also be equal.

bbolker:
In theory this is possible.  You'll need to negate some of the predictor variables (we can use `map` to set parameters to be *equal to* each other, but not any more general constraints).  Something like this:

```r
ncorr <- 5*4/2 ## number of correlation parameters
glmmTMB(y ~ 1 + I(-x1) + I(-x2) + x3 + x4 + (1 + x1+ x2 + x3+ x4|subject),
    map = list(beta = factor(c(1,2,3,2,3), 
                       theta = factor(c(1,2,3,2,3, 3+seq(ncorr))),
    ...)
```

odemodelselection:
The following code works perfectly
```
ncorr <- 5*4/2 ## number of correlation parameters
model <- glmmTMB(y ~ 1 + I(-G11) + I(-G12) + G21 + G22 + (1 + G11+ G12 + G21+ G22|subject),
                 map = list(beta = factor(c(1,2,3,2,3)), 
                                 ## first 5 parameters are the log-SDs of the variance components;
                                 ## remainder are the correlation parameters.  See
                                 ## https://cran.r-project.org/web/packages/glmmTMB/vignettes/covstruct.html#mappings
                                 theta = factor(c(1,2,3,2,3, 3+seq(ncorr)))
                                 ),
                  data = X_hat)
```
Result:
```
> model
Formula:          y ~ 1 + I(-G11) + I(-G12) + G21 + G22 + (1 + G11 + G12 + G21 +  
    G22 | subject)
Data: X_hat
      AIC       BIC    logLik  df.resid 
 49181.36  49315.72 -24573.68     19983 
Random-effects (co)variances:

Conditional model:
 Groups   Name        Std.Dev. Corr                    
 subject  (Intercept) 0.2807                           
          G11         0.6041   -0.40                   
          G12         0.7098    0.39 -0.87             
[...]

PR #1010, opened 2024-04-02
Mapped pairs

bbolker:
This definitely needs more testing, and a NEWS item, but these are the minimal changes that seem to be needed when `map` specifies sets of parameters *set equal to each other*, rather than when it is used (with `NA`) to fix parameters to their starting values.  See #1009.


bbolker:
`vcov(., full = TRUE)` isn't fully working for this case.

- do we want to allow the option of including or *not* including the mapped-equal parameters?
- should we go to the effort of duplicating elements in the cov matrix for this case?

bbolker:
If we do merge this PR (which is probably worthwhile) we should probably still leave the related [issue](https://github.com/glmmTMB/glmmTMB/issues/1009) open so we know that there are some unresolved issues with `vcov(., full = TRUE)` in this case ...

PR #1011, opened 2024-04-02
update TMB version check

bbolker:
this moves the TMB version check from a test of an installed file to checking against an environment variable (as suggested by Mikael Jagan). Also, no longer checking the Matrix version, as glmmTMB doesn't depend directly on the Matrix ABI, only via TMB.


kaskr:
Plan is, that from TMB version 1.9.12 the ABI version can be obtained by `TMB.Version()$abi`.
ABI version for older versions of TMB can be obtained by copy-pasting `abi()` function from here:
https://github.com/kaskr/adcomp/blob/master/TMB/R/abi_table.R

PR #1020, opened 2024-04-05
Inconsistency in priors vignette

### LINKED ISSUE #1017: Inconsistency in priors vignette
In the priors vignette, priors are created using `"beta"` and `"theta"` as `class` values. However, these are not documented in `?glmmTMB::priors`. Maybe you either add those aliases, or if you want to keep them undocumented, I suggest using officially documented values in the vignette (or at least mention that these are undocumented aliases).

strengejacke:
Fixes #1017


strengejacke:
Addressing #1017, and "lintering" some code chunks.

PR #1022, opened 2024-04-10
Mvlgcp

ElliotDovers:


bbolker:
Hi. Can you clarify what's going on here? Was this a mistake, or do you have a bunch of changes you think should be integrated into `glmmTMB` ... ??

ElliotDovers:
Hi @bbolker, clarifying this pull request was indeed a mistake (sorry about that!). Feel free to disregard/delete (I thought I had done the latter).

For context, this work extends `glmmTMB` to fit uni- and multi-variate point patterns via log Gaussian Cox processes. However, as I have set this up as a wrapper function, `mvlgcp()`, I figured this might not be in line with the `glmmTMB` ethos and so I haven't considered suggesting it be added to package yet. There are also a few other bespoke functions needed - *e.g.* for creating basis functions to approximate latent fields - which are all contained in `R/mvlgcp.R`. If you are inclined, you can find this package extension at [https://github.com/eco-stats/glmmTMB/tree/mvlgcp](https://github.com/eco-stats/glmmTMB/tree/mvlgcp) - I have documented it so there are examples for `mvlgcp()` to try (I have also added a couple of datasets).

If you are interested in adding the ability to fit these types of models, I would be happy to discuss how this could be done in a way that perhaps better suits `glmmTMB`. Currently, the wrapper function sets up a dummy random variable structure (with `doFit = F`) and then replaces the relevant part of the Z matrix with the appropriate basis function matrix - so this is effectively fitting a smoother as in `mgcv`. I suppose a way forward would be to create a `s()`-style random effect structure instead - which would also open up a range of other models that could be fitted using `glmmTMB`.

Issue #1024, opened 2024-04-12
`simulate()` always returning the same values for `t_family`


davorvr:
Hi, I have a similar problem to #942 in that the `simulate()` function doesn't seem to work with t distribution family (`family=t_family`) models. I always get the same output:

```R
library(glmmTMB)
library(lme4)
#  Loading required package: Matrix
library(datawizard)

data(sleepstudy)
sleepstudy$y <- normalize(sleepstudy$Reaction)

m <- glmmTMB(y ~ Days + (1 | Subject), data = sleepstudy, family = t_family)

head(simulate(m, nsim = 5))
#       sim_1     sim_2     sim_3     sim_4     sim_5
# 1 0.2030275 0.2030275 0.2030275 0.2030275 0.2030275
# 2 0.2366451 0.2366451 0.2366451 0.2366451 0.2366451
# 3 0.2075882 0.2075882 0.2075882 0.2075882 0.2075882
# 4 0.4672708 0.4672708 0.4672708 0.4672708 0.4672708
# 5 0.5974521 0.5974521 0.5974521 0.5974521 0.5974521
# 6 0.8100759 0.8100759 0.8100759 0.8100759 0.8100759

head(simulate(m, nsim = 1))
#       sim_1
# 1 0.2030275
# 2 0.2366451
# 3 0.2075882
# 4 0.4672708
# 5 0.5974521
# 6 0.8100759
head(simulate(m, nsim = 1))
#       sim_1
# 1 0.2030275
# 2 0.2366451
# 3 0.2075882
# 4 0.4672708
# 5 0.5974521
# 6 0.8100759
head(simulate(m, nsim = 1))
#       sim_1
# 1 0.2030275
# 2 0.2366451
# 3 0.2075882
# 4 0.4672708
# 5 0.5974521
# 6 0.8100759
```

bbolker:
This was an oversight, thanks for pointing it out. Can you try `remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref = "t_simulate")`) and re-run?  That should fix the problem ...

davorvr:
That did it, works like a charm :) Thank you for a very quick fix!

PR #1025, opened 2024-04-12
implement simulation for t family (GH #1024)

bbolker:
This could be more thoroughly tested (I did a side test, not included in the package tests, along the lines of

```r
library(glmmTMB)
set.seed(101)
dd <- data.frame(x = 5 + 3*rt(1e4, df = 4))
m <- glmmTMB(x ~ 1, data = dd, family = t_family)
 c(fixef(m)$cond, sigma(m), exp(m$obj$env$parList()$psi))
## (Intercept)                         
##   4.942622    2.994766    3.810653 
s <- simulate(m)[[1]]
c(mean(s), sd(s))
## [1] 4.968806 4.240221
MASS::fitdistr(s, "t")
##      m            s            df    
##  4.96364830   3.00063550   3.93074002 
## (0.03556123) (0.03786045) (0.16207183)
```


bbolker:
For what it's worth, the GH actions checks are failing because of [this](https://github.com/leeper/prediction/issues/55) ... sigh ... (we could add `dotwhisker` and `prediction` as remote dependencies/install from github, but perhaps better to wait and see if the mess resolves itself ...)

bbolker:
I have continued to fight with GH actions/YAML incantations to find a way to install the missing packages (prediction/margins/dotwhisker) directly from GitHub (see the last few commits on the master branch), but so far without success. Still waiting for the `prediction` maintainer to act ...

PR #1028, opened 2024-04-17
Nbinom12

bbolker:
This adds a generalized form of the nbinom where the variance is a more general quadratic function of the mean (`V = mu*(1+phi + mu/psi)`). This is used e.g. in the `DESeq2` package and was proposed independently by Lindén & Mäntyniemi. 2011.  https://doi.org/10.1890/10-1831.1. The usual question of parameterization arises: in particular, should the quadratic term be parameterized as `mu/psi` (to match the usual nbinom2 parameterization) or `mu*psi` (which is a little more coherent)?


mebrooks:
I hadn't seen Lindén & Mäntyniemi's paper before, but it's very nice. I see that in the description, you showed how to transform this parameterization to theirs, so I think it's fine. I see that this parameterization is a blend of our 2 existing ones which is appealing.  Is it also like DESeq2? I didn't see their parameterization in a quick skim of some documents.

bbolker:
Yes, DESeq2 uses `Var = mu*(1+alpha*mu)`, `alpha = a + b/mu`.  (It also uses shrinkage; when we complete the "allow random effects in dispersion model" stuff I think we'll be one step closer to being able to replicate DESeq2 within glmmTMB ...

PR #1029, opened 2024-04-22
Add skewnormal likelihood

psmorris15:
Add skewnormal likelihood


bbolker:
This looks good, at a first glance.

It could be OK as is, but would be improved by:

- supporting simulation (via the SIMULATE macro), or as fallback (if this is challenging), making sure to throw an error if simulation is attempted
- some basic tests

bbolker:
Please go ahead and update to throw an error if simulation is attempted (for now).

bbolker:
This is failing on an undefined export ("sumTerms") because of [this line](https://github.com/glmmTMB/glmmTMB/pull/1029/commits/89a215917f0103d2fb0b05bd39c4aa2932201c7a#diff-6023410109d24b1448aac5b4e8246b78846845da397bc207f75528badee8e7b7L89); is that a mistake, or ?

psmorris15:
Thank you for adding the line to throw an error. Apologies for the delay in implementing the simulation. I am hoping to get to it this week. 

I am not sure where the "sumTerms" export came from. I did not define or call a function with that name. I followed the instructions in the vignette to update the namespace file. Truly not sure what happened there.

bbolker:
I'm not sure what was up with those additional exports either - things got moved around in the master branch (some functions were split off into a separate package, `reformulas`) and I think Git couldn't figure out that we should get rid of those lines in `NAMESPACE`. Looks OK now ...

PR #1032, opened 2024-05-01
Increase minimal R version

fdchevalier:
Dear glmmTMB team,

First thank you for developing and maintaining this package.

I would like to submit a change regarding the current minimal version of R required for installing the package. It is `>=3.6.0` ([source](https://github.com/glmmTMB/glmmTMB/blob/99dffd69bb371307505c7eb978c1ffc499c7d507/glmmTMB/DESCRIPTION#L45)). This leads to an error after compilation using R 3.6.0:
```R
"/usr/lib/R/bin/Rscript" -e "cat(as.character(packageVersion('TMB')), '\n', sep="", file='../inst/TMB-version')"
"/usr/lib/R/bin/Rscript" -e "cat(as.character(packageVersion('Matrix')), '\n', sep="", file='../inst/Matrix-version')"
installing to /home/XX/.R-libs/3.6.3/00LOCK-glmmTMB/00new/glmmTMB/libs
** R
Error in parse(outFile) :
  /tmp/RtmpH6yjZS/R.INSTALL67b7702b1f28/glmmTMB/R/priors.R:95:49: unexpected '>'
94:                                 if (length(cc) == 0) return(integer(0))
95:                                 cc <- c(1, cc) |>
                                                    ^
ERROR: unable to collate and parse R files for package ‘glmmTMB’
* removing ‘/home/XX/.R-libs/3.6.3/glmmTMB’
```

Pipe operators (`|>`) are present in several files of the package (like [here](https://github.com/glmmTMB/glmmTMB/blob/99dffd69bb371307505c7eb978c1ffc499c7d507/glmmTMB/vignettes/priors.rmd#L100)). Native pipe operator has been implemented in R only in version 4.1.0. So this should be the minimal version required for installation.

I tested the proposed change with R 4.1.3 and it worked as expected. Installation with R version lower than 4.1.0 will generate an expected error message regarding R version.

### DIFF
--- a/glmmTMB/DESCRIPTION
+++ b/glmmTMB/DESCRIPTION
@@ -42,7 +42,7 @@ Description: Fit linear and generalized linear mixed models with various
     out using the Laplace approximation. Gradients are calculated using automatic
     differentiation.
 License: AGPL-3
-Depends: R (>= 3.6.0)
+Depends: R (>= 4.1.0)
 Imports:
     methods,
     TMB (>= 1.9.0),


bbolker:
This is a very good point, but ideally I would prefer to fix the problem by restoring backward compatibility rather than increasing the required version of R. If it's only this stray use of the native pipe that's the problem then IMO we should fix it by getting rid of the pipe ...

fdchevalier:
Hi Ben,

I see your point. I have a R 3.6.0 installed, so I can test any new version without pipes.

A note on dependencies as you are mentioning it: when testing the installation of the package on R 4.1.3, I had to install older versions of MASS and Matrix packages. The current versions of these package on CRAN required R >= 4.4.0.

Anyway, I let you decide what is the best course of action.

Let me know if you have any version you would like me to test.

Fred

bbolker:
Can you test the version at https://github.com/glmmTMB/glmmTMB/tree/no_pipes ? (`remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref = "no_pipes")` should work, I think ...)

The issue with the latest versions of MASS and Matrix requiring R >= 4.4.0 is not technically our problem. In general any user who has R installed should have these packages (they are in the "Recommended" category), and shouldn't need to upgrade them in order to install `glmmTMB`.

fdchevalier:
I ran the test version on R 3.6.3 and everything went smoothly. Below the output for the record:
```R
R> remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref = "no_pipes")
Downloading GitHub repo glmmTMB/glmmTMB@no_pipes
Skipping 4 packages not available: RcppEigen, TMB, numDeriv, lme4
✔  checking for file ‘/tmp/RtmpVGKv9h/remotes74c1721e1037/glmmTMB-glmmTMB-2d7bfcd/glmmTMB/DESCRIPTION’ (597ms)
─  preparing ‘glmmTMB’: (986ms)
✔  checking DESCRIPTION meta-information ...
─  cleaning src
─  installing the package to process help pages
─  saving partial Rd database (5m 53.5s)
─  cleaning src
─  checking for LF line-endings in source and make files and shell scripts
─  checking for empty or unneeded directories
   Removed empty directory ‘glmmTMB/inst/misc’
─  looking to see if a ‘data/datalist’ file should be added
─  building ‘glmmTMB_1.1.9.tar.gz’

Installing package into ‘/home/XX/.R-libs/3.6.3’
(as ‘lib’ is unspecified)

[COMPILATION]

installing to /home/XX/.R-libs/3.6.3/00LOCK-glmmTMB/00new/glmmTMB/libs
** R
** data
*** moving datasets to lazyload DB
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
** building package indices
** installing vignettes
** testing if installed package can be loaded from temporary location
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
** testing if installed package keeps a record of temporary installation path
* DONE (glmmTMB)
```

Regarding dependencies, I understand your point.

Thank you for the quick turn around. Feel free to close the request if no more tests are needed.

bbolker:
resolved in cb6424a7a

Issue #713, opened 2021-06-20
setting/resetting number of cores for prediction (and other model calls)


bbolker:
At present we are being careful to [set the number of openmp threads appropriately when fitting a model](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/glmmTMB.R#L1245-L1257). I have now added code to the `rr_fixes` branch that (1) includes a `$parallel` element in the `$modelInfo` component of a fitted model (so we can tell how many threads it was originally fitted with) and (2) added code in `predict` to reset the number of threads accordingly.  A couple of questions and concerns:

- if users access the `$obj` component directly to do stuff with the model the resetting won't happen; we might want to add a note to the documentation telling users to call `TMB::openmp(<something>)` at the start of a session if they want to control the OpenMP threading ...
- it would be best if `glmmTMB` did not change whatever state OpenMP has before it's used. In particular, does calling `TMB::openmp(n = TMB::openmp(NULL))` leave the system state unchanged, by definition?

bbolker:
More of this is now handled correctly (?)  in the `openmp_fixes` branch ... although the bullet points are still relevant. (@kaskr, opinions/insight on those comments?)

markus-schaffer:
Hi, I wanted to ask if there are any new regarding the number of threads used by `predict()` ?
I run into the issue that `predict`() seems to use all available cores and overwrites `TMB::openmp(<something>)` if set beforehand. 
I have `glmmTMB` version 1.1.7.

bbolker:
With the current development version, you can set an OpenMP debugging flag (i.e. R code will report when it sets/resets number of threads) interactively. Can you try `remotes::install_github("glmmTMB/glmmTMB/glmmTMB")` and see if you get similar results with the testing code below? And can you run your code with the debugging flag on and report what you see?

```r
library(glmmTMB)
options(glmmTMB_openmp_debug=TRUE)
data("sleepstudy", package = "lme4")
m1 <- glmmTMB(Reaction ~ Days + (Days|Subject), sleepstudy)
## setting OpenMP threads to  1 
## setting OpenMP threads to  1
p <- predict(m1)
## resetting TMB threads to  1 
## setting OpenMP threads to  1 
## TMB threads currently set to  1 
## setting OpenMP threads to  1
m2 <- update(m1, control=glmmTMBControl(parallel = 4))
## setting OpenMP threads to  4 
## setting OpenMP threads to  1
p2 <- predict(m2)
[...]

PR #1054, opened 2024-06-22
modify parallel to allow autopar

bbolker:
This allows users to take advantage of the new autopar stuff in TMB (in particular, to allow parallelization with rr models).

* might need to update TMB dependence to a newer version (although this *might* still work even with older TMB versions -- haven't tested)
* this *should* be harmless to include in a new release, but I might wait until after the next CRAN-mandated release just to give ourselves a bit more time
* at the moment `autopar = FALSE` is the default, which I think makes sense until we have more info. For example, this test shows that for a simple example non-autopar is slightly faster than autopar (with an even bigger advantage for larger N ...)
* may need to worry about the (unlikely) event that glmmTMB has been compiled/installed using CppAD rather than (default) TMBAd?
* should update the parallel vignette accordingly (this could also close out https://github.com/glmmTMB/glmmTMB/issues/713)

Separately, can confirm that this works to speed up `rr()` fits (which are only possible with autopar).

```
$ Rscript --vanilla inst/misc/test_parallel.R TRUE
Loading required package: glmmTMB
setting OpenMP threads to 1L, autopar = NULL
setting OpenMP threads to 1L, autopar = FALSE
setting OpenMP threads to 4L, autopar = FALSE
setting OpenMP threads to 1L, autopar = FALSE
setting OpenMP threads to 4L, autopar = TRUE
setting OpenMP threads to 1L, autopar = FALSE
elapsed time for N=1e+04: serial=3.6, parallel_noauto=1.3, parallel_auto=1.8 (4 threads)
ratios=2.8 (noauto), 2.0 (auto), 0.7 (noauto vs auto)
platform: x86_64-pc-linux-gnu 
```

### DIFF
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -3,11 +3,15 @@ openmp_debug <- function() {
     getOption("glmmTMB_openmp_debug", FALSE)
 }
 
-## glmmTMB openmp controller copied from TMB (Windows needs it).
-openmp <- function (n = NULL) {
-    if (openmp_debug() && !is.null(n)) {
-        cat("setting OpenMP threads to ", n, "\n")
-    }
+## glmmTMB openmp controller copied from TMB (Windows needs it),
+## and it also lets us have more control of debug tracing etc.
+openmp <- function (n = NULL, autopar = NULL) {
+    report <-  (openmp_debug() && (!is.null(n) || !is.null(autopar)))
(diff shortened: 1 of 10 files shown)


bbolker:
This may be broken if we try to specify just `list(autopar = TRUE)` (somewhere we are checking `control$parallel$n` and not allowing for the possibility of a NULL value ... need to add a bit of logic and a test for that case.

Issue #1012, opened 2024-04-02
Obtaining site factors from `rr`


hrlai:
Thanks for implementing the reduced rank vcov structure. 

Following the [vignette](https://cran.r-project.org/web/packages/glmmTMB/vignettes/covstruct.html#reduced-rank) I was able to obtain the species loadings, but would it also be possible to obtain the site factors? It'd be great for making biplots.

If `fit.rr$fit$parfull` contains all the coefficients, then the site factors don't seem to be tracked? 

Thanks in advance!

mebrooks:
@mmcgillycuddy do you know the answer?

bbolker:
Hmm, I don't know offhand, but presumably we could look to see how the `gllvm` package does this (and in particular what's done in the [getLV() function](https://github.com/JenniNiku/gllvm/blob/0818869319441a24f54b180e9e34f981506a7201/R/getLV.gllvm.R) ?

mmcgillycuddy:
Hi, there are a few ways to do it, but I have this function to extract the information needed for biplots. It could probably be tidied up. @bbolker yes, happy to do something like this. 

```
extract_rr <- function(object){
  listname <- "cond"
  cnms <- object$modelInfo$reTrms[[listname]]$cnms   ## list of (named) terms and X columns
  flist <- object$modelInfo$reTrms[[listname]]$flist ## list of grouping variables
  flist_asgn <- attr(flist, "assign")
  levs <- lapply(seq_along(flist_asgn), function(i) levels(flist[[flist_asgn[i]]]))
  
  reStruc <- object$modelInfo$reStruc[[paste0(listname, "ReStruc")]] ## random-effects structure
  nc <- vapply(reStruc, function(x) x$blockSize, numeric(1)) ## number of RE params per block
  pl <- object$obj$env$parList(object$fit$par, object$fit$parfull)
  
  #function to split b by the random effect terms
  split.bseq <- function(object){
    listname <- "cond"
    reStruc <- object$modelInfo$reStruc[[paste0(listname, "ReStruc")]] ## random-effects structure
    nc <- vapply(reStruc, function(x) x$blockSize, numeric(1)) ## number of RE params per block
    nb <- vapply(reStruc, function(x) x$blockReps, numeric(1)) ## number of blocks per RE (may != nlevs in some cases)
    ### splitting the b's into their respective random effects
    nbseq <- rep.int(seq_along(nb), nb * nc)       ## splitting vector
    return(nbseq)
  }
[...]

PR #1062, opened 2024-07-09
add prediction of latent variables.

bbolker:
added a possible prediction type `predict(., type = "latent")`. This is potentially redundant with `ranef()`. I've been encountering some problems trying to extract latent variables reliably for reduced-rank models; not quite sure why this is happening (`rr()` models are a bit special in that they are the only case where the latent variables are modified *in place* ... that said, I would have expected to see *either* the "spherical" (prior/unconditional distribution is iid N(0,1)) *or* the "nonspherical" (spherical values multiplied by the factor loading matrix) ... possibly ranef() is sticking the spherical values (which are of lower dimension than the nonspherical ones) into a random-effects matrix of the wrong dimension?

In any case, sorry that's a bit stream-of-consciousness, this is something of a placeholder.  We need to fix `ranef()` so that it does something sensible for `rr()` models in any case; if it can be done with less messing around (adding REPORT/ADREPORT for `b`), then that might be better.


bbolker:
`ranef()` is now fixed (I think), and I've added some stuff to the covstruct vignette (at least partially addresses #1012). This could be ready to merge now, if nothing is obviously broken ...

mebrooks:
There's a problem with the `covstruct` vignette, but I can't see why. When I try to run it, I get this

```
Quitting from lines 697-700 [get-fl] (covstruct.rmd)
                                                                                                                      
Error in `source()`:
! empty file/url name
Backtrace:
 1. base::source(system.file("misc", "extract_rr.R", package = "glmmTMB"))
Execution halted
```
To me, this looks the same as all the other calls of `source(system.file(`.

bbolker:
I think I mistakenly had a pattern matching "misc" in `.Rbuildignore`.  Removed it, we'll see if that fixes the GH actions.

bbolker:
@mebrooks, please take another look (I mistakenly cherry-picked something to the master branch that depended on the `spider_long` data  set introduced in this PR ... rather than cherry-pick that as well it would be easier to merge this branch ... thanks ...)

Issue #391, opened 2018-09-19
Leverage/Cook's distance for glmmTMB outcomes


MelanieDou:
Good afternoon,
I have been using the glmmTMB package to fit a beta mixed-effect regression model to a variable between 0 and 1. My final model hopefully stands now and I am at the stage of checking its validity. One thing I would like to check is the influence of the various observations using ideally leverages and/or Cook's distance. I did not find any information on how to do this for glmmTMB. Did I miss it? If not, do you think it would be possible to have a function like the influence() function from the influence.ME package that can be applied to glmmTMB models?
Thank you already for your help!

bbolker:
I don't think there's anything available at present. To do this the easy (i.e. computationally efficient way) we'd need to to extract the diagonal of the hat matrix from the TMB fit: at present it [doesn't look like this is easy to do](https://github.com/kaskr/adcomp/issues/39). It might be possible to hack the influence.ME package to work with glmmTMB objects -- I haven't looked too closely ...

bbolker:
Bump. Any suggestions for strategies that will allow us to get hat values for glmmTMB fits reasonably straightforwardly/efficiently ... ?

alexhanrahan1:
Hello,

I had a question somewhat similar to this regarding fit of a beta mixed-effect regression model with varying dispersion and thought I should just add on to this post (sorry if it should be separate post). Has any progress been made toward calculating a hat matrix as you've said "without ridiculous effort"? Right now it seems only the response residuals are available:

"> res = residuals(bglmm, type = 'pearson')
Error in residuals.glmmTMB(bglmm, type = "pearson") : 
  pearson residuals are not implemented for models with zero-inflation or variable dispersion"

which aren't terribly helpful on their own. Is there anyway to calculate the following equation 2 from"Diagnostic tools in beta regression with varying dispersion; Ferrari et al. 2011)?

![eqn 2 Ferrari et al 2011](https://user-images.githubusercontent.com/69479880/89818961-44e46580-db10-11ea-80dc-5f9b6ae6222a.png)

Thank you in advance for your time!
Alex

bbolker:
[...]

PR #1065, opened 2024-07-19
Test implementation of glmmTMB leverage function

kaskr:
Here's a starting point if someone wants to look into #391.

TODO:

- Test that it's actually correct (compare with `numDeriv::jacobian`)
- Test that it's efficient on large data sets.
- What about discrete observations?

### DIFF
--- a/misc/leverage.R
+++ b/misc/leverage.R
@@ -1,0 +1,209 @@
+## This file contains:
+##
+## 1. A simple standalone RTMB example that calulates the entire hat
+##    matrix.
+##
+## 2. A general leverage function in glmmTMB based on RTMB with an
+##    option to only calculate the diagonal in an efficent manner.
+##
+## NOTE: Assumes glmmTMB to be compiled in a way compatible with RTMB !
+##   *  -DTMBAD_INDEX_TYPE=uint64_t
+##   *  NO OPENMP !
+
+######################################################################
+## A simple standalone RTMB example
+######################################################################
+
+library(RTMB)
+
+parameters <- list(
+    mua=0,          ## Mean slope
+    sda=1,          ## Std of slopes
+    mub=0,          ## Mean intercept
+    sdb=1,          ## Std of intercepts
+    sdeps=1,        ## Residual Std
+    a=rep(0, 50),   ## Random slope by chick
+    b=rep(0, 50)    ## Random intercept by chick
+)
+f <- function(parms) {
+    getAll(ChickWeight, parms, warn=FALSE)
+    ## Optional (enables extra RTMB features)
+    weight <- OBS(weight)
+    ## Initialize joint negative log likelihood
+    nll <- 0
+    ## Random slopes
+    nll <- nll - sum(dnorm(a, mean=mua, sd=sda, log=TRUE))
+    ## Random intercepts
+    nll <- nll - sum(dnorm(b, mean=mub, sd=sdb, log=TRUE))
+    ## Data
+    predWeight <- a[Chick] * Time + b[Chick]
+    nll <- nll - sum(dnorm(weight, predWeight, sd=sdeps, log=TRUE))
+    ## Get predicted weight uncertainties
+    ADREPORT(predWeight)
+    ## Return
+    nll
+}
+obj <- MakeADFun(f, parameters, random=c("a", "b"))
+opt <- nlminb(obj$par, obj$fn, obj$gr)
+## Leverage calc
+TMB::config(tmbad.atomic_sparse_log_determinant=0)
+p <- obj$env$parList()
+p$weight <- ChickWeight$weight
+F <- GetTape(MakeADFun(f, p)) ## Objective: (theta, u, obs) -> nll
+## Indices are as follows:
+i_theta <- 1:5
+i_u <- 6:105
+i_obs <- 106:683
+## Apply LA to integrate random effects out. Then opimize wrt theta
+That <- F$laplace(i_u)$newton(i_theta) ## That: (obs) -> theta_hat
+## Likewise, find Uhat: (theta, obs) -> u_hat
+Uhat <- F$newton(i_u)
+## Yhat tape
+Yhat <- GetTape(MakeADFun(f, p, ADreport=TRUE)) ## (theta, u, obs) -> muhat
+## Glue the stuff: (obs) -> muhat(theta(obs), u(obs), obs)
+Hat <- MakeTape(function(y) {
+    that <- That(y)
+    uhat <- Uhat(c(that, y))
+    Yhat(c(that, uhat, y))
+}, p$weight)
+## Entire hat matrix
+zx <- Hat$jacobian(Hat$par())
+## TODO: Compare with numDeriv::jacobian
+
+######################################################################
+## glmmTMB implementation
+######################################################################
+library(Matrix)
+library(glmmTMB)
+library(RTMB)
+
+## @param diag Get diagonal only?
+leverage <- function(fm, diag=TRUE) {
+    obj <- fm$obj
+    ## We mess with these... (cleanup on exit!)
+    restore.on.exit <- c("ADreport",
+                         "parameters",
+                         "data")
+    oldvars <- sapply(restore.on.exit, get, envir=obj$env, simplify=FALSE)
+    restore.oldvars <- function(){
+        for(var in names(oldvars)) assign(var, oldvars[[var]], envir=obj$env)
+    }
+    on.exit({restore.oldvars(); obj$retape()})
+    ## #################################################################
+    ## Get derivatives of prediction
+    ##
+    ##    mu_hat( b_hat( theta_hat(yobs), yobs) , theta_hat(yobs) )
+    ##
+    ## wrt. yobs.
+    ## Note the three partial derivative 'paths' to consider.
+    ## We can get this derivative by
+    ##  1. yobs -> theta_hat(yobs)
+    ##  2. theta -> mu_hat( b_hat( theta, yobs) , theta ) [fixed yobs]
+    ##  3. yobs -> mu_hat( b_hat( theta, yobs) , theta ) [fixed theta]
+    ## #################################################################
+    ##parhat <- obj$env$last.par.best
+    parhat <- fm$fit$parfull
+    pl <- obj$env$parList(par=parhat)
+    yobs <- obj$env$data$yobs
+    obj$env$parameters <- pl
+    theta <- parhat[!obj$env$lrandom()]
+    b <- parhat[obj$env$lrandom()]
+    Hbb <- obj$env$spHess(parhat, random=TRUE) ## Needed later
+    ## #################################################################
+    ## 1. Get partial derivatives of theta_hat wrt to yobs
+    ## Note: length(yobs) much greater that length(theta)
+    ##       ==> Reverse mode AD is suitable !
+    ## #################################################################
+    ## Move 'yobs' from data -> parameters (preserve C++ template order!)
+    obj$env$parameters <- c(list(yobs = yobs), obj$env$parameters)
+    obj$env$data$yobs <- NULL
+    obj$retape()
+    ## New objective parameters: (yobs, b, theta)
+    nobs <- length(obj$env$parameters$yobs)
+    nb <- length(obj$env$random)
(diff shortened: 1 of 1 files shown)


bbolker:
@kaskr, I did some testing. Basic comparisons with `lme4::hatvalues()` work OK. At present the code only works if the model has random effects; it would be convenient if it could be extended to work for non-mixed models. (This may be simple, but I didn't take the time to figure it out ...)

mebrooks:
@kaskr should we merge this?

bbolker:
I don't think it's ready yet. At the very least I'd like to see it extended to allow models without RE.

bbolker:
Have integrated it as `hatvalues()`, adapted to models w/o RE. Rudimentary checks seem to work. Issues:

* only matches `hatvalues.merMod` on `weight ~ Time + (Time || Chick), data = ChickWeights)`  to a tolerance of 0.009. Is that OK?
* warning from `RTMB::GetTape` about "Permanently changing the global pointer of DLL 'glmmTMB'"
* lots of dependence on RTMB at present: can we reduce this?
* uses private function `TMB:::solveSubset` (need to export from `TMB` or copy ...)
* possibly because of RTMB integration, requires changes to compilation settings (no OpenMP, etc.), which (IIUC) commits us to using `autopar` universally ...

kaskr:
@bbolker There's no easy fix and I don't think we can make it before the deadline #1084:

- If we keep the RTMB version we have to drop parallel support in glmmTMB (at least for now). Obviously not an option.
- On the other hand, translating the code to not depend on RTMB is time consuming and painful.

Before investing time on the second option, I'd love to get a bit more feedback on whether the computational approach works as expected (1) For a broad range of families, possibly including discrete observations. (2) For large datasets.

bbolker:
@kaskr, good points.  I guess the question will be how to enable reasonably broad testing. In the absence of students hanging around looking for projects, or paid research assistant help, it might help to post some relatively easy to use instructions for installing from this PR (`remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref = github_pull("1065"))` and asking people to try it out/compare with other packages (`lme4` provides hat values for LMMs and (with a warning) for GLMMs; not sure what other platforms there are for comparison ...)

bbolker:
Hi @kaskr, me again ...

Good news: I've done some more testing, comparing brute-force leverage calcs against your RTMB code, and things do seem to be working, including for discrete (nbinom) observations with non-identity link functions.

Bad news: the code here takes a huge amount of memory for large models (currently working on reduced-rank models with hundreds of taxa). Is there a relatively straightforward way **either**

* to make the computation less memory-hungry (i.e., do you see any low-hanging fruit/places where memory is obviously used unnecessarily?)
* allow us to conserve memory by computing leverage only for a subset of the original observations ... ??

kaskr:
@bbolker Overall I take this as good news. The bad news isn't that bad for a first hack...
Do you mind stepping through the calculations e.g. with `debug(glmmTMB:::hatvalues.glmmTMB)` and see where exactly the memory goes crazy? I tried with the `ou` example from another thread and found there might be a problem around here (which seems easy to fix): https://github.com/glmmTMB/glmmTMB/blob/1774fa4234b545ffbcce310a91ebbefbb009910a/glmmTMB/R/hatvalues.R#L97
but there might also be an issue here:
https://github.com/glmmTMB/glmmTMB/blob/1774fa4234b545ffbcce310a91ebbefbb009910a/glmmTMB/R/hatvalues.R#L77
which goes back to this FIXME:
https://github.com/glmmTMB/glmmTMB/blob/1774fa4234b545ffbcce310a91ebbefbb009910a/glmmTMB/R/hatvalues.R#L73

bbolker:
The retaping is definitely not the problem.  The proximal problem is the  `ThetaHat <- F$laplace(r)$newton(p)` call. When I do a moderate-sized problem, I'm at 2.7Gb memory used before starting the test, it then jumps up to just under 16Gb when I run that line ... (then goes back down to 7.3Gb, some of which can be retrieved by garbage collection)

See https://github.com/glmmTMB/glmmTMB/blob/formula_env/misc/memusg_master.R; this is actually on the `formula_env` branch because I needed to fix something about formula evaluation, but I could merge/cherry-pick stuff to `leverage` if that were more convenient ...

One thing that occurs to me is that all of these examples use reduced-rank term components, which might be a worst-case scenario - I *think* that the sparsity pattern of the Hessian might be nasty because of the way these are implemented. (I had some brief conversations with @fishfollower about related topics this summer. I think I get the general idea [when latent variables are not shared between observations the Hessian is sparser/nicer]; is there somewhere that the relationship of the latent-variable definition to the Hessian pattern is discussed in more detail?)  If you want me to pursue this I could try some similarly sized examples without rr terms ...

kaskr:
> The retaping is definitely not the problem.

Just to be sure, did you verify that on all the 3 (!) retape statements in the code?

> The proximal problem is the ThetaHat <- F$laplace(r)$newton(p) call.

That makes a lot of sense given the 'TMB FIXME' comment and the fact that the Hessian lacks sparsity.

> If you want me to pursue this I could try some similarly sized examples without rr terms

That would be super useful to know, i.e. are these problems specific to models with  rr terms, before we start fixing TMB.

bbolker:
OK, I'm now fairly confident that the reduced rank construction is indeed the problem. Now trying to figure out how to formulate the reduced rank problem in an efficient way. We are currently defining the random effects as $u$ (i.e. *spherical* effects), with $\eta = X \beta + Z \Lambda u$, $u \sim \textrm{MVN}(0, I)$. To make the Hessian pattern nicer we need to define the model in terms of $b = \Lambda u$, with $\eta = X \beta + Z b$, $b \sim \textrm{MVN}(0, \Lambda \Lambda^\top)$. The tricky part is that $\Lambda \Lambda^\top$ is degenerate, so we need the negative log-likelihood for a [degenerate MVN](https://en.wikipedia.org/wiki/Multivariate_normal_distribution#Degenerate_case), which is $(1/2) \cdot b^\top (\Lambda \Lambda^\top)^+ b) + (1/2) \log \textrm{det}^*(\Lambda \Lambda^\top)$, where + indicates a pseudo-inverse and det* indicates a pseudo-determinant (product of non-zero eigenvalues).  If we parameterize this on the precision scale (i.e. make `theta` the elements of an 'inverse factor loading matrix' $\Lambda^+$ where $\Sigma^+ = \Lambda^), is there a way we can efficiently compute log(det*) directly from

kaskr:
@bbolker 

> OK, I'm now fairly confident that the reduced rank construction is indeed the problem.

How so? There's still the possibility that the problems are due to inefficiencies in the leverage calculations... Did you look at the sparsity pattern?

FWIW, in theory, the 'spherical' construction should give rise to a dense, but small, Hessian block corresponding to 'u'. In addition, each data point links to *all* u's so each observation adds `length(u)` non zeros to the Hessian. However, the resulting Hessian should still be fairly sparse?

By working out an alternative construction (degenerate MVN...), you'd be able to reduce the non-zero fill to one per observation, *however* at the cost of a much bigger dense 'u'-block. Whether that's beneficial, will depend on the Lambda dimension *and* on the number of observations.

bbolker:
OK, here's where I am (details [here](https://github.com/glmmTMB/glmmTMB/blob/rr_inv/misc/rr_ex.R); I don't know how much detail/what format is best for you [i.e. briefer vs longer, R code with comments vs nicely formatted Quarto/Rmarkdown, etc.]).  Here's the brief summary: I'm puzzled by the results of the inner optimization that's part of the Laplace approximation machinery. The two neg-log-lik functions below give identical results when run in R or via RTMB/`MakeADFun` *without* random effects/LA (leaving out the Jacobian correction, which I implemented but decided I didn't need). When I evaluate the RTMB object *with* random effects, I seem to get different results of the inner optimization (i.e. the optimized non-spherical LVs `b` are not equal to the factor loading matrix times the optimized spherical LVs `u`). This is obviously a digression from the original question of whether the leverage computations are working/how to make them more efficient/whether changing representations will help, but I'd like to solve it if I can, and I'm not sure what the next debugging steps are.

For what it's worth, in this example the joint precision matrix for the spherical case is 18% sparse (5188 non-zero elements), while for the non-spherical case is 74% sparse (39960 non-zero elements)

These are images of log10(abs(jointprecision[random elements])):

rr_spher_nonspher


```r
f_spher <- function(par) {
    getAll(par, dd)
    Lambda <- to_factormat(theta, d)
    b <- Lambda %*% matrix(u, nrow = d)
    mu <- drop(beta0 + Z %*% c(b))
    REPORT(mu)
    REPORT(b)
    nll <- -sum(dnorm(y, mu, exp(logsd), log = TRUE))
    nllpen <- -sum(dnorm(u, log = TRUE))
    nll + nllpen
}

f_nonspher <- function(par) {
    getAll(par, dd)
    ## construct factor matrix with or without constraints?
    if (!raw_phi) {
        Phi <- t(to_factormat(theta, d))   ## zero-triangle
    } else {
        Phi <- t(matrix(theta, ncol = d))  ## no
    }
    mu <- drop(beta0 + Z %*% b)
    b <- matrix(b, nrow = nlev)
    u <- drop(Phi %*% b)
    nll <- -sum(dnorm(y, mu, exp(logsd), log = TRUE))
    nllpen <- -sum(dnorm(u, log = TRUE))
    if (jac_corr) {
        logdetphi <- ngrp*logdet(tcrossprod(Phi))
        nllpen <- nllpen - logdetphi  ## sign ???
    }
    REPORT(mu)
    REPORT(b)
    nll + nllpen 
}
```

bbolker:
@kaskr, a small bump on this. 

* If you think the non-spherical random-effects formulation is a possibly worthwhile route, any suggests for further tests/debugging directions on the issues suggested above would be welcome.
* Alternatively, if the non-spherical approach doesn't seem as though it's likely to be useful, we're back to "what is causing/how do we avoid the memory blow-up in the leverage calculation"?  I could run some more thorough/careful tests; I tested a compound-symmetric matrix against an rr matrix with a similar number of latent variables, but I should possibly double-check that the examples were really comparable (and I could check the Hessian patterns at the same time &mdash; it would be good to confirm that the way I did it above is appropriate ...)

bbolker:
@kaskr, another bump (really sorry to nag but would love to sort this out!)

kaskr:
@bbolker Finally managed to go though [this](https://github.com/glmmTMB/glmmTMB/blob/rr_inv/misc/rr_ex.R). I may be misunderstanding something, but I think there's a conceptual problem.
The spherical approach specifies the distribution of b by the singular *covariance* matrix `Lambda0 %*% t(Lambda0)`.
By contrast, the 'non spherical' approach specifies the distribution of b (implicitly) by the singular *precision* matrix `t(Phi0) %*% Phi0`. These can't possibly be equivalent given that one has zero variance in unspecified directions, while the other has infinite variance in unspecified directions.

bbolker:
Thanks, @kaskr. I think I understand the general issue (although to be honest I don't see exactly where my approach breaks down &mdash; I don't see any step at which I'm *explicitly* trying to use the infinite-variance precision matrix ...)  But in any case I am ready to give up on the non-spherical approach.

Can you give me any suggestions on where to continue with diagnosing/debugging why the memory usage is blowing up in the leverage computation for the reduced-rank model ... ???

kaskr:
@bbolker

> I don't see any step at which I'm explicitly trying to use the infinite-variance precision matrix

The problem is when you transform `b` *before* assigning its distribution:

```r
u <- Phi %*% b
nllpen <- -sum(dnorm(u, log = TRUE))
```

This corresponds to the quadratic form

```
nllpen = -sum(dnorm(Phi %*% b, log = TRUE))
       = .5 * t(Phi %*% b) %*% (Phi %*% b)
       = .5 * t(b) %*% ( t(Phi) %*% Phi ) %*% b
```

I.e. multivariate normal with improper precision `t(Phi) %*% Phi`.


> suggestions on where to continue

You previously found the problem to be the `ThetaHat <- F$laplace(r)$newton(p)` call.
It would be useful to know if the memory blows up already in the simpler call `tmp <- F$laplace(r)` ?
If yes, it would also be useful to know if restoring the global flag `TMB::config(tmbad.atomic_sparse_log_determinant=1, DLL="RTMB")` makes the call `tmp <- F$laplace(r)` use less memory ?

Issue #861, opened 2022-09-19
Exclude some but not all random effects in predict()?


GatesDupont:
The gam R package `mgcv` has a convenient feature in `predict.gam()` with the argument `exclude = ` that can take a value such as `exclude = "s(x)"`. 

The is super convenient, especially for random effects, particularly as it allows SOME terms to be excluded but not ALL.

Is there an equivalent way of doing this in `glmmTMB`? 

The closest I can imagine is `re.form = ~0` in `predict()`, but that's only for random effects, and it excludes ALL random effects, without providing you with selectivity.

bbolker:
Not at present. `lme4`'s `re.form` argument is more capable, it allows the user to specify which `re.form` values to **include** (not exclude).  There is some discussion [here](https://github.com/glmmTMB/glmmTMB/pull/511) about extending `re.form` to handle intermediate cases. The hard (not impossible) part is parsing `re.form` to figure out which elements of the random-effects vector `b` need to be excluded ... the relevant code to modify is highlighted in [this commit](https://github.com/glmmTMB/glmmTMB/pull/511/commits/ad073a6ba2ec040fa99d470a7d6d567a88221beb).

GatesDupont:
Understood. Thank you!

PR #1067, opened 2024-07-23
Extend RE form

bbolker:
This PR addresses a lot of (too many?) different things, primarily the need for finer control of random effects when predicting/simulating (i.e., allowing choices of `re.form` that specify particular RE terms or sub-terms). It's definitely not finished; I'm opening this PR as a place for discussion. 

What this does/should do:

- allow specification of `re.form` by a formula, even one specifying a submodel, e.g. if the RE in a model is `(1 + f1 | g1) + (1 + f2 | g2)`, then `re.form = ~ (1 | g1)` should specify that *only* the latent variables corresponding to intercept variation among `g1` levels should be included in predicted values, (?? or in simulated values ??). I could have sworn I had started to implement a bunch of stuff here (parsing names of terms and matching them up with elements of `b`) but now I can't find it anywhere (!!??!!)  I see this referred to in [this issue as well](https://github.com/glmmTMB/glmmTMB/issues/861). Guess I have to try to remember what my general approach was for identifying the subset of `b` corresponding to a particular subset of variables ...
- map (i.e., fix) relevant sets of `b` values when simulating or predicting (or setting population-level effects to zero), especially when fixed effect parameters may have been modified (`simulate.glmmTMB` docs currently say that it's not possible to condition on random effects when simulating ...) &mdash; especially important to prevent re-evaluation of the inner optimization in Laplace approximation. The [map_b branch](https://github.com/glmmTMB/glmmTMB/tree/map_b) should also probably get rolled in ...


mebrooks:
There's 1 failed test and 15 warnings all comming from `test-simulate_new.R`. @bbolker, I'll leave it to you since I don't want to change what you intended to test.

bbolker:
Looking at it now.

bbolker:
I think I've fixed this, although it's slightly scary to be working on a release in a hurry. (However, I think that anything I screw up will only affect edge cases of `simulate_new()` that no-one else is using, so it will be my/my student's problem ...)  Passes tests locally, waiting for GH actions to finish.

mebrooks:
Do you want to resolve the conflicts or should I have a look? I may have caused them when I merged the `prop2_covstruct` branch a few minutes ago.

bbolker:
I'll take a look.

bbolker:
:tada:

PR #1068, opened 2024-07-24
propto covstruc

mmcgillycuddy:
Changes made to include `propto` structure. Let me know if you want me to change where I have done things, or how.

Some things I need to do:
- [x] Need to add a check for the order of the matrix
- [ ] I think if you try to use map and start it breaks
- [ ] Need to add documentation to vignette


bbolker:
@mmcgillycuddy, I've cleaned things up so the current version passes checks.  If you give an example of breakage with `map`/`start` we *might* be able to find a reasonably straightforward way to engineer around the problem ... even if we don't (and just say "this doesn't work right now" in the docs), it would be good to have a reproducible example of the breakage, for future development ...

mmcgillycuddy:
Hi Ben, thanks for cleaning it up - I ran away on holidays for a little bit and then work.

I am working on checking if the row/column names of a matrix matches the LHS terms. I saw that the names are stored in `cnms`, but for factors it adds the name of the variable to the start of the level. Do you know if it's possible to get names before this happens? 
Here is an example:
```
dat <- data.frame(y = rnorm(60), x = rnorm(60), species = rep(paste0("spp", 1:10), 6), g = 1)
library(glmmTMB)
ranslp <- glmmTMB(y ~ 1  + ( x | species), doFit = F, data = dat )
ranslp$condList$reTrms$cnms
```
This returns `[1] "(Intercept)" "x"`
```
sppint <- glmmTMB(y ~ 1  + ( 0 + species | g), doFit = F, data = dat )
sppint$condList$reTrms$cnms
```
This returns: 
```
[1] "speciesspp1"  "speciesspp10" "speciesspp2"  "speciesspp3"  "speciesspp4"  "speciesspp5"  "speciesspp6"  "speciesspp7" 
 [9] "speciesspp8"  "speciesspp9"
```

bbolker:
Sorry, I started to respond to this earlier but lost the thread somewhere.

I'm assuming that you want to get the variable name so that you can check against the levels in the original data?

I think we might need to go back as far as `mkBlist` in `reformulas::mkTerms`; the corresponding code is [here](https://github.com/bbolker/reformulas/blob/main/R/mkReTrms.R#L182-L193). This function returns the  factor corresponding to the grouping variable  (`ff`), the transposed Z matrix (`sm`), the number of levels (`nl`), and the column names associated with the varying terms (`cnms`); we could ask it to return the names of the varying variables as well 

Alternately, this is ugly but does more or less the same thing:

```r
all.vars(reformulate(names(sppint$condReStruc))[[2]][[2]])
```

Breaking this down with pipes:

```r
(names(sppint$condReStruc)   ## "stuff1 | stuff2"
    |> reformulate()         ## convert back to a formula (could also get this from output of `findbars` ...) [ ~ stuff1 | stuff2 ]
    |> getElement(2)        ## this is the second element, i.e. the argument of ~  [ stuff1 | stuff2 ]
    |> getElement(2)        ## this the first argument of `|` [stuff1]
    |> all.vars()                  ## extract symbols
```

result "species".

Does that help?

mmcgillycuddy:
Hi Ben, 

Ah ok, so I am doing this within `getXReTrms` so I can't use `condReStruc`? But maybe I can still use `formula` and `fr` - I haven't looked into this yet. 

I was thinking of getting the order of the levels, instead of the variable name itself. So instead of getting 
```
[1] "speciesspp1"  "speciesspp10" "speciesspp2"  "speciesspp3"  "speciesspp4"  "speciesspp5"  "speciesspp6"  "speciesspp7" 
 [9] "speciesspp8"  "speciesspp9" 
```
from `cnms`, I would like
```
[1] "spp1"  "spp10" "spp2"  "spp3"  "spp4"  "spp5"  "spp6"  "spp7"  "spp8"  "spp9" 
```
From memory gsub isn't a good way to do this? 

I could also be overcomplicating this!

mmcgillycuddy:
> peciessp

```
require("nlme")
data(lizards)
tree <- read.tree(text=lizards$hprA)
liz <- lizards$traits[tree$tip.label, ]
liz$spp <- factor(rownames(liz), levels=rownames(liz))
liz$dummy <- factor(0)
mat <- vcv(tree, corr=TRUE)# construct matrix
fit_phylo <- glmmTMB(matur.L ~ age.mat + propto(0 + spp | dummy, mat),
                     data = liz)

# Say you want to fix the proporitonal parameter using start and map
# The prop par is the last theta  
fit_err_start <- glmmTMB(matur.L ~ age.mat + propto(0 + spp | dummy, mat),
                         data = liz,
                         map = list(theta = NA), 
                         start = list(theta = 1))
# This works but doesn't use the thetas for the specified matrix `mat` 
fit_2 <- glmmTMB(matur.L ~ age.mat + propto(0 + spp | dummy, mat),
                    data = liz, 
                    map = list(theta = c(1:171, NA )),  
                    start = list(theta = c(rep(0, 171), 1 ) ) )
```

bbolker:
I think you can get what you need from `mkReTrms` (which is called within `getXReTrms`, via the `$flist` component, e.g.:

```r
dat <- data.frame(y = rnorm(60), x = rnorm(60),
                  species = rep(paste0("spp", 1:10), 6),
                  g = 1,
                  other = factor(sample(LETTERS[1:5], size = 60, replace = TRUE)))
library(reformulas)
form <- y ~ 1  + ( x | species) + (1|other) + (x|other)
fr <- model.frame(subbars(form), dat)
rr <- mkReTrms(findbars(form), fr)
str(rr$flist)
List of 2
 $ species: Factor w/ 10 levels "spp1","spp10",..: 1 3 4 5 6 7 8 9 10 2 ...
 $ other  : Factor w/ 5 levels "A","B","C","D",..: 2 2 4 4 3 2 1 4 3 2 ...
 - attr(*, "assign")= int [1:3] 1 2 2
```

mmcgillycuddy:
I want to match the rownames of the matrix to what is on the LHS, not the RHS. For example, 

```
dat <- data.frame(y = rnorm(60), x = rnorm(60),
                  species = rep(paste0("spp", 1:10), 6),
                  g = 1,
                  other = factor(sample(LETTERS[1:5], size = 60, replace = TRUE)))
library(reformulas)
form <- y ~ 1  + ( x | other) + (0 + species |g) 
fr <- model.frame(subbars(form), dat)
rr <- mkReTrms(findbars(form), fr)
str(rr$cnms)
List of 2
 $ other: chr [1:2] "(Intercept)" "x"
 $ g    : chr [1:10] "speciesspp1" "speciesspp10" "speciesspp2" "speciesspp3" ...
```
To match the levels of species is not possible right now because cnms has the extra species in `cnms` .

PR #1075, opened 2024-07-27
Save smooth info

bbolker:
This is preliminary/to soon to actually merge, but it is now possible to extract information on which indices of `beta` and `b` correspond to each smooth term in a model:
```r
devtools::load_all("glmmTMB")
data("Nile")
ndat <- data.frame(time = c(time(Nile)), val = c(Nile))

sm1 <- glmmTMB(val ~ s(time, bs = "tp"), data = ndat,
                REML = TRUE, start = list(theta = 5))

lapply(sm1$modelInfo$reTrms,
       \(y) lapply(y$smooth_info, \(x) x$re$beta_ind))

lapply(sm1$modelInfo$reTrms,
       \(y) lapply(y$smooth_info, \(x) x$re$b_ind))
```

This should probably be built into a `getME()` option ... (and possibly we should simultaneously set up the ability to extract `b_ind` for *any* RE term, or any component of a RE term, which will help with the `extend_re_form` branch ...)

### DIFF
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -587,6 +587,7 @@ getXReTrms <- function(formula, mf, fr, ranOK=TRUE, type="",
             ##  far up do we have to go with eval.parent? Or do we
             ##  use the environment of the formula?
             if (!is.null(old_smooths)) {
+                ## we are predicting, want to use old smooths rather than constructing new ones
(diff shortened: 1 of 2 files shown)


bbolker:
For what it's worth it may be worth comparing what other packages that incorporate smooths do with them (`sdmTMB`, `tramME` ...

Issue #874, opened 2022-10-25
Load old fit to newer version/tryCatch to point to up2date?


bri2020:
Dear all, 

is there an official function or strategy on how to load old fits (saved as RDdata/or Rds for example) into a newer R (and thus glmmTMB) version. I've seen this discussion, but as a newbie and or not fully privy to this vocabulary, it's hard for me to figure out your final choices you had. [here https://github.com/glmmTMB/glmmTMB/issues/651](https://github.com/glmmTMB/glmmTMB/issues/651)

1. should I update my R to the same version everywhere and run my model again?
2. can i load an old fit somehow?
- with update function
- or do you have something integrated that I haven't found yet?

My problem occurs when I want to plot the data 

```
plot_model(m11, type = "pred", terms = c("sex", "delta_age"), mdrt.values="quart",colors = "bw")
```
```
Error in optimHess(oldPar, obj$fn, obj$gr) : 
  gradient in optim evaluated to length 1 not 8
> traceback()
13: optimHess(oldPar, obj$fn, obj$gr)
12: eval(substitute(expr), data, enclos = parent.frame())
11: eval(substitute(expr), data, enclos = parent.frame())
10: with.default(object, optimHess(oldPar, obj$fn, obj$gr))
9: with(object, optimHess(oldPar, obj$fn, obj$gr))
8: predict.glmmTMB(model, newdata = data_grid, type = "link", se.fit = se, 
       re.form = ref, ...)
7: stats::predict(model, newdata = data_grid, type = "link", se.fit = se, 
       re.form = ref, ...)
6: get_predictions_glmmTMB(model, data_grid, ci.lvl, linv, type, 
       terms, value_adjustment, condition, ...)
5: select_prediction_method(model_class = model_class, model = model, 
       data_grid = data_grid, ci.lvl = ci.lvl, type = type, model_info = model_info, 
       ppd = ppd, terms = original_terms, value_adjustment = typical, 
       vcov.fun = vcov.fun, vcov.type = vcov.type, vcov.args = vcov.args, 
       condition = condition, interval = interval, ...)
4: ggpredict_helper(model = model, terms = terms, ci.lvl = ci.lvl, 
       type = type, typical = typical, ppd = ppd, condition = condition, 
       back.transform = back.transform, vcov.fun = vcov.fun, vcov.type = vcov.type, 
       vcov.args = vcov.args, interval = interval, ...)
[...]

PR #1077, opened 2024-07-31
Up2date mapped disp

bbolker:
This addresses #874 in two ways:

* better handling of updating models with mapped parameters
* convert `Xd` and `XdS` data components to `Xdisp` and `XdispS` after integration of random effects for dispersion/expansion of "d" to "disp" in naming dispersion coefficients throughout

It works on the (large) examples provided by the OP of the issue.

It's a bit hard to test more generally because it involves using stored objects.  This does

```r
library(glmmTMB)
m1 <- glmmTMB(count ~ mined + (1|site),
   family=poisson, data=Salamanders)
m1_map <- update(m1, map=list(theta=factor(NA)),
    start=list(theta=log(10)))
## need to save and read back in order to exercise reconstruction/retaping part of `up2date()`
fn <- tempfile()
saveRDS(m1_map, fn)
m1_mapr <- readRDS(fn)
m1_mapu <- up2date(m1_mapr)
testthat::expect_equal(vcov(m1_map), vcov(m1_mapu))
```

Before the fix, this triggers

> Error in TMB::MakeADFun(data, parameters, map = map, random = random,  :  A map factor length must equal parameter length


bbolker:
OP confirms that this works for them. Think it's ready to merge (might add something to NEWS ...)

Issue #1101, opened 2024-09-30
Changes to `residuals()` or `fitted()`?


strengejacke:
In the [performance-package](https://easystats.github.io/performance/), the `r2()` function used to return identical results for `glmmTMB()` and `pscl::zeroinfl()` for non-mixed zero-inflated models. Now, the R2 is lower for glmmTMB, which seems to break the performance package (https://cloud.r-project.org/web/checks/check_results_performance.html).

``` r
library(performance)
data(bioChemists, package = "pscl")
m <- glmmTMB::glmmTMB(
  art ~ fem + mar + kid5 + ment,
  ziformula = ~ kid5 + phd,
  family = poisson(),
  data = bioChemists
)
r2(m) ## *was* identical to pscl::zeroinfl
#> # R2 for Zero-Inflated and Hurdle Regression
#>        R2: 0.149
#>   adj. R2: 0.145

# validate against pscl::zeroinfl
m2 <- pscl::zeroinfl(
  art ~ fem + mar + kid5 + ment | kid5 + phd,
  data = bioChemists
)
r2(m2)
#> # R2 for Zero-Inflated and Hurdle Regression
#>        R2: 0.180
#>   adj. R2: 0.175
```

<sup>Created on 2024-09-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The underlying code in `r2()` is not very complex (https://github.com/easystats/performance/blob/main/R/r2_zeroinflated.R), and I wonder, whether anything in `fitted()` or `residuals()` has changed that lead to the different results?

strengejacke:
Calculation of Pearson's residuals seems to have changed:

## Output for glmmTMB 1.1.10

``` r
data(bioChemists, package = "pscl")
m <- glmmTMB::glmmTMB(
  art ~ fem + mar + kid5 + ment,
  ziformula = ~ kid5 + phd,
  family = poisson(),
  data = bioChemists
)

y <- insight::get_response(m, verbose = FALSE)
sum((stats::fitted(m) - mean(y))^2)
#> [1] 275.8094
sum(stats::residuals(m, type = "pearson")^2)
#> [1] 1569.995
```

<sup>Created on 2024-09-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

## Output for glmmTMB 1.1.9

``` r
data(bioChemists, package = "pscl")
m <- glmmTMB::glmmTMB(
  art ~ fem + mar + kid5 + ment,
[...]

PR #1102, opened 2024-09-30
Fix pearson resids

bbolker:
#1101


bbolker:
A lot of this is inertia. For a while the formatting of `.md`-formatted news was considerably uglier; now it's been fixed, although still slightly different  (compare [performance NEWS](https://cran.r-project.org/web/packages/performance/news/news.html) with [glmmTMB NEWS](https://cran.r-project.org/web/packages/glmmTMB/news.html)). If we wanted to convert I think that [xfun::news2md](https://search.r-project.org/CRAN/refmans/xfun/html/news2md.html) would get us most of the way there ...

Issue #1122, opened 2024-11-22
column names of model.matrix of glmmTMB isn't in right order


dongwenluo:
I have a data with three factor: `type` with levels of R and S; `strainN` with levels of 1, 2 and 3; `dose` with levels of N and C.

After fitting the following model 

```
> mod_tmb <- glmmTMB(l3 ~ (type/strainN)*dose + (1|rep/aid),
+                   family="poisson",
+                   data=cyc_1)
```
I got column names of model matrix  with wrong order (see the bold part)

```
> colnames(model.matrix(mod_tmb))
 [1] "(Intercept)"          "typeS"                "doseC"               
 [4] "typeR:strainN2"       "typeS:strainN2"       "typeR:strainN3"      
 [7] "typeS:strainN3"       "typeS:doseC"          **"typeR:doseC:strainN2"
[10] "typeS:doseC:strainN2" "typeR:doseC:strainN3" "typeS:doseC:strainN3"**
```

If I fit the same model using glmer

```
> mod_lmer <- glmer(l3 ~ (type/strainN)*dose + (1|rep/aid),
+                     family="poisson",
+                     data=cyc_1)
```

I got the right column names of the  model matrix

```
> colnames(model.matrix(mod_lmer))
 [1] "(Intercept)"          "typeS"                "doseC"               
 [4] "typeR:strainN2"       "typeS:strainN2"       "typeR:strainN3"      
 [7] "typeS:strainN3"       "typeS:doseC"          **"typeR:strainN2:doseC"
[10] "typeS:strainN2:doseC" "typeR:strainN3:doseC" "typeS:strainN3:doseC"**
```

Could you please fix the issue?

Best regards

Dongwen

bbolker:
Fixed in a branch, waiting for [pull request](https://github.com/glmmTMB/glmmTMB/pull/1123) to be merged. You can install from that branch if you want, `remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref = github_pull(1123))`

dongwenluo:
Many thanks, Ben.

Dongwen

PR #1123, opened 2024-11-23
Don't rearrange formula

bbolker:
The solution that I was using to remove offsets from a formula (from [here](https://stackoverflow.com/questions/40308944/removing-offset-terms-from-a-formula)) was sometimes rearranging the order of terms in a formula (#1122).  Replaced `reformulas::drop.special` with `reformulas::noSpecial(., specials = "offset")` ...


bbolker:
Don't be misled by the red X/apparently failed check - that was manually canceled because it was a trivial change. This is a backward-incompatible change because it will change the order of parameters in some cases ...

PR #1134, opened 2024-12-03
more careful evaluation of  family component (fix #1133)

### LINKED ISSUE #1133: Error in `glmmTMB.Effect()`
Reported here: https://github.com/strengejacke/sjPlot/issues/957

``` r
library(glmmTMB)

#Data frame
set.seed(12345)
x <- round(rnorm(800, mean=20, sd=10.5),0)
x[x<6] <- 6
x[x>40] <- 40
set.seed(12345)
y <- rnbinom(800, mu=10.6, size=1.4)
set.seed(12345)
y[floor(runif(n=140, min=1, max=800))] <- 0
site <- c(rep("A",25), rep("B",50), rep("C", 75), rep("D",100), rep("E",150), rep("F",200), rep("G",200))
example <- data.frame(cbind(x,y,site))
example$x <- as.integer(example$x)
example$y <- as.integer(example$y)

mod_test <- glmmTMB(y ~ x + (1 | site), data=example, ziformula=~., family=truncated_nbinom2)
effects::Effect("x", mod_test)
#> Warning in Effect.glmmTMB("x", mod_test): overriding variance function for
#> effects/dev.resids: computed variances may be incorrect
#> Error in log(mu): non-numeric argument to mathematical function
```

<sup>Created on 2024-12-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

bbolker:
I think this should be OK, but I am mildly nervous about squeezing it into the upcoming release because it does change user-visible behaviour ... on the other hand it would be a shame to leave it out ... (not sure why the issues only become apparent when calling `effect::Effect()`, but that's part of the mystery of delayed evaluation ...)

Maybe we can merge this and then run reverse-dependency checks again ... ??

### DIFF
--- a/glmmTMB/R/family.R
+++ b/glmmTMB/R/family.R
@@ -32,8 +32,10 @@ make_family <- function(x, link, needs_nonneg = FALSE, needs_int = FALSE) {
         x <- c(x,list(aic=function(...) NA_real_))
     }
     if (is.null(x$initialize)) {
-        x <- c(x,list(initialize=
-                          substitute(env = list(FAMILY=x$family),
+        x <- c(x, list(initialize=
+                           substitute(env = list(FAMILY=x$family,
+                                                 needs_nonneg = needs_nonneg,
+                                                 needs_int = needs_int),
             expr = expression({
             ## should handle log-links adequately
             mustart <- y+0.1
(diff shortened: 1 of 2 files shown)


mebrooks:
This plan sounds fine to me, but the error is quite advanced, so I don't have any input on that.

bbolker:
I think I'm willing to risk merging for the upcoming release, if you're willing to rev-dep checks one more time ...

PR #1135, opened 2024-12-04
Fix mcmc vignette

bbolker:


mebrooks:
Thanks for that. I can focus on this tomorrow morning. 

I like this example, but I was thinking that one strength of showing the old example is that we can show how bounding it makes the chains behave better (Kasper told me to try it for my own problem). Is it too much to show both?

bbolker:
I think showing both examples is a good idea.

I have mixed feelings about the bounding -- it's easy, but it corresponds (I think) to a uniform prior, which I generally think is bad practice -- if we had the energy it would be best to show both the bounding method and a better way to do it, which is to use 'real' priors (but ... we might discover bugs/holes in the prior implementation when we tried to do that ...)

kaskr:
@bbolker 'Bad practice' perhaps from a Baysian perspective, because bounding might severally affect posterior moments.
However, for a *normal* glmmTMB model *without priors*, I consider sampling with bounds the only meaningful (and useful) way to sample:

- Set infinite bounds on random effects
- Set whatever bounds you like on the parameters to ensure the sample exists.

Interpretation is then straight forward in a non-bayesian setting: parameter samples represent the *unmodified marginal likelihood* within the selected window. For visual inspection this is probably what most users want.

bbolker:
Thanks @kaskr. I agree that this is a reasonable approach. 

I do get uneasy when we start to get into the murky areas of sort-of-kind-of Bayesian approaches.  Not quite sure what you mean by marginal likelihood here: I usually interpret "marginal likelihood" as $\iint L(\beta | y, \mathbf b) L( \mathbf b | \theta) \, d{\mathbf b}$, where we're integrating over the latent variables. Now I guess if we're looking at $L(\beta_i)$ (i.e., looking at the density of the marginal distribution of MCMC samples of a particular parameter) then we're integrating over $\beta_{-i}$ as well ... but I'm not sure how we interpret this except in a Bayesian way ... ??

kaskr:
Yes, this is what I understand by marginal likelihood: integration over latent variables.
Sampled $\beta$ would have density proportional to

$\int L(\beta|y, b)L(b| \theta) db \times 1_{(\beta \in B)}$

where $B$ is the bound.

Sure, it's not necessarily easy to get from a sample representation to quantities of interest such as MLE, confidence intervals, profiles likelihood etc., but in theory it's possible.

Issue #1115, opened 2024-11-10
Censored Normal / Tobit regression


dke-chan:
I've done some work to hack in a censored normal (tobit) response into glmmTMB with a global interval (not per observation as mentioned in https://github.com/glmmTMB/glmmTMB/issues/690) for a work project recently (see https://github.com/dke-chan/glmmTMB/tree/cens_norm). I'm wondering if it is something that might be useful to develop into an end-user solution. If so, here is a list of things I have identified that will need some input before spending more time on it...

- Is `glmmTMB(..., family = censored_normal(a = 0, b = Inf), ...)` an ideal end-user interface?
- Currently, I have hacked `glmmTMB()` to specify the `start` and `map` arguments for the global interval of the censored normal response as they are implemented as `psi` parameters on the C++ end (lines 1388-1400 of https://github.com/dke-chan/glmmTMB/blob/cens_norm/glmmTMB/R/glmmTMB.R). Is this okay?
- [Wikipedia](https://en.wikipedia.org/wiki/Tobit_model) suggests that there is literature on reparameterizing the likelihood function to be better behaved than the one mentioned in the [censReg](https://cran.r-project.org/web/packages/censReg/vignettes/censReg.pdf) R package. Would this be worthwhile to investigate to reduce the number of NA / NaN evaluations?
  - Similarly, I "think" the choice of starting values matters more for a censored normal response due to the NA / NaN evaluations described above (based on what I experienced during the work project). I'm not sure where would be a good place to override the default starting values if that is indeed the case.
- See how to include the global interval of the fitted censored normal response in the `print()`/`summary()`.

bbolker:
>     * Is `glmmTMB(..., family = censored_normal(a = 0, b = Inf), ...)` an ideal end-user interface?

Seems reasonable (or `tobit(...)`, but `censored_normal()` is probably better (as Tobit is more specific, and also would be less obvious to users who don't know that terminology)

>     * Currently, I have hacked `glmmTMB()` to specify the `start` and `map` arguments for the global interval of the censored normal response as they are implemented as `psi` parameters on the C++ end (lines 1388-1400 of https://github.com/dke-chan/glmmTMB/blob/cens_norm/glmmTMB/R/glmmTMB.R). Is this okay?

[...]

PR #1141, opened 2024-12-18
Censored normal family

dke-chan:
A continuation of the issue (for a feature request) that I mentioned in https://github.com/glmmTMB/glmmTMB/issues/1115. Based on the last time a new family was implemented, I am missing `testthat` tests in my branch. It would be good to (a) see if there is anything I need to iron out, and (b) if my implementation is sensible (I have some examples you can access here: https://gist.github.com/dke-chan/c523122adca15f61d638c0e0bf406057).


moosterwegel:
Thanks for giving this a go! I remember from my ventures into multilevel tobit model implementations that the censReg implementation was not entirely uncontroversial (see https://stat.ethz.ch/pipermail/r-sig-mixed-models/2021q3/029715.html). Maybe an idea to test the solutions vs GLMMadaptive too? 

I'm unfortunately not familiar with the glmmTMB codebase so I can't comment on that, but if I understand correctly you now specify a censoring limit (left or right) for all observations similar to censReg? Would it be possible to specify censoring per observation (so per observation the value where it was censored and an indicator of the type of censoring)? Similar to how the brms, survival and GLMMadaptive packages specify censoring.

dke-chan:
> Maybe an idea to test the solutions vs GLMMadaptive too?

I can do that and update the gist I have linked above.

> I'm unfortunately not familiar with the glmmTMB codebase so I can't comment on that, but if I understand correctly you now specify a censoring limit (left or right) for all observations similar to censReg? 

Yes, that is the case. Currently, I pass the censoring limits through the `psi` slot in the TMB template. Now that I think about it... I could allow the limits to differ per observation, but I'm not too sure if having a dynamic-sized `psi` slot will be okay for the long-term (see Ben's comment in the issue I opened up https://github.com/glmmTMB/glmmTMB/issues/1115#issuecomment-2467008149).

PR #1153, opened 2025-02-04
Kenwardroger

bbolker:


bbolker:
While the basic pieces are here to provide Kenward-Roger and Satterthwaite denominator-degree-of-freedom approximation, there are a lot of things to consider about when these approximations are appropriate, and what we should do to (dis)allow them when they're not appropriate (not offer them at all? warn? message? automatically downgrade to something more supported?) Some of the relevant components are:

* *presence of random effects* (`hasRandom(model)`): not appropriate? (Use default `df.residuals()` for Gaussian models, otherwise asymptotic)
* *presence of non-trivial dispersion component*: ??
* *non-REML fits*: K-R not appropriate?
* *GLMMs*: ??? 
* *component*: I don't know if these approximations would be applicable to the zero-inflation or dispersion components (*suggestion*: disallow them completely, drop back to asymptotic df)

I've tried to implement some of this logic [here](https://github.com/glmmTMB/glmmTMB/blob/kenwardroger/glmmTMB/R/emmeans.R#L108-L133), but would welcome feedback ...

Issue #1155, opened 2025-02-07
Memory-management errors in parallel mode


Kodiologist:
Consider this example:

```R
library(glmmTMB)

n.workers = 8L

n = 1e5L
set.seed(10L)
d = transform(
    data.frame(x = seq(0, 1, len = n)),
    y = rbinom(n, size = 1, prob = x / 3))

i = 0
repeat
   {i = i + 1
    message(i)
    glmmTMB(
        data = d,
        formula = y ~ x,
        family = nbinom2,
        control = glmmTMBControl(parallel = n.workers))}
```

One run looks like this:

```
1
2
3
4
5
6
7
8
9
10
11
Error in MakeADFunObject(data, parameters, reportenv, ADreport = ADreport,  : 
  INTEGER() can only be applied to a 'integer', not a 'pairlist'
Calls: source ... eval -> eval -> MakeADFun -> retape -> MakeADFunObject
In addition: Warning messages:
1: In finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting'), help('diagnose')
2: In finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting'), help('diagnose')
3: In finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting'), help('diagnose')
4: In finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting'), help('diagnose')
5: In finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting'), help('diagnose')
6: In finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting'), help('diagnose')
7: In finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting'), help('diagnose')
8: In finalizeTMB(TMBStruc, obj, fit, h, data.tmb.old) :
  Model convergence problem; false convergence (8). See vignette('troubleshooting'), help('diagnose')
[...]

PR #1167, opened 2025-03-01
Improve par control

bbolker:
this does a better job sorting out the possible ways to specify threads/autopar in `glmmTMBControl`:  

* `parallel = list(n=2, autopar = TRUE)`, `parallel = list(2, autopar = TRUE)` both set 2 threads, autopar on.
* `parallel = 2`, `parallel = list(n=2)` both set 2 threads, autopar defaults to whatever it was set to previously

I may have added some unnecessary complexity by accident ...

partially addresses #1155 (not the fundamental issue, which is an upstream TMB issue).


mebrooks:
I still get the same error (above) on this branch.

mebrooks:
This could be hard to debug if it only fails on certain platforms ("All checks have passed").

bbolker:
I sporadically encounter this kind of problem with `devtools::test()`. What do you get with `make check-cran`? That's admittedly slower than `make test`, but it should be the gold standard (i.e. if that works and `make test` fails, I would ignore `make test` for now ...)

mebrooks:
You're right; `make check-cran` doesn't produce the DLL error. It's weird that I'm only getting it on this branch though.

mebrooks:
Oddly this branch gives me a different error when I run it on the cluster. I made sure that the master branch doesn't give the error.  While it's a trivial error, should we be worried that it's different from the master branch?
```
Failure ('test-downstream.R:71:7'): joint_test with non-estimable terms
c(suppressMessages(joint_tests(nested0))) not equal to c(suppressMessages(joint_tests(nested))).
Component “p.value”: Mean relative difference: 1.553134e-05
```

PR #1172, opened 2025-03-12
Aggregate3

mebrooks:
Of course we'll have to squash all these back-and-forth commits into one, but I think this is ready for review. Could the structures be optimized, or is this the optimum?

I confirmed that (for 1 test case) this branch gives the same predictions as the `aggregate` branch from a few years ago.


bbolker:
I'm still not entirely clear on what this does. There doesn't seem to be a NEWS entry explaining it; could there be?  What's the use case? Should there be something in a vignette?

mebrooks:
I agree that a NEWS item is needed now and preferrably also a vignette. @casperwberg could we use some of your simulation tests for a vignette?

I need to use it to first model bycatch per unit fishing effort and then make predictions of the total annual bycatch for this type of fishery across the whole area. The model has spatial and temporal random effects.

```
bc.model = glmmTMB(porpoises ~ vessel.length + mesh +  
		(1 | vessel.id) + (1 | year) + exp(pos + 0 | quarter:year) + 
		log(soak) + offset(log(netlength)),  
		data = datporpoises, family = nbinom2)

annual.bycatch =  predict(bc.model,
			 newdata=datfp0noping,
			 allow.new.levels=TRUE,
			 aggregate=factor(datfp0noping$year),
			 se.fit=TRUE,
			 do.bias.correct=TRUE,
			 type="response")
```
Then `predict` returns an object with nrows = `length(unique(datfp0noping$year))` and columns "Estimate", "Std. Error", "Est. (bias.correct)", "Std. (bias.correct)", so you can look at the annual values with and without the bias correction.


This paper, [Thorson & Kristensen (2016)](http://dx.doi.org/10.1016/j.fishres.2015.11.016) which was based on [Tierney et al. 1989](  https://www.jstor.org/stable/2289652),  says it's based on the Taylor expansion where if $B = f (A)$ then $E[B]∼=f(E[A])+f′(E[A]/2)Var[A]$. I think this is also the topic of the recent book by Thorson & Kristensen, but I haven't had a chance to read it.

`TMB::sdreport` says
> Using $\phi(\hat u,\theta)$ as estimator of
>  $\phi(u,\theta)$ may result in substantial bias. This may be
>  the case if either $\phi$ is non-linear or if the distribution
>  of $u$ given $x$ (data) is sufficiently non-symmetric.  A
>  generic correction is enabled with `bias.correct=TRUE`. It is
>  based on the identity
>  $E_{\theta}[\phi(u,\theta)|x] =$ 
> $\partial_\varepsilon(\log \int \exp(-f(u,\theta) + \varepsilon \phi(u,\theta))\:du)_{|\varepsilon=0}$
>  stating that the conditional expectation can be written as a
>  marginal likelihood gradient wrt. a nuisance parameter
>  $\varepsilon$.
>  The marginal likelihood is replaced by its Laplace approximation.
> 
>  If `bias.correct.control$sd=TRUE` the variance of the
>  estimator is calculated using
>  $V_{\theta}[\phi(u,\theta)|x] =$
>  $\partial_\varepsilon^2\left(\log \int \exp(-f(u,\theta) + \varepsilon \phi(u,\theta))\:du\right)_{|\varepsilon=0}$
>  A further correction is added to this variance to account for the
>  effect of replacing $\theta$ by the MLE $\hat\theta$
>  (unless `ignore.parm.uncertainty=TRUE`.
> 
>  Bias correction can be be performed in chunks in order to reduce
>  memory usage or in order to only bias correct a subset of
>  variables. First option is to pass a list of indices as
>  `bias.correct.control$split`. E.g. a list
>  `list(1:2,3:4)` calculates the first four ADREPORTed
>  variables in two chunks.
>  The internal function `obj$env$ADreportIndex()`
>  gives an overview of the possible indices of ADREPORTed variables.
> 
>  Second option is to pass the number of
>  chunks as `bias.correct.control$nsplit` in which case all
>  ADREPORTed variables are bias corrected in the specified number of
>  chunks.
>  Also note that `skip.delta.method` may be necessary when bias
>  correcting a large number of variables.

I'll do some timing tests also.

bbolker:
I've merged for now (I think if you add vignettes/NEWS to this branch you can make a further pull request ...)

kaskr:
@mebrooks : No, the bias correction method is **not** based on a Taylor expansion. Rather, as the `TMB::sdreport` documentation says, it is based on a Laplace approximation *gradient*. The real strength of the method is that it 'magically' corrects for skewness of the random effect distribution conditional on data (which would not be the case when using a Gaussian approximation).

PR #1173, opened 2025-03-17
test for phylo mat rownames/colnames in wrong order

bbolker:
Responding to [this r-sig-mixed post](https://stat.ethz.ch/pipermail/r-sig-mixed-models/2025q1/030775.html), added a specific test to report if row/col names of 'propto' mat were the same as the levels of the term *but in the wrong order*


bbolker:
I think we should make `checkPropNames` return a (possibly reordered) `propto` matrix: if `reorder=TRUE`, it could automatically reorder the scaled covariance matrix to match the order of the rows in Z ... @mmcgillycuddy , does that seem reasonable?

mmcgillycuddy:
Do you want to fit it with the reordered matrix, or just return the matrix?

bbolker:
I was going to suggest fitting with the reordered matrix. (The reordering should generally be harmless, and if users don't want it to happen they can make the order match in the first place ...)  On the other hand, we could merge this PR now, for the imminent release, and propagate the documentation enhancements before worrying about actual changes to the code ...

mmcgillycuddy:
Yes, sounds like merging the current changes is a good idea. I will modify the code to reorder the matrix

Issue #1188, opened 2025-04-10
storing (or not) large correlation matrices, corr printing


bbolker:
I'm working my way through some significant updates to correlation handling, and have some thoughts/questions/requests for comment.  (Code is in the [ar_memfix branch](https://github.com/glmmTMB/glmmTMB/tree/ar_memfix).)

* https://github.com/glmmTMB/glmmTMB/issues/995 highlights that we are sometimes creating giant correlation matrices, purely for reporting purposes, when we don't really have to. (There are some other issues relating to the creation of giant (dense) *contrast matrices*, but that's a problem that we will have to deal with in the `reformulas` package.)
* I started by simply chopping out the code that creates the correlation matrices from the `ar1` and `ou` covstructs, e.g. [here](https://github.com/glmmTMB/glmmTMB/blob/ar_memfix/glmmTMB/src/glmmTMB.cpp#L534-L548), only storing the `phi` value. 
* However, I think there should probably be a flag, to be set in `glmmTMBControl` (e.g. `return_full_corrmat`?); the default could be (denoted by `NULL` value?) to return the full corrmat *if* it would be (say) less than 20x20, or something reasonable. This behaviour should probably apply across all covstruct types.
* This is separate from how `VarCorr` objects are constructed on the R side after the model is fitted (`VarCorr`/`mkVC`), and how `VarCorr` objects are formatted/printed.   What should a `VarCorr` object actually contain when the covariance matrix is too big to be practical (when `return_full_cormat` is FALSE)?  In some cases it could be a sparse object, but in many cases (e.g. AR1/OU/spatial) it should probably be something like an empty (0x0) numeric matrix ... 
* I've been writing (S3) methods for printing individual components of VarCorr objects based on their cov structs. These will have to work whether or not the full matrix is available ...

bbolker:
This is largely resolved but perhaps the issue should stay open until we make sure that tests are relatively complete and as many correlation structures as possible have sensible printing methods.  There's a bit of minimal testing [here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/tests/testthat/test-VarCorr.R#L237-L244), but all this does is make sure that printing doesn't actually break for any of the model fits we have stored. (I don't think we are currently storing any models *without* the full correlation structure saved ...)

Issue #1140, opened 2024-12-13
Strange confusing name of the ar1 effect in the summary() output


telenskyt:
Why is the AR1 effect reported with strange name, which includes the first level of that factor? See this summary output: 

```
 Family: gaussian  ( identity )
Formula:          x ~ ar1(year + 0 | group)
Data: dat0

     AIC      BIC   logLik deviance df.resid 
    70.1     75.7    -31.1     62.1       26 

Random effects:

Conditional model:
 Groups   Name     Variance Std.Dev. Corr      
 group    year1993 0.94465  0.9719   0.77 (ar1)
 Residual          0.04344  0.2084             
Number of obs: 30, groups:  group, 1

Dispersion estimate for gaussian family (sigma^2): 0.0434 

Conditional model:
            Estimate Std. Error z value Pr(>|z|)
(Intercept)  -0.4224     0.4599  -0.918    0.358

```

The name `year1993` is strange and confusing. It suggests it has to do something with the first level of that factor (the first year of the time series), but logically it shouldn't. Could you please polish this part a bit and keep there just some proper name? Probably just `year`, or maybe better `ar1(year)` or something like that?

Here is the code to reproduce:

```
library(glmmTMB)

n <- 30                                             ## Number of time points
x <- MASS::mvrnorm(mu = rep(0,n),
             Sigma = .7 ^ as.matrix(dist(1:n)) )    ## Simulate the process using the MASS package

year <- factor(1992 + (1:n))
head(levels(year))

group <- factor(rep(1,n))
dat0 <- data.frame(x, year, group)

g2 <- glmmTMB(x ~ ar1(year + 0 | group), data=dat0)
summary(g2)
```

bbolker:
For the record: in order to fix this, we probably have to back a long way up to incorporate the information we need (i.e. the name of the *factor* used to construct the covariances, not just the names of the *levels*); by the time we get inside `mkVC()`, it's already too late (we've passed the correlation, sd, a vector of term names (the aforementioned levels), scale parameter, vector of block codes, and a flag for the scale parameter: we need to back up to `VarCorr.glmmTMB` and extract the left-hand side of the name of the RE term (e.g. "row + 0 | Subject" -> "row"), then pass that info to `mkVC()` and attach it as an attribute to the varcor object ...

PR #1191, opened 2025-04-22
Ar memfix

bbolker:
This branch isn't entirely complete (e.g., not fully tested), but resolves most of the issues highlighted in https://github.com/glmmTMB/glmmTMB/issues/1188

* `glmmTMBControl` gains a `full_cor` flag which can be set to TRUE (compute and store correlation matrices for all terms), FALSE (don't store any), or a logical vector (pick which ones to store)
* internally each RE term gains a 0/1 `fullCor` flag
* if `fullCor` is 0, the correlation structure is returned as a 1x1 matrix containing `NaN`
* RE terms now have S3 classes corresponding to their covstruct type, with corresponding S3 methods for correlation and std dev formatting
* I believe this doesn't currently break any tests, should be back-compatible.

To do (possibly):

* much more testing, more examples
* more documentation
* `reformulas` stills needs a bit of work to avoid bottlenecks (see https://github.com/bbolker/reformulas/tree/sparse_contrasts_hack, https://github.com/bbolker/reformulas/tree/sparse_contrasts_fix)
* can probably incorporate a fix for https://github.com/glmmTMB/glmmTMB/issues/1140 fairly easily ...


mebrooks:
Should this efficiency fix improve all the correlation structures? I'm starting to work on a problem with 500 spatial grid cells and wondering if I should use `glmmTMB` or maybe switch to `sdmTMB`. @kaskr, do you know?

PR #1199, opened 2025-05-13
Predict contrast fix

### LINKED ISSUE #1197: Interactions in random effects causes `predict.glmmTMB()` to fail with data subsets
[An issue opened in `vincentarelbundock/marginaleffects`](https://github.com/vincentarelbundock/marginaleffects/issues/1443) resulted in this reprex where the first `predict` fails but the second one works:

```r
library("glmmTMB")
data(Salamanders)

zipm3 = glmmTMB(
        count ~ spp * mined + (1 | site:mined),
        Salamanders,
        family = "poisson"
)

predict(zipm3, newdata = head(Salamanders, 22))

predict(zipm3, newdata = head(Salamanders, 23))
```

Its apparently related to an expected factor level missing in the data.

My current work around is to create the grouping variable explicitly using `paste(site, mined)` in the data and referencing that instead.

bbolker:
fixes #1197.  Spent some time thinking about how to deal with contrast mismatches between prediction data and new data: decided that warning was sufficient (contrasts in new data will be ignored ...)

### DIFF
--- a/glmmTMB/R/predict.R
+++ b/glmmTMB/R/predict.R
@@ -334,6 +334,8 @@ predict.glmmTMB <- function(object,
   }
   augFr <- rbind(object$frame,newFr)
   facs <- which(vapply(augFr,is.factor,logical(1)))
+       ##  I believe all the machinery here is for resolving
+       ##    https://github.com/glmmTMB/glmmTMB/issues/439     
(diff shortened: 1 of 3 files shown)


mebrooks:
I think a check and a warning sounds good since users of the advanced technique should know what they're doing.

PR #1201, opened 2025-05-19
incremental improvements to varcor printing

bbolker:
This may not be worth merging yet (there's more work to be done in this area ...) but opening the PR to make the branch more visible (it wouldn't *hurt* to merge ...)


bbolker:
Right now smooth terms are coded as `homdiag` internally.  The output is good (only prints the variance/std dev associated with the smooth), but it would be good to follow the suggestion [here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/glmmTMB.R#L834-L835) to rename the covstruct type so it can be printed more usefully (ideally the varcorr object would also carry along the original formula for the smooth term ...)

bbolker:
(comment about renaming has moved [here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/glmmTMB.R#L839-L840))

glmmTMB/R/VarCorr.R
```diff
@@ -309,11 +309,19 @@ get_sd.default <- function(x, ...) {
...
 ##' @export
 get_sd.vcmat_ar1 <- function(x, ...) {
     attr(x, "stddev")[1]
 }
 
+##' @export
+get_sd.vcmat_homdiag <- get_sd.vcmat_ar1
+
+##' @export
+get_sd.vcmat_homcs <- get_sd.vcmat_ar1
```
annahuynhly: Coincidentally, this is the same code that I wrote for reformulas (https://github.com/bbolker/reformulas/blob/709cbb7e8bf4fa35a5ea88a06c888df8981ee9c0/R/VarCorr.R#L56)

bbolker:
I think all of this should be covered by the newer integration with `reformulas`.

Issue #1204, opened 2025-06-16
`hetar1` log-likelihood needs to be corrected to provide correct model results


danielinteractive:
Hi `glmmTMB` team,

recently @NikKrieger and I have been working on debugging results from `glmmTMB` models with binary outcome and `hetar1` correlation structure. We found out that the log-likelihood needs to be corrected.

Here is the problem:

## Intro: The `ar1` code

This is the code from `glmmTMB` [here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/src/glmmTMB.cpp#L499) annotated by Nik (`NIK`) and Daniel (`DSB`), and after removing the now not relevant simulation parts:

```{c++}
else if (term.blockCode == ar1_covstruct ||
	   term.blockCode == hetar1_covstruct) {
    // case: ar1_covstruct
    //  * NOTE: Valid parameter space is phi in [-1, 1]
    //  * NOTE: 'times' not used as we assume unit distance between consecutive time points.
    
	// NIK: n is the number of timepoints for each subject
	int n = term.blockSize;
    
	// NIK: theta is a vector containing all the parameters.
	// The last parameter is the correlation (rho).
	// The rest are the logged standard deviations
	
	// NIK: These are all the elements of theta except the last one,
	// so logsd is a vector of all the logged standard deviations.
	vector<Type> logsd = theta.head(term.blockNumTheta-1);
	
	// NIK: This is the last element of theta, which is the correlation.
    // DSB: To be precise, this is the transformed correlation, therefore we need the
    // backtransformation below to get the actual correlation here called phi.
    Type corr_transf = theta(term.blockNumTheta-1);
	
	// NIK: This is the transformation you make to get phi. 
	// see: https://glmmtmb.github.io/glmmTMB/articles/covstruct.html
    Type phi = corr_transf / sqrt(1.0 + pow(corr_transf, 2));
	
	// NIK: Use exp() to create sd, a vector of the actual standard deviations
    vector<Type> sd = exp(logsd);
    Type cursd;
	
	// NIK: For each subject...
    // DSB: Here zero based indexing is used for the loop variable j.
    for(int j = 0; j < term.blockReps; j++){
	
	  // NIK: subtract density for the first timepoint's value for subject j+1
	  // U(0, j) is the value of the first timepoint of subject j+1
      // DSB: yes subject j+1 in one based indexing
	  // mean = Type(0), which is just 0
	  // sd = sd(0), which is the first sd
[...]

PR #1205, opened 2025-06-16
1204: fix `hetar1` log-likelihood

danielinteractive:
This PR fixes the `hetar1` log-likelihood, as laid out in #1204. We tested this in comparison to another software implementation, and the results matched well with this fix. We tried to contribute to the tests and documentation of `hetar1`.

One question came up: Currently, the transformation of the `theta` parameter to obtain the correlation `rho` is not the same as for `ar1`. Is that intended?

This is joint work with @NikKrieger - thanks a lot Nik!


bbolker:
Sorry, can you clarify

>  Currently, the transformation of the theta parameter to obtain the correlation rho is not the same as for ar1. Is that intended?

(¿ maybe ? you mean that the transformations for $\theta \to \rho$ are different for the AR1 and the compound symmetric structures?)

NikKrieger:
Hi @bbolker, perhaps I can clarify:

> Sorry, can you clarify
> 
> > Currently, the transformation of the theta parameter to obtain the correlation rho is not the same as for ar1. Is that intended?
> 
> (¿ maybe ? you mean that the transformations for θ → ρ are different for the AR1 and the compound symmetric structures?)

The appropriate `ar1` transformation is this:
https://github.com/glmmTMB/glmmTMB/blob/b1c2c6560abf4e12701da561671cab2eed7997b9/glmmTMB/src/glmmTMB.cpp#L507

But we discovered that the appropriate `hetar1` transformation is not the same---and it in fact needs to be the same transformation that the compound symmetric structure uses:

https://github.com/glmmTMB/glmmTMB/blob/b1c2c6560abf4e12701da561671cab2eed7997b9/glmmTMB/src/glmmTMB.cpp#L453C4-L454C58

We were wondering whether it was intentional that the `ar1` and `hetar1` correlation parameters as stored in `theta` were transformed differently.

glmmTMB/R/VarCorr.R
```diff
@@ -345,8 +345,7 @@ format_corr.vcmat_ar1 <- function(x, maxdim = Inf, digits=2, ...) {
     } else {
         cc <- format(round(x[2,1], digits), nsmall = digits)
     }
-    cov_struct <- if (inherits(x, "vcmat_hetar1")) "(hetar1)" else "(ar1)"
-    return(matrix(paste(cc, cov_struct)))
+    return(matrix(paste(cc, "(ar1)")))
```
bbolker: is the idea here that it will be obvious to the user whether a homogeneous or a heterogeneous ar1 structure is being used (because hetar1 will print out multiple SD values while ar1 will only print one?)

bbolker:
I guess  I'm still puzzled by [this](https://github.com/glmmTMB/glmmTMB/pull/1205#issuecomment-2977093640).

"the appropriate hetar1 transformation is not the same".

I see your math in #1204, and it all makes sense, but I don't see where it implies we need a different mapping from $\theta$ (unconstrained parameter) to $\phi$ (correlation parameter).

For CS, we have

$$
\begin{split}
a & = 1/(n-1) \\
\phi & = \left(1+\exp(-\theta)\right)^{-1} \cdot (1+a) - a \\
\end{split}
$$

which maps $\mathbf R \to \left( -\frac{1}{n-1}, 1 \right)$ &mdash; i.e. the allowed domain of a CS correlation with a block size of $n$.

For AR1, we have $\phi = \theta/\sqrt{1+\theta^2}$, which maps  $\mathbf R \to \left(-1, 1\right)$ (since the whole domain is allowed for AR1 processes). 

Other than the domain, these are just different smooth curves mapping $\mathbf R$ onto an open set ...

invlogit

It's still not immediately obvious to me why the required transformation/allowed domain is different for hetAR1 vs AR1 ... ??

glmmTMB/R/VarCorr.R
danielinteractive: yeah that was my idea (this is how it was currently printed) - but of course would be good to align with how you treat other homogeneous / heterogeneous covariance pairs for the printing

danielinteractive:
Thanks @bbolker for your fast response!

Yeah regarding the transformation from the last `theta` element to `phi` I am a bit lost. I tried to look through the `glmmTMB` codebase to understand where the `theta` element would be translated to `phi` for printing, but could not really find it - is it done in R or in the TMB reporting in C++?

In any case, the problem we had was that when we first used the AR1 transformation, we ran into examples where the result was wrong. And when we changed to the CS transformation (still using ARH1 as the covariance model, just the transformation), then it worked correctly (our examples were with `phi` close to 1, so the lower bound problem did not appear there).

bbolker:
> the problem we had was that when we first used the AR1 transformation, we ran into examples where the result was wrong. And when we changed to the CS transformation (still using ARH1 as the covariance model, just the transformation), then it worked correctly 

That's interesting, but I think I would still want to know **why**.  (1) the current hetAR1 parameterization doesn't allow the full range of correlations (it *should* be possible to have correlations as low as -0.999 ...) (2) it's awkward to have AR1 and hetAR1 using different parameterizations, unless there's a good reason for it.

* perhaps there was some kind of numerical instability that only manifested in this case, and changing the parameterization fixed it?
* perhaps you accidentally changed something else at the same time?

Do you have a reproducible version of the example that you were working with that wasn't OK with the AR1 parameterization?

danielinteractive:
Thanks Ben, yeah I can cook up the example for sure. I also have not understood this yet.

Just to confirm: Is there any other place in `glmmTMB` where this transformation, or the inverse of it, is defined for each correlation structure? Or not? Because that was my guess somehow that something does not match together there with the C++ code we modified.

bbolker:
I believe it's only in that one place in the C++ code.  The computed correlation matrices are stored/passed back to the R side via `REPORT(corr)`. To confirm, searching for "ar1" everywhere in the R and C++ source code:

``` 
grep ar1 *.R
enum.R:  ar1  = 3,
enum.R:  hetar1 = 12,
glmmTMB.R:               "ar1" = 2,
glmmTMB.R:               "hetar1" = blksize + 1,
glmmTMB.R:        if(ss[i] == "ar1") {
glmmTMB.R:                stop("ar1() expects a single, factor variable as the time component")
glmmTMB.R:##' \item \code{ar1} (autoregressive order-1, homogeneous variance)
utils.R:##' sim_obj <- simulate_new(~ 1 + (1|Subject) + ar1(0 + factor(Days)|Subject),
utils.R:#' fm1 <- glmmTMB(Reaction ~ 1 + (1|Subject) + ar1(0+factor(Days)|Subject), sleepstudy)
VarCorr.R:        ## (diag, ar1, cs, ...)
VarCorr.R:        ## may not want full names (e.g. ou/ar1 models)
VarCorr.R:get_sd.vcmat_ar1 <- function(x, ...) {
VarCorr.R:format_corr.vcmat_ar1 <- function(x, maxdim = Inf, digits=2, ...) {
VarCorr.R:    return(matrix(paste(cc, "(ar1)")))
VarCorr.R:format_corr.vcmat_hetar1 <- format_corr.vcmat_ar1
```


```
$ grep ar1 *.cpp
glmmTMB.cpp:  ar1_covstruct  = 3,
glmmTMB.cpp:  hetar1_covstruct = 12,
glmmTMB.cpp:  vector<Type> times;// For ar1 case
glmmTMB.cpp:  else if (term.blockCode == ar1_covstruct ||
glmmTMB.cpp:	   term.blockCode == hetar1_covstruct) {
glmmTMB.cpp:    // case: ar1_covstruct
glmmTMB.cpp:        if (term.blockCode == hetar1_covstruct) {
glmmTMB.cpp:      if (term.blockCode == hetar1_covstruct) {
glmmTMB.cpp:  } // [het]ar1_covstruct
glmmTMB.cpp:    //  * NOTE: this is the continuous time version of ar1.
```

danielinteractive:
Thanks Ben. OK then probably it was a red herring somehow... 

Here is the example:

```r
# remotes::install_github("danielinteractive/glmmTMB/glmmTMB")
library(glmmTMB)
library(here)
library(tidyverse)

df <- read_csv(
    "https://www.utstat.toronto.edu/~brunner/data/legal/Bball1.data.txt",
    col_names = c("rownum", "Subject", "Hand", "Spot", "Hit"),
    skip = 1
) |>
    mutate(
        Subject = factor(Subject),
        Hand = factor(Hand),
        Spot = factor(
            Spot,
            levels = c("LeftBaseline", "Middle", "RightBaseline")
        ),
        Hit = as.integer(Hit)
    )

head(df)
summary(df)

fm_hetar1 <-
    glmmTMB(
        Hit ~ Hand + Spot + hetar1(0 + Spot | Subject),
        family = binomial(link = "logit"),
        data = df
    )
fm_hetar1
VarCorr(fm_hetar1)
```

The results we would like to have are close to these:

```
# Formula:          Hit ~ Hand + Spot + hetar1(0 + Spot | Subject)
# Data: df
#       AIC       BIC    logLik -2*log(L)  df.resid
#  227.3693  258.1870 -105.6847  211.3693       340
# Random-effects (co)variances:

# Conditional model:
#  Groups  Name              Std.Dev. Corr
#  Subject SpotLeftBaseline  24.53    0.99 (ar1)
#          SpotMiddle        18.28
#          SpotRightBaseline 13.82

# Number of obs: 348 / Conditional model: Subject, 58

# Fixed Effects:

# Conditional model:
#       (Intercept)          HandRhand         SpotMiddle  SpotRightBaseline
#             9.604              2.464             -0.693             -1.679
```

bbolker:
A few quick observations:

* I had to change `read_csv()` to `read_table()` above (this is a space-separated, not comma-separated, file)
* I find random-effects standard deviations of this magnitude extremely surprising for a logit-link binomial model: 24 units on a log-odds scale is *gigantic* ... are you sure you didn't fit this with a Gaussian response ... ? (Although if I fit the same data with a Gaussian response I get even smaller std. dev. estimates ...) (What is the source of these correct estimates? `mmrm`, SAS PROC MIXED, ... ?)

I seem to get the same answer with either the old or the new parameterization (with the correct code), at least to the default-printed precision:

```
Conditional model:
 Groups  Name              Std.Dev. Corr       
 Subject SpotLeftBaseline  0.97032  0.995 (ar1)
         SpotMiddle        1.90355             
         SpotRightBaseline 0.52173             
```

hetar1 model with master branch (wrong) implementation:

```
Conditional model:
 Groups  Name              Std.Dev.   Corr       
 Subject SpotLeftBaseline  1.2952e+00 0.605 (ar1)
         SpotMiddle        4.1546e-05            
         SpotRightBaseline 1.0000e+00            
```

I might be losing my marbles, but could you please double-check your results?

danielinteractive:
Thanks Ben, sorry for not rerunning this actually yesterday (it was pasted together from examples we actually ran), so you hit the error right at the start... 

I will recheck this now and paste results. The example is in any case a very borderline case so not really representative of real applications, as you noted with the large standard deviations of the random effects.

danielinteractive:
Sorry @bbolker for all the confusion with the correlation parametrization!

Just for clarification:
- The example above was wrong as you noted, I had used another modified `Bball2` data set to obtain those results. Was probably too late yesterday evening when I copy/pasted things together, apologies! 
- In that example the correlation parametrization did indeed work better with the wrong `plogis` based parametrization, in the sense that with the default optimizer settings the correct estimates were obtained. ("Correct" in the sense that we got results from SAS `PROC MIXED` which align with that well)
- However, when using the correct correlation parametrization we could still obtain the correct estimates when tweaking the optimizer settings using `glmmTMBControl()` with `optim()`, `BFGS` choice and increasing precision requirements.

In summary, this was kind of a detour, but in the end we arrived at the sensible result that we can use of course the same correlation parametrization as for AR1. I changed the code accordingly, please have a look if anything is missing.

Thanks so much for your great help with this!

danielinteractive:
Thanks a lot @bbolker !

Issue #1206, opened 2025-06-18
document how to get log-likelihood for models with convergence problems


ccsarapas:
Log-likelihood and fit indices are returned as `NA` when there are convergence problems (eg, non-positive-definite Hessian or singular convergence). In #201, it looks like there was a plan to "inform users that they can get the log likelihood of an unconverged model using `-object$fit$objective`," but it looks like this never made it into the documentation? I wasn't able to find it, at any rate, and only figured it out by finding that old issue. Might be helpful to add to the docs, or even as a warning / message when printing the object?

bbolker:
`-object$fit$objective` should work: so should `ee <- obj$env; with(ee, obj$fn(last.par.best[-random]))`.  

Would putting this in the troubleshooting vignette (which is already pointed to by the non-pos-def Hessian warning) be sufficient? Perhaps a bit snarky/pessimistic, but I don't like making the idea of working around this too easy (at the very least "even if you have a non-pos-def Hessian you can get the likelihood as follows ..." should always be accompanied by a "... but you can really get yourself in trouble by papering over cracks in this way" warning ...

ccsarapas:
Including in the troubleshooting vignette, with appropriate caveats and warnings, seems reasonable to me. (FWIW, I'm trying to get this to compare with a model with fewer random effects that does converge. I'm following an example from Singer and Willett's ALDA text (section 5.2.2, on dealing with boundary / non-positive-definite fits) so hope this is appropriate, but might be completely misguided.)

bbolker:
New text in the vignette (third paragraph of the "non-positive-definite Hessian matrix" section):

> You can retrieve the log-likelihoods for such poorly fitted models (and use them to compute AIC and BIC in turn); the log-likelihood is intentionally set to `NA` in the model output to prevent users from inadvertently including poorly fitted models in a model comparison. For a fitted model `fit`, `fit$fit$objective` (if the model was fitted with the default `nlminb` optimizer) will retrieve the *negative* log-likelihood. Alternatively, `fit$obj$fn()` will call the underlying TMB objective function with the best-fitted parameters and should retrieve the same value. *Blindly using such values can be dangerous*; a non-positive-definite Hessian often indicates non-ignorable problems with the model.

What do you think?

ccsarapas:
Looks great!

PR #1209, opened 2025-06-29
update troubleshooting vignette (GH #1206)

bbolker:
Responds to #1206, and cleans up some other bits in the first section (fewer hard-coded values in the text). It would be nice to revisit/update/clean up the other sections as well ...


bbolker:
Gemini's overview is accurate and the comment is reasonable; I did in fact check that `sqrt(diag(vcovs)) ...` is the correct warning message ...

Issue #1193, opened 2025-04-25
Parallel fitting not working on MacOS X with CRAN binaries


gavinsimpson:
Following #910 and some other issues, I note that glmmTMB is instrumented to use OpenMP if detected/instructed to. Unfortunately, the way this is configured in glmmTMB isn't allowing CRAN's binaries for MacOS X to compile glmmTMB with OpenMP support.

The reason what you have doesn't work is that while the build machine for MacOS X has the required runtime, the usual `make` macros that you use in `Makevars` aren't sufficient on that OS to enable support, you need a `configure` script to actually do some checking to see if libomp is found and it R can actually compile something with it.

FWIW, my system info is:

```r
> glmmTMB::omp_check()
[1] FALSE
r$> sessionInfo()
R version 4.5.0 (2025-04-11)
Platform: aarch64-apple-darwin20
Running under: macOS Sequoia 15.4.1

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.1

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: Europe/Copenhagen
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

loaded via a namespace (and not attached):
 [1] nlme_3.1-168        TH.data_1.1-3       estimability_1.5.1  reformulas_0.4.0    minqa_1.2.8
 [6] xtable_1.8-4        zoo_1.8-14          TMB_1.9.17          lme4_1.1-37         grid_4.5.0
[11] MASS_7.3-65         mvtnorm_1.3-3       numDeriv_2016.8-1.1 compiler_4.5.0      multcomp_1.4-28
[16] codetools_0.2-20    sandwich_3.1-1      emmeans_1.11.0      coda_0.19-4.1       Rcpp_1.0.14
[21] mgcv_1.9-3          lattice_0.22-7      nloptr_2.2.1        Rdpack_2.6.4        splines_4.5.0
[26] rbibutils_2.3       Matrix_1.7-3        tools_4.5.0         boot_1.3-31         survival_3.8-3
[31] glmmTMB_1.1.11
```

[...]

PR #1214, opened 2025-07-16
WIP: Support openmp out of the box on macosx

gavinsimpson:
This PR is a starter to address #1193 and allow for parallel fitting using openmp out of the box on macosx.

This PR does a couple of things:

* deletes `Makevars`,
* replaces it with `Makevars.in` and `Makevars.win`
* adds a `configure` script to check for and test openmp on the system where the package is compiled
* adds a cleanup script to remove the `Makevars` created during the running of `configure`
* modifies the `.Rbuildignore` file as its `figure` entry was overly greedy and stopped the `configure` file from being included in the package tarball with `R CMD build`

These changes *should* result in no effect on Linux and Windows but will allow the binary built by CRAN to produce models with parallel fitting.

I have tested this locally on my machine (a M1 Pro Macbook Pro and a M3 Macbook Air) and via Simon Urbanek's mac-builder service. The configure script detects the openmp install on both my machines and Simon's service.

I have retired my Linux machines so I haven't had chance to check this there yet, nor on Windows.

The `configure` script is modified from the one used by the *data.table* package and which has seen a lot of development and testing to make openmp work for macosx binaries.

This is a WIP and needs some testing by others, and also I suspect it will require some changes to the GHActions tests, esp for the macos runner.

## Testing

* With this `configure` and *glmmTMB* installed installed from source, I now see on macosx:

      > glmmTMB::omp_check()
      [1] TRUE

### DIFF
--- a/glmmTMB/configure
+++ b/glmmTMB/configure
@@ -1,5 +1,8 @@
 #!/usr/bin/env sh
 
+## configure script modified from that used by the data.table package
+## https://github.com/Rdatatable/data.table/blob/master/configure
+
 # Find R compilers
 CXX=`${R_HOME}/bin/R CMD config CXX`
 CPPFLAGS=`${R_HOME}/bin/R CMD config CPPFLAGS`


bbolker:
Thanks!  This works for me on R-devel/Linux without any obvious problems.

You say "WIP". Other than testing across a range of platforms and possibly adjusting GH Action runners, is there anything else you think remains to be done ... ??

```r
remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref = remotes::github_pull("1214"))
# * checking if R installation supports OpenMP without any extra hints... yes
glmmTMB::omp_check() ## TRUE
sessionInfo()
```

```
R Under development (unstable) (2025-07-18 r88431)
Platform: x86_64-pc-linux-gnu
Running under: Pop!_OS 22.04 LTS

Matrix products: default
BLAS/LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so;  LAPACK version 3.10.0
```

gavinsimpson:
Thanks @bbolker --- yes, just testing across platforms and also against CRAN's persnickety combination of `R CMD check` flags/setup. It would be good to run this against winbuilder (I didn't get chance yet to do that as I would need to swap out the maintainer email for my own).

It also occurred to me that I should double check the licenses of data.table and glmmTMB to be sure taking something from them to use here is OK and also making sure I have correctly attributed the data.table team.

Other than that I think this is done from a technical point of view, unless other's tests throw up issues.

bbolker:
Re: licensing, seems like the simplest thing is to ask the data.table folks whether they're willing to have the `configure` stuff distributed under AGPL-3, as AGPL-3/MPL-2 combination seems [a little bit complicated](https://talk.openmrs.org/t/mpl-v2-and-agpl-v3-license-question/17419/3) ... (seems like the `configure` file has had contributions from [a lot of different people](https://github.com/Rdatatable/data.table/commits/master/configure) ...) Maybe opening an issue would be the easiest way to have that conversation? (I'd be pretty surprised if they objected.)

mebrooks:
It looks like they put [a lot of thought](https://github.com/Rdatatable/data.table/pull/2456) into the license and changed it from GPL to MPL.

gavinsimpson:
It seems that combining a single file (or two files in this case) with MPL-2 is possible *if* the authors of glmmTMB are OK with shipping the two affectedfiles, `Makevars.in` and `configure`, with the original MPL-2 licence. My understanding is that this is explicitly allowed under MPL-2 (from answer to Q9 in the [FAQ](https://www.mozilla.org/en-US/MPL/2.0/FAQ/)):

> To see the complete set of requirements, read the license. However, generally:
> 
> * You must inform the recipients that the source code is made available to them under the 
>   terms of the MPL (Section 3.1), including any Modifications (as defined in Section 1.10)
>   that you have created.
>
> * You must make the grants described in Section 2 of the license.
>
> * You must respect the restrictions on removing or altering notices in the source code
>   (Section 3.4).

While I approach Matt and the data.table team to let them know about this, @bbolker @mebrooks is this something that would be acceptable to you?

gavinsimpson:
I have opened an issue in the data.table repo to notify them of the intention to reuse some of their code and ask for comments etc: Rdatatable/data.table#7204

mebrooks:
Yes, I'm fine with that. Thanks for looking into it!

bbolker:
If we think @aitap's example [here](https://codeberg.org/aitap/ompdetect) is better than the current hack (and better than OpenMP on macos not working at all out of the box, as is currently the case), is it worth trying/substituting for the current version? (We have a CRAN deadline in 5 days so it would be good to try this soon if we can ...)

gavinsimpson:
@bbolker I still need to test that example against newer XCode versions, as the modified data.table code I added in the PR seems to not work with the latest minor versions. I'll try to look at this tomorrow, but teaching is looming 😱...

aitap:
I can't test this myself, but latest Xcode versions need a different
OpenMP runtime *and* a different path to libomp.dylib:

https://stat.ethz.ch/pipermail/r-sig-mac/2025-May/015192.html
https://github.com/Rdatatable/data.table/issues/6949#issuecomment-2868021414

Presumably, when the CRAN builders update their Xcode versions, the
bundled libomp.dylib will be upgraded as well. I'm not sure what's the
best way for an R package to auto-detect the new Xcode and a compatible
libomp.dylib. One more special case with /usr/local/lib/libomp.dylib
instead of -lomp? Adjusting ~/.R/Makevars after installing the OpenMP
runtime should also work.

bbolker:
New version is up at CRAN, so there is less urgency, but would still be good to integrate this before we lose momentum? (Teaching is looming for me too ...)

mebrooks:
I had to reinstall XCode and CommandLineTools while testing updates for the upcoming release, so I decided to test this branch. This is what I got
```
ℹ Loading glmmTMB
ℹ Re-compiling glmmTMB (debug build)
── R CMD INSTALL ────────────────────────────────────────────────────────────────────────────────────────────────
─  installing *source* package ‘glmmTMB’ ...
   ** this is package ‘glmmTMB’ version ‘1.1.11.9000’
   ** using staged installation
─  checking if R installation supports OpenMP with "-Xclang -fopenmp" ... yes (1s)
   ** libs
   using C++ compiler: ‘g++ (GCC) 14.2.0’
   using SDK: ‘’
   g++ -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -DTMBAD_FRAMEWORK -I'/Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library/TMB/include' -I'/Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/library/RcppEigen/include' -I/opt/R/arm64/include   -Xclang -fopenmp  -fPIC  -falign-functions=64 -Wall -g -O2  -UNDEBUG -Wall -pedantic -g -O0  -c glmmTMB.cpp -o glmmTMB.o
   g++: error: unrecognized command-line option '-Xclang'
   make[2]: *** [glmmTMB.o] Error 1
   ERROR: compilation failed for package ‘glmmTMB’
─  removing ‘/private/var/folders/8t/zsp25_b17td7j1y5d_nmfqt00000gq/T/RtmpufX3QS/devtools_install_23b833adc2d1/glmmTMB’
Error in `(function (command = NULL, args = character(), error_on_status = TRUE, …`:
! System command 'R' failed
---
Exit status: 1
stdout & stderr: <printed>
---
Backtrace:
 1. base::suppressWarnings(roxygen2::roxygenize("glmmTMB", roclets = c("collate", …
 2. base::withCallingHandlers(expr, warning = function(w) if (inherits(w, …
 3. roxygen2::roxygenize("glmmTMB", roclets = c("collate", "rd"))
 4. local load_code(base_path)
 5. pkgload::load_all(path, helpers = FALSE, attach_testthat = FALSE)
 6. pkgbuild::compile_dll(path, quiet = quiet)
 7. pkgbuild:::withr_with_makevars(compiler_flags(debug), { …
 8. pkgbuild:::withr_with_envvar(c(R_MAKEVARS_USER = makevars_file), { …
 9. base::force(code)
10. base::force(code)
11. pkgbuild:::withr_with_envvar(c(DEBUG = "true"), build())
12. base::force(code)
13. pkgbuild::build()
14. pkgbuild:::install_min(path, dest = install_dir, components = "libs", args = if (needs_clean(path)) "--preclean", …
15. pkgbuild::rcmd_build_tools("INSTALL", c(path, paste("--library=", dest, …
16. pkgbuild::with_build_tools({ …
17. base::withCallingHandlers(callr::rcmd_safe(..., env = env, spinner = FALSE, …
18. callr::rcmd_safe(..., env = env, spinner = FALSE, show = FALSE, …
19. callr:::run_r(options)
20. base::with(options, with_envvar(env, do.call(processx::run, c(list(bin, …
21. base::with.default(options, with_envvar(env, do.call(processx::run, …
22. base::eval(substitute(expr), data, enclos = parent.frame())
23. base::eval(substitute(expr), data, enclos = parent.frame())
24. callr:::with_envvar(env, do.call(processx::run, c(list(bin, args = real_cmdargs, …
25. base::force(code)
26. base::do.call(processx::run, c(list(bin, args = real_cmdargs, stdout_line_callback = real_callback(stdout), …
27. (function (command = NULL, args = character(), error_on_status = TRUE, …
28. base::throw(new_process_error(res, call = sys.call(), echo = echo, …
29. | base::signalCondition(cond)
30. (function (e) …
31. asNamespace("callr")$err$throw(e)
Execution halted
make[1]: *** [doc-update] Error 1
make: *** [all] Error 2
```

I can launch R from a Terminal window, so I guess the statement `! System command 'R' failed` is specific to OpenMP.

I guess we should go ahead and get the new version on CRAN without this update, for now. Right?

PR #1218, opened 2025-07-22
Rankdef pred

bbolker:
Fixes problem reported in https://stackoverflow.com/q/77517125/190277  (rank-deficient columns were breaking prediction machinery because we attempted to drop them even though they get dropped in a preceding step)

Incidentally also fixes a problem caused by the development version of `reformulas` (no longer exports a `head.language` method ...


bbolker:
Thanks for quick merge!

PR #1219, opened 2025-07-27
Up2date bugs

### LINKED ISSUE #1211: Failing backward compatibility with 'up2date'
Dear glmmTMB-devs team, 
I encountered an issue with a fitted growth model, fitted under a specific version of TMB (1.9.11) and glmmTMB (1.1.9). 

Making predictions under the package versions, where the model was fitted on, gave no issues. However, when I updated the TMB/glmmTMB versions and wanted to derive predictions now with the old model utilising 'up2date', I got some pretty high residuals. Eventually I wanted to make predictions with the model also adding residual noise within a larger dynamic simulation, but this led to exponentially exploding weights in the fish I was simulating. 
So I was wondering what causes the issue in the backward compatibility and if it`s possible to fix it. 

Reproducible code and the model I am refering to, can be found here: https://github.com/BernhardKuehn/check_growth_model

Many thanks. 
Bernhard

Image

bbolker:
(Hopefully) resolves #1211 (and cleans up/refactors `up2date` slightly)


mebrooks:
This is great! It looks like `rename_fun` will make it easier to update model fits in the future if things change.

PR #1222, opened 2025-07-31
Add the sandwich estimator for ML fitted glmmTMB models

### LINKED ISSUE #443: Function vcovHC from package sandwhich does not work on glmmTMB object
Hello, I am trying to better understand a model I have run using glmmTMB() by following https://stats.idre.ucla.edu/r/dae/poisson-regression/

They use the package "sandwhich" and the function vcovHC() but when I try to use it, I receive an error: 
`Error in UseMethod("estfun") : 
  no applicable method for 'estfun' applied to an object of class "glmmTMB"`

danielinteractive:
closes #443

Included are:

- [x] description in the `model_evaluation` vignette
- [x] `getGroups` method (generic is in `nlme`)
- [x] `bread` (generic is in `sandwich`)
- [x] `estfun` (generic is in `sandwich`)
- [x] `meatHC` (new generic based on `sandwich` function)
- [x] `sandwich` (generic is in `sandwich`)
- [x] `vcovHC` (generic is in `sandwich`)
- [x] `vcov` need to be adapted because it handles the separation of parameters etc.
- [x] make sure we can use the sandwich SEs when ...
  - [x] using `emmeans` on a fit - via passing the `vcov` argument
  - [x] using `summary` on a fit - once `vcov` can return sandwich result this is easy


bbolker:
unfortunately the `getGroups` method signature in `nlme` is 

```
function (object, form = formula(object), level, data, sep = "/") 
```

The signature here is

```
getGroups.glmmTMB <- function(object, level = 1, ...)
```

The signatures have to match *exactly* (up to extra arguments in the derived methods getting swallowed by `...`, if there is one ...)

danielinteractive:
Thanks @bbolker for the fixes - I always tend to forget that when we import a generic function, export a method for it, does not mean that the generic function is exported, too...

I have aligned the `getGroups` signature now with its generic function's definition (and just ignore the unused arguments).

danielinteractive:
@bbolker ok finally the checks are all green 😄

bbolker:
FWIW if we wanted we might be able to re-export the `bread`, `meat`, `vcovHC` generics so users didn't have to prefix with `sandwich::` or load the `sandwich` package, i.e. see https://stackoverflow.com/a/31541044/190277

* I don't know if they would then need to be documented as well?
* I believe that specific Stack Overflow question (and the corresponding section of *Writing R Extensions* refer specifically to S4 methods, but should work for S3 as well ... ??

for example ` grep VarCorr glmmTMB/NAMESPACE`

```
S3method(VarCorr,glmmTMB)
S3method(print,VarCorr.glmmTMB)
export(VarCorr)
importFrom(nlme,VarCorr)
```

`VarCorr.R` has


```
VarCorr.R:##' @importFrom nlme VarCorr
VarCorr.R:##' @export VarCorr
```

whereas `VarCorr.merMod` has just a plain `@export` (which is sufficient to register the S3 method)

I guess it might make sense to put together a single Rd page (via `@rdname`) that (briefly) documents `vcovHC`, `bread`, and `sandwich` ?

I think this would be good, on balance, but am open to arguments in the other direction ("too much work", "just merge the PR already", ...)

PPS Now I see that you already have man pages for most of this so it might be as simple as adding a few `@export` statements ... ??

danielinteractive:
Thanks @bbolker , no problem at all:

- [x] `getGroups`
- [x] `bread`
- [x] `estfun`
- [x] `meatHC` (already exported because it is a new function)
- [x] `vcovHC`

PR #1233, opened 2025-09-03
Zigamma err msg

### LINKED ISSUE #1213: ziGamma distribution family not accepting data with zeros
I have a dataset with some zeros that I'm hoping to use the Gamma hurdle family to model, but it doesn't seem to be working properly. Maybe this is a regression from some earlier version?

Minimum reproducible example, just using some off-the-cuff numbers:

`glmmTMB(y~x, family=ziGamma, data = list(y=c(0, 0, 0, .5, .75, 2, 4), x=c(5, 10, 5, 15, 18, 30, 40)))`

Gives the error:
`Error in eval(family$initialize) :  non-positive values not allowed for the 'Gamma' family`

bbolker:
fix #1213 (improved error message)


bbolker:
@mebrooks 
Even though the new glmmTMB is on CRAN already, it would be worth merging this (very minor) improvement while we're active

PR #1244, opened 2025-12-14
Bump dependency to latest reformulas, removed redundant printing

annahuynhly:
Right now, we're trying to transfer some of the `VarCorr` machinery into `reformulas` so both `lme4` and `glmmTMB` can use them to remove redundancies.

See the new `VarCorr.R` in `reformulas`: https://github.com/annahuynhly/reformulas/blob/main/R/VarCorr.R

Might try to transfer more functions such as `mkVC` in the future...


bbolker:
@mebrooks, I think this OK but out of an abundance of caution could you glance at it and see if you're OK with merging it ... ?

PR #1255, opened 2026-01-08
Fix theta initialization for propto/equalto and enable them in disp/zi components

mmcgillycuddy:
- Fixed bug in t01 function where theta was not properly initialized for propto/equalto random effects for zi and disp params
- Modified mapping function to allow propto/equalto random effects in all model components


bbolker:
thanks! can we have a test for this too?

mmcgillycuddy:
Absolutely. I will add some tests.

PR #1256, opened 2026-01-08
Formalities

mebrooks:
I'm putting this here so we don't forget. Feel free to add news.


bbolker:
We should also remember to remove `Remotes:` from the DESCRIPTION file before submitting. reformulas 0.4.3.1 is on CRAN now, but the windows binaries haven't been built, so ideally I'd like to wait a day or two (I think the R-universe build of glmmTMB will fail if we don't have Remotes: and the windows binaries for reformulas 0.4.3.1 aren't available on CRAN)

bbolker:
I'm going to go ahead and merge this even though r-release windows binaries for `reformulas` are still on 0.4.3 ...

Issue #1260, opened 2026-01-28
Default settings for vcov.glmmTMB() break downstream functions when a parameter is not estimated.


dke-chan:
I have been fitting constrained longitudinal data analysis (cLDA), see https://doi.org/10.1111/j.1541-0420.2009.01332.x. Ideally, when using downstream packages such as `emmeans`, it is best not to fit these models using a modified design matrix. One way to do this in `glmmTMB` is to use the `map` and `start` arguments to ensure parameters are not estimated and are appropriately constrained. However, the default arguments for `vcov.glmmTMB()` produce the following error.

```
Error in mm[nms, nms] <- m : 
  number of items to replace is not a multiple of replacement length
```

Which then breaks `summary.glmmTMB()` and `emmeans::emmeans()`. I find this quite peculiar, as `confint.glmmTMB()` works with the default arguments, and using different arguments for `vcov.glmmTMB()` produces the requested variance-covariance matrix.

I'm unsure what to call this issue, as it isn't a model-fitting bug. But it would be worthwhile to report in case other glmmTMB users run into a similar issue. Example code can be found here, which should reproduce the above error message: https://gist.github.com/dke-chan/c0f31bceb7902f558703888f402e22b7

bbolker:
Presumably making the default `include_nonest=FALSE` would fix this? We could consider this - would be back-incompatible, but might be worth it ?

dke-chan:
I did some further digging into the code, and I think the length of `ff` and rows of `vv` in `summary.glmmTMB()` will be mismatched if you include `include_nonest=FALSE`.

Based on the source code comments for `vcov.glmmTMB()`, https://github.com/glmmTMB/glmmTMB/blob/8e4a3f35fb3a74f56b27465ac2560594ce0e6824/glmmTMB/R/methods.R#L547-L548, I wonder if editing the definition of the `estNameList` in `vcov.glmmTMB()` to have `include_mapped = TRUE` instead of `include_mapped = FALSE` be a better fix if changing anything to fix this issue introduces backwards-incompatibility issues?

dke-chan:
[...]

PR #1269, opened 2026-03-20
Update vcov() to no longer create unnamed columns for non-estimated parameters

dke-chan:
A pull request to address #1260 where vcov.glmmTMB() throws an error message when trying to include entries for non-estimated parameters, and passes the unit tests. Essentially, what happens is that the following call

https://github.com/glmmTMB/glmmTMB/blob/8e4a3f35fb3a74f56b27465ac2560594ce0e6824/glmmTMB/R/methods.R#L518

will add `NA` rows/columns to `m` due to `as.numeric(cur_map)`. The current implementation does not break any of the tests in `./glmmTMB/glmmTMB/tests/testthattest-mapopt.R`, because it fails silently if only the last parameter is fixed.

My proposed fix is to filter out the NA entries in `cur_map` object.

Even though it does not quite fix the issue with `emmeans()` downstream (see https://gist.github.com/dke-chan/c0f31bceb7902f558703888f402e22b7), this can be circumvented by manually passing down the `vcov(fit)$cond` object. Plus, I feel that many things would and could break if the downstream method for `emmeans()` was updated...


dke-chan:
I'm just following up to see whether this pull request will get reviewed.

bbolker:
This seems very reasonable. Could you extend your PR to include (1) a test and (2) a NEWS entry?

dke-chan:
I have done what was requested and should have covered all possible regression test cases..

PR #1276, opened 2026-04-10
fix bug for mapping parameters for rr strucs

mmcgillycuddy:
Prevent map.theta.propto from overwriting map for rr parameters

### DIFF
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -434,8 +434,8 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
     ## add X matrices, prior info
     data.tmb <- c(data.tmb, Xlist, prior_struc)
 
-  # function to set value for dorr
-  rrVal <- function(lst) if(any(lst$ss == "rr") || any(lst$ss == "propto") || any(lst$ss == "equalto")) 1 else 0
+  # check for any covariance structures which have non-default theta values
+  ndThetaStruc <- function(lst) if (any(lst$ss %in% c("rr", "propto", "equalto"))) 1 else 0
 
   getVal <- function(obj, component)
     vapply(obj, function(x) x[[component]], numeric(1))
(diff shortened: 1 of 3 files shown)


glmmTMB/R/glmmTMB.R
```diff
@@ -517,13 +517,14 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
                               start_method = control$start_method)
   }
 
+  
   ### Change mapping for propto 
   for (component in c("cond", "zi", "disp")) {
     rList <- get(paste0(component, "List"))
-    if(rrVal(rList)){
-      restruc <- get(paste0(component, "ReStruc"))
+    if(length(rList$ss) > 0 && any(rList$ss %in% c("propto", "equalto") )){
```
bbolker, line 524: I'm curious: does it make sense to change the definition of `rrVal` above instead of doing a nearly equivalent test here? The current definition is 

```
if(any(lst$ss == "rr") || any(lst$ss == "propto") || any(lst$ss == "equalto")) 1 else 0
```

Is the main point to differentiate `propto`/`equalto` from `rr`, or to make sure `length(lst$ss) >0`, or ... ?

Where else is `rrVal` used/what is it used for?

bbolker (commented):
In addition to comments below about `rrVal`, could you add an item to the NEWS file please ...?

glmmTMB/R/glmmTMB.R
mmaechler: well ... and the above `if(any(...) || any(...) ......)) 1 else 0`
does not look very R-like to me.

Really use the *logical*  `A` instead of  `if(A) 1 else 0`    (!!)
and then also use
   `any(val %in% c(a,b,c))`                          instead of 
   `any(val == a) || any(val == b) || any(val == c)`
no?

mmcgillycuddy: Good point.  The extra terms were added a bit ad hoc. I’ll update the code.

mmcgillycuddy: @bbolker rrVal is only used here as an indicator to initialise theta with non-default values for those structures. I will update the name. 
length(rList$ss) > 0 is needed for components which have no terms. It was the only way I could get it to work without breaking other models. Is there a better way to handle this?

Issue #1278, opened 2026-04-26
should we be dropping unused levels in grouping variables?


bbolker:
[This line](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/glmmTMB.R#L1357), present since 2015, tells R to drop unused levels in grouping variables (specifically, this is passed to the `model.frame()` call). In the example [here](https://stats.stackexchange.com/a/675760/2126), this causes problems (fitting an AR1 model to data with observations missing at some time steps). `reformulas::mkReTrms` has a `drop.unused.levels` argument that is TRUE by default, and `drop.unused.levels = TRUE` is hard-coded in various other places in `lme4` ...  what will break if we change this? (We could warn on unused levels instead? We could allow coarse or fine-grained user control? Adding this to `glmmTMBControl` seems pretty straightforward ...)

```r
library(glmmTMB)
dd <- expand.grid(g = 1:50, time = 1:10) |>
    transform(ftime = factor(time)) |> 
    subset(! time %in% 7:9)

dd$y <- simulate_new( ~ 1 + ar1(0 + ftime | g),
               newdata = dd,
               newparams = list(beta = 0, theta = c(1,1), betadisp = -2),
               family = gaussian,
               seed = 101)[[1]]
fit <- glmmTMB( y ~ 1 + ar1(0 + ftime | g), family = gaussian,
               data = dd)

print(attr(VarCorr(fit)$cond$g, "correlation"), digits=2)
        ftime1 ftime2 ftime3 ftime4 ftime5 ftime6 ftime10
ftime1   1.000   0.67   0.46   0.31   0.21   0.14   0.095
ftime2   0.675   1.00   0.67   0.46   0.31   0.21   0.140
ftime3   0.456   0.67   1.00   0.67   0.46   0.31   0.208
ftime4   0.307   0.46   0.67   1.00   0.67   0.46   0.307
ftime5   0.208   0.31   0.46   0.67   1.00   0.67   0.456
ftime6   0.140   0.21   0.31   0.46   0.67   1.00   0.675
ftime10  0.095   0.14   0.21   0.31   0.46   0.67   1.000
```

bbolker:
```r
remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref = "unused_levels")
fitC <- update(fit, control = glmmTMBControl(drop_unused_levels = FALSE)
```

[...]

PR #1279, opened 2026-04-27
Unused levels

bbolker:
See https://github.com/glmmTMB/glmmTMB/issues/1278

Not necessarily fully baked, but seems reasonable.


bbolker:
Based on Gemini stuff so far, reminds me I should also add tests ... (I think checking `drop_unused_levels` for logical type is overkill ...)

bbolker:
Of the points copilot makes above, the most interesting/difficult one is that ideally we would have finer control: we usually do want to drop unused levels in factors used in fixed effects (we'll end up with a rank-deficient X matrix, which can be handled downstream but is ugly), we may or may not want to drop unused levels in factors used in random effects.  Current approach does this at the level of the full model frame.

That said, allowing this additional control is better than what we had, it at least lets the user decide. (Users can always drop unused levels `a priori`.)

Another alternative would be to set `drop_unused_levels` to FALSE globally, then use `droplevels` to drop levels in specified factors? (Will it ever be the case that we use a specified factor variable as both a fixed effect and as a grouping variable?  This use case doesn't make sense but ...)

Issue #592, opened 2020-06-05
We cannot have Kronecker product / direct product / varComb() as variance structure, right?


SchmidtPaul:
Hi there, great work so far!
I don't think it is possible to fit variance-covariance matrices that are the Kronecker product (_a.k.a._ direct product) of mutiple terms in the model. I am only assuming this, because I could not find any hint about it in the documentation or anywhere else. 

I would be thankful, if you could confirm this and in case I am correct and maybe let me know if it is planned to include this feature.

**Here is a reproducible example:**
```{r}
dat <- agridat::mcconway.turnip %>%
  mutate(unit = 1:n() %>% as.factor,
         densf = density %>% as.factor) %>%
  mutate(date_densf = interaction(date, densf))

# nlme
mod1.nlme <- nlme::lme(
  fixed   = yield ~ gen * date * densf,  
  random  = ~ 1 | block, 
  weights = varIdent(form =  ~ 1 | date_densf),
  data    = dat)

mod2.nlme <- nlme::lme(
  fixed   = yield ~ gen * date * densf,  
  random  = ~ 1 | block, 
  weights = varComb(varIdent(form =  ~ 1 | date),
                    varIdent(form =  ~ 1 | densf)),
  data    = dat)

# glmmTMB
mod1.glmmTMB <- glmmTMB::glmmTMB(
  formula = yield ~ gen * date * densf + (1 | block) + diag(date:densf+0|unit),
  dispformula = ~ 0,   # force variance into G-side
  REML = TRUE,       
  data = dat)

# mod2.glmmTMB not possible?
```
([here are some more details](https://schmidtpaul.github.io/MMFAIR/heterogeneous_error_variance.html#5_models) on this)

glmmTMB_1.0.1 
nlme_3.1-147

Thanks in advance!

bbolker:
You're correct that it's not available at the moment.  It shouldn't be impossible, but we'd have to think about the interface and the implementation ...

If you were thinking about digging in and trying to implement this yourself (don't know your level of interest/technical capability), the implementation of various covariance structures is in the `termwise_nll` function that starts [here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/src/glmmTMB.cpp#L361) ...

SchmidtPaul:
Thank you!
I am not sure if my technical capability is good enough, but I might give it a try in the future.

bbolker:
[...]

PR #1285, opened 2026-06-04
Xar1 covstruct

Pascal-Kueng:
This PR suggests two experimental separable member-by-time covariance structures and new syntax for such structures:

* `homcsxar1(membertime(member, time) + 0 | group)`
* `unxar1(membertime(member, time) + 0 | group)`

They define a Kronecker-product covariance of the form

$$\Sigma = \Sigma_{\text{member}} \otimes R_{\text{AR1,time}}.$$

`homcsxar1()` uses a homogeneous compound-symmetric member covariance crossed with an AR(1) time covariance. `unxar1()` uses an unstructured member covariance crossed with an AR(1) time covariance.

The main use case is dyadic or grouped intensive longitudinal data (with a theoretically arbitrary number of groups and timepoints), where users  need to model dependence both between members/roles and across time.

## Motivation

Existing covariance structures can model same-time member dependence or temporal dependence, but not their separable product as a single latent process. For example, a model like

```r
homcs(member + 0 | group:time) + ar1(time + 0 | person)

```

captures same-time member dependence and within-person autocorrelation as separate terms, but it does not directly imply the lagged cross-member covariance of a separable process:

$$\text{Cov}(u_{A,t}, u_{B,t+h}) = \text{Cov}_{\text{member}}(A,B)\phi^{|h|}$$

A full

```r
us(member_time + 0 | group)
```

could, in principle, represent this pattern, but quickly becomes overparameterized. `propto()` can use a supplied covariance matrix, but does not estimate the member-correlation and AR(1) parameters inside that matrix.

The proposed new structures provide a direct and parsimonious way to estimate this kind of covariance.

While a fully generic Kronecker interface (e.g., crossing any two arbitrary structures) is a conceptually appealing alternative, it is much harder for users to apply in their research than well-documented narrower structures. Additionally, such a general solution would be computationally very demanding for intensive longitudinal data. It would require building and factorizing the full dense member-by-time covariance block. Because the dimensions of this matrix grow multiplicatively with the number of time points, likelihood evaluation scales poorly. By explicitly targeting the `member-by-AR1` case (and expanding to other structures in the future), we can bypass the dense matrix entirely:

## Implementation

To avoid the computational cost of a dense Kronecker product, the C++ implementation does not build the full covariance block. Instead, it uses the equivalent AR(1) innovation likelihood over the member vector at each time point. This achieves the exact same separable covariance structure while scaling efficiently for long time series. 

The helper I suggest:

```r
membertime(member, time)

```

and its alias:

```r
mt(member, time)

```

create the required two-dimensional coordinate factor. The helper is intentionally small: it returns an ordinary factor, built in the same spirit as `numFactor()`, with member labels stored as an attribute for display. During model setup, the factor levels are parsed into member and time coordinates and passed to TMB as integer metadata for the random-effect block.

The implementation supports two or more member levels. Dyads are the motivating example, but the structures are not restricted to two members.

## Proposed user-facing design

Example syntax:

```r
glmmTMB(
  y ~ x + homcsxar1(membertime(member, time) + 0 | group),
  data = dat,
  dispformula = ~0
)

```

```r
glmmTMB(
  y ~ x + unxar1(membertime(role, time) + 0 | group),
  data = dat,
  dispformula = ~0
)

```

The `print()` and `summary()` output use a compact covariance display rather than printing the full member-by-time correlation matrix. For example, `homcsxar1()` reports the member correlation, the AR(1) time correlation, and the implied lag-1 cross-member correlation.

## Limitations and Interpretation

These structures are separable. In particular:

* `unxar1()` allows member-specific variances and an unstructured same-time member correlation matrix, but it uses one common AR(1) parameter across member/role levels.
* The time coordinate must be discrete and unit-spaced. Missing observed rows are allowed within groups as long as the factor levels contain the full intended member-by-time grid. In that case the unobserved latent member-time effects remain part of the random-effect block and are integrated over, so observations at times 1 and 3 still have the implied marginal AR(1) relationship through time 2. Users do not need to pad the data with `NA` rows, but if some member-time cells are globally absent they should construct the coordinate factor with levels from the full grid.
* For Gaussian models, `dispformula = ~0` is usually the right choice if the structured term is intended to account for the residual member-by-time covariance. With the default dispersion model, the fit includes both the separable latent covariance term and an additional independent residual variance.
* For non-Gaussian models, these structures are latent Gaussian random-effect covariance structures on the link scale.
* Standard simulation-based residual diagnostics may need care, especially for Gaussian models where the structured term is being used as a residual covariance model. This PR does not add new residual diagnostic machinery.
* `homcsxar1()` treats member labels as exchangeable, but the labels still need to identify the same individual or stable member position within each group across time.

Future nonseparable VAR-style structures would be needed for role-specific inertia or cross-lag transition parameters. 

## Documentation

This PR updates the covariance-structure list in `vignettes/covstruct.rmd`, the `glmmTMB()` covariance-structure documentation, the `numFactor()`/`membertime()` documentation, and `NEWS.Rd`.

## Validation and Checks

This PR adds tests for:

* formula parsing and parameter counts
* covariance matrix recovery for both structures
* reductions when member correlation and/or AR correlation are zero
* equivalence of `unxar1()` and `homcsxar1()` when member SDs are equal
* missing observations within groups
* malformed coordinate factors and helpful error messages
* factor/numeric member labels in compact output
* more than two member levels
* Gaussian dispersion warnings
* non-Gaussian fits on the latent/link scale

Checks run locally:

```r
devtools::test("glmmTMB", filter = "varstruc|xar1|VarCorr")

```

Result:

```text
PASS 124, FAIL 0, WARN 0, SKIP 0

```

Also run:

```sh
R CMD build glmmTMB --no-build-vignettes

```

and

```r
tools::checkRd("glmmTMB/inst/NEWS.Rd")

```

Both completed successfully.

--- 

I am happy to keep working on this and look forward to any feedback. 

## Transparancy statement

For transparency, I used an AI assistant during implementation, especially for the C++ code and tests. I have reviewed and tested the implementation, and would be happy to do another full pass once there is initial feedback on the scope and general approach.


Pascal-Kueng:
I realize this is a substantial addition to the covariance structure. I’m happy to treat this as a design discussion first or move the broader discussion to an issue if preferred.

Relatedly, #1288 addresses stable dyad/member random coefficients, while this PR addresses separable member-by-time AR(1) dependence across repeated occasions. I believe these aspects are very complementary.

kaskr:
Some quick comments:

> While a fully generic Kronecker interface (e.g., crossing any two arbitrary structures) is a conceptually appealing alternative, it is much harder for users to apply in their research than well-documented narrower structures. 

Yes, a generic Kronecker interface would be nice.

> Additionally, such a general solution would be computationally very demanding for intensive longitudinal data. It would require building and factorizing the full dense member-by-time covariance block. Because the dimensions of this matrix grow multiplicatively with the number of time points, likelihood evaluation scales poorly.

I disagree. TMB's `SEPARABLE` construction, and RTMB's `dseparable` allow one to create such models without Kronecker product expansion.

I propose to keep all your good work on a separate branch for testing, and then make a simple and generic Kronecker re-implementation that can hopefully match the specialized implementation in terms of correctness and speed.

Pascal-Kueng:
Thanks, this is super helpful! I am happy to give this a shot.

To check that I understand your suggestion correctly: would a good next step be to try to re-implement the same two structures (`homcsxar1` and `unxar1`) through a more generic separable backend that can later be extended to additional compatible structures?

I think this could become a generic separable framework, but not immediately support any structure x any structure. Combinations may still need to be added deliberately, because they differ in their parameter counts, coordinate requirements, scale conventions, and efficient TMB representations. 

For an initial low-level but explicit syntax, would something like this make sense?

```r
separable(sepgrid(member, time) + 0 | group,
	homcs(member),
	ar1(time))

separable(sepgrid(member, time) + 0 | group,
	us(member),
	ar1(time))
```

Or would a simpler and higher-level interface be more desirable? For instance:

```r
kron(homcs(member), ar1(time), group = group)
kron(us(member), ar1(time), group = group)
```

Does that match what you had in mind? If so, I'm happy to try this and test and benchmark it against the current implementation in this PR.

bbolker:
Note that we've thought about this a little bit before: https://github.com/glmmtmb/glmmtmb/issues/592

I've always gotten stuck on a couple of steps of how the general Kronecker product machinery would be implemented internally. We need

* a way to process the formulas and expand the data structure used to describe random effects terms: the top-level processing of formulas uses `reformulas::splitForm`, that might need to be adjusted
* modifications to [getReStruc](https://github.com/glmmTMB/glmmTMB/blob/a43566b9e03e90b9350a5181ea97b20f7aa28ea2/glmmTMB/R/glmmTMB.R#L995) that would allow it to break these terms down and process their components
* a way to process these internally, presumably via separate calls to the `nll_termwise()` function followed by a `separable()` statement?  https://kaskr.github.io/adcomp/_book/Densities.html#separable-construction-of-covariance-precision-matrices

Pascal-Kueng:
Thanks both, this was very helpful.

Following @kaskr's suggestion, I have been reworking the prototype to use TMB's `density::SEPARABLE` directly. I have updated this PR, but kept the previous specialized implementation on a separate branch on my fork.

The currently supported combinations for this early prototype are the dense-correlation margins `cs`, `homcs`, or `us` crossed with `ar1`. For the original `us x ar1` and `homcs x ar1` cases, results match the specialized implementation. Speed seems comparable to the specialized structures from the initial PR!

Thanks @bbolker for the pointers. The points you mentioned are exactly the ones I ended up struggling with most. The prototype uses some workarounds or compromises in exactly those areas. A brief summary:

1. `reformulas::splitForm()`

For the user-facing syntax, I implemented:

```r
separable(
    margin1(0 + x1) %x% margin2(0 + x2) | group,
    scale = margin1(0 + x1) # inferred when only one margin can carry scale
)

# For example:
separable(us(0 + member) %x% ar1(0 + time) | group)
```

Internally, `glmmTMB` rewrites this before the usual random-effect processing to:

```r
separable(sepgrid(member, time) + 0 | group, list(...metadata...))
```

The metadata is then passed through the existing `reTrmAddArgs` path.

This keeps the implementation within the existing random-effect machinery and should leave room for extensions like additional margins and different scale modes. It is intentionally limited to one variable per margin and no intercept. More complex marginal designs, such as random slopes, would likely need a much more substantial redesign and may not be a feasible target.

2. `getReStruc()`

`getReStruc()` now recognizes `separable` terms, validates the complete grid, computes marginal theta counts, and passes the information to C++ as fields. This might be fine, but it means the separable margins are slotted into the existing random-effects representation rather than being represented as their own marginal designs. If random slopes etc. should be implemented much later, this part may need to be revised.

3. `termwise_nll()`

The C++ side adds a new branch in `termwise_nll()` for the separable term. It parses theta into the supported marginal structures, constructs the marginal density objects (`MVNORM_t`/`UNSTRUCTURED_CORR_t` and `AR1`), and combines them with `density::SEPARABLE`.

I am not currently calling `termwise_nll()` recursively for each margin, because `termwise_nll()` returns a complete NLL contribution, while `SEPARABLE` needs marginal density objects. The current implementation means that some code is duplicate. A later cleanup could try to refactor the existing covariance code into smaller helpers, if this is desired.

This updated PR is not meant to be merge-ready, but just to get some initial feedback on the current direction. Many things are still missing, like simulation support or clean reporting. 

---


If the general direction looks right to you, what do you think the next useful target should be for this prototype? I could try to:

1. add support for some additional high-priority covariance pairs and generalize the code further where possible as I go
2. try to implement scaling modes  `global()` and `product()` to enable the implementation of combinations of margins that both caryy (or both don't carry) scale
3. try to refactor parts like the marginal-density construction before extending scope and capabilities, if you think that should happen first

kaskr:
I proposed a **simple and generic** re-implementation. This PR is neither.
We want an interface that allows you to combine *any* pair of existing structures, to give the separable extension - including the Kronecker structure itself, so you can do `a %x% b %x% c %x% ...`

The 'metadata' stuff seems overly complex. I'm not convinced it's necessary.

FWIW, the timing of this work is a bit unfortunate. If the GSOC project (https://summerofcode.withgoogle.com/programs/2026/projects/KIFNdzRU) becomes a success, it'll provide a much easier path to implement a general Kronecker structure (because it avoids the limitation in the TMB `SEPARABLE` construct, where every Kronecker operand combination must be expanded compile time).

Pascal-Kueng:
I believe that with the current code structure of glmmTMB, a fully generic implementation cannot also be simple, without large refactors.

The GSoC work seems interesting and may remove some of the backend/C++ complexity, but many R-side issues would remain. To keep the actual Kronecker evaluation path simple and generic, substantial package-level and possibly upstream infrastructure would also be needed on the R side, such as a common margin interface and scale conventions for existing covariance structures, a way to carry margin-specific metadata and model-frame variables through the parser, and generic reporting/simulation/prediction behavior for arbitrary supported margin combinations.

This PR was intended as a compromise/proof of concept for my original use case, with a path for incremental expansion and generalization before merging. The question is whether this limited prototype is useful as a stepping stone and which generalizations to attempt next. 

With that being said, I might keep experimenting a bit on a new branch. I am happy for specific suggestions regarding any aspects that should be addressed or completly changed in their implementation.

PR #1288, opened 2026-06-09
Add indisting covariance structure for indistinguishable dyads

RobAckerman:
Adds `indisting()`, a covariance structure for indistinguishable dyads (Kenny, Kashy & Cook, 2006). The structure imposes equality constraints on the covariance matrix appropriate when dyad members are exchangeable. Implementation follows the `cs` structure approach, passing the constrained covariance matrix directly to `MVNORM_t`. Includes input validation via `check_indisting_order()` with informative error messages.

### DIFF
--- a/glmmTMB/R/enum.R
+++ b/glmmTMB/R/enum.R
@@ -49,7 +49,8 @@
   hetar1 = 12,
   homcs = 13,
   homtoep = 14,
-  equalto = 15
+  equalto = 15,
+  indisting = 16
 )
 .valid_zipredictcode <- c(
   corrected = 0,
--- a/glmmTMB/R/glmmTMB.R
+++ b/glmmTMB/R/glmmTMB.R
@@ -434,8 +434,8 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
     ## add X matrices, prior info
     data.tmb <- c(data.tmb, Xlist, prior_struc)
 
-  ## identify covariance structures that have non-default theta values
-  ndThetaStruc <- function(lst) any(lst$ss %in% c("rr", "propto", "equalto"))
+  # function to set value for dorr
+  rrVal <- function(lst) if(any(lst$ss == "rr") || any(lst$ss == "propto") || any(lst$ss == "equalto")) 1 else 0
 
   getVal <- function(obj, component)
     vapply(obj, function(x) x[[component]], numeric(1))
@@ -458,12 +458,12 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
 
   # theta is 0, 1 for rr_covstruct
   # theta is parameterised to corr matrix for propto
-  t01 <- function(ndTheta, ReStruc, List = NULL){
+  t01 <- function(dorr, ReStruc, List = NULL){
 
     nt <- sum(getVal(ReStruc, "blockNumTheta"))
     theta <- rr0(nt)
     
-    if (ndTheta) {
+    if (dorr) {
       blockNumTheta <- getVal(ReStruc,"blockNumTheta")
       blockCode  <- getVal(ReStruc, "blockCode")
       thetaseq <- rep.int(seq_along(blockNumTheta), blockNumTheta)
@@ -495,9 +495,9 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
                        b       = rep(beta_init, ncol(Z)),
                        bzi     = rr0(ncol(Zzi)),                       
                        bdisp   = rep(betadisp_init, ncol(Zdisp)),
-                       theta   = t01(ndTheta = ndThetaStruc(condList), condReStruc, condList),
-                       thetazi = t01(ndTheta = ndThetaStruc(ziList), ziReStruc, ziList),                       
-                       thetadisp = t01(ndTheta = ndThetaStruc(dispList), dispReStruc, dispList),
+                       theta   = t01(dorr = rrVal(condList), condReStruc, condList),
+                       thetazi = t01(dorr = rrVal(ziList), ziReStruc, ziList),                       
+                       thetadisp = t01(dorr = rrVal(dispList), dispReStruc, dispList),
                        psi  = psi_init
                      ))
 
@@ -517,14 +517,13 @@ mkTMBStruc <- function(formula, ziformula, dispformula,
                               start_method = control$start_method)
   }
 
-  
   ### Change mapping for propto 
   for (component in c("cond", "zi", "disp")) {
     rList <- get(paste0(component, "List"))
-    if(length(rList$ss) > 0 && any(rList$ss %in% c("propto", "equalto") )){
-      restruccomp <- get(paste0(component, "ReStruc"))
+    if(rrVal(rList)){
+      restruc <- get(paste0(component, "ReStruc"))
       mapArg.orig <- mapArg
-      mapArg <- map.theta.propto(restruccomp, mapArg.orig, component)
+      mapArg <- map.theta.propto(restruc, mapArg.orig, component)
     }
   }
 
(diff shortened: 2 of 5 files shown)


Pascal-Kueng:
As a dyadic researcher, thanks for adding this. I think this targets a useful gap!

A few thoughts:

1. For `k > 1`, is the current parameterization guaranteed to stay positive definite? It looks like the constrained correlation matrix is filled directly from independently transformed pairwise correlations. Pairwise correlations in `(-1, 1)` are not sufficient to guarantee that the full covariance matrix is positive definite.

Building on the common/difference (Idiff) approach you have described in your previous work, I wonder whether `indisting()` could use that parameterization internally. For example, users can already fit the equivalent model manually by assigning partners arbitrary but stable contrast codes, e.g. `Idiff = -1/1`, and fitting something like:

```r
(1 + x | coupleID) +
(0 + Idiff + I(Idiff * x) | coupleID)
```

Using this basis internally would guarantee positive definiteness because the common and difference blocks are valid covariance matrices. The member-level variance/covariance terms can then be recovered by transforming back from the common/difference basis.

2. The proposed `P1`/`P2` syntax still requires users to construct indicators and order terms carefully. Would it be possible to instead expose something like `indisting(1 + x | exch(coupleID, memberID))`, where `indisting()` defines the exchangeable random-coefficient covariance, `coupleID` is the grouping factor, and `memberID` identifies exchangeable members within group? I like `exch()` because it could generalize beyond dyads. However, this is, of course, a larger design decision where maintainers would have to weigh in.

3. I also see this as complementary to #1285. For a repeated-measures model, I think the ideal exchangeable-dyad specification would combine stable dyad/member random coefficients with a separable within-dyad member-by-time covariance structure, e.g.:

```r
outcome ~ 1 + x_cw + x_cb +
  ## stable exchangeable dyad/member differences
  indisting(1 + x_cw | exch(coupleID, memberID)) +

  ## within-dyad member-by-time dependence
  homcsxar1(mt(memberID, time) + 0 | coupleID)
```

The first term would handle stable exchangeable random intercepts/slopes, while the second handles same-day partner dependence, within-member AR(1), and lagged cross-partner dependence across repeated occasions. For Gaussian models intended as residual covariance models, this would usually go with `dispformula = ~0`.

I’d be very interested in your thoughts on whether #1285 fits with this direction.

Happy to discuss further and potentially contribute code or tests.

RobAckerman:
Thank you for the kind words. Please find my responses to your questions below.

1. Yes, you are correct about the positive definiteness issue. However, my understanding is that, in practice, a non-PD solution would cause the likelihood to blow up, which the optimizer would flag as a convergence problem anyway. I also note that the current `indisting()` implementation follows the same approach as the `cs()` structure, a helpful suggestion provided by Ben in a previous discussion. 

   Unfortunately, the sum-and-difference approach you suggest only works when there are no missing data. This is one of the motivations for having a native covariance structure like this.

4. Yes, I think that the `indisting(1 + x | exch(coupleID, memberID))` syntax is more user-friendly than the current P1/P2  dummy code approach and I had at one point thought of doing something similar. However, I did not see any other native covariance structures in `glmmTMB` with this structure, and I didn't want to deviate from their structures too much. I suppose if the maintainers feel like this would be beneficial, I would be open to considering it.

5. I took a quick look at #1285, and I agree that the two pull requests are complementary. However, I am much less familiar with this specification, and so unfortunately, I do not feel knowledgeable enough to comment on it. In theory, though, yes, having this functionality would be great.

Best,

Rob

bbolker:
1. Relying on 'bad' returns seems dicey. If there's any reasonably straightforward way to constrain the parameters to guarantee positive definiteness, and to implement the model parameters on an unconstrained space that then gets mapped to the constrained space, that would be better.  For example, the `cs()` correlation structure maps an unconstrained parameter $\theta_{n+1}$ to $(-1/(n+1), 1)$ (where $n$ is the block size), which is the feasible range for $\rho$ that maintains positive definiteness:

$$
a=1/(n-1); \rho = \textrm{plogis}(\theta_{n+1}) \cdot (1+a) -a 
$$

([code here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/src/glmmTMB.cpp#L454-L455))

2. This definitely represents an extension of the syntax (and might require some coordination between `reformulas` and `glmmTMB`, but is at least worth exploring.

RobAckerman:
Sounds good. I will try to see if I can figure something out and report back. Thank you both for your thoughtful feedback!

RobAckerman:
@Pascal-Kueng, your insight about using the sum-and-difference approach internally is actually a great suggestion! Sorry about seemingly dismissing it before, as I had initially thought you meant using the sum-and-difference approach with the observed data (i.e., creating the sum and difference variables, and then using them in the random-effects statement).

So this is what I am thinking (please note that I used assistance from Claude)...

For the dyadic growth curve model (k=2), the `indisting()` structure produces this 4×4 correlation matrix R:

|          | P1_int | P2_int | P1_slope | P2_slope |
|----------|--------|--------|----------|----------|
| P1_int   | 1      | rho_p1 | rho_w    | rho_c    |
| P2_int   | rho_p1 | 1      | rho_c    | rho_w    |
| P1_slope | rho_w  | rho_c  | 1        | rho_p2   |
| P2_slope | rho_c  | rho_w  | rho_p2   | 1        |

where:
- **rho_p1, rho_p2** = partner correlations (P1 with P2, within the same variable type)
- **rho_w** = within-person correlation (same person, different variable type)
- **rho_c** = cross-person correlation (different person and different variable type)

As [Pascal-Kueng](https://github.com/Pascal-Kueng) noted, this R matrix is PD if and only if two smaller 2×2 matrices (for k = 2), a sum block S and a difference block D, are both PD:

**Sum block S (covariance matrix):**

|       | int           | slope         |
|-------|---------------|---------------|
| int   | 1 + rho_p1    | rho_w + rho_c |
| slope | rho_w + rho_c | 1 + rho_p2    |

**Difference block D (covariance matrix):**

|       | int           | slope         |
|-------|---------------|---------------|
| int   | 1 - rho_p1    | rho_w - rho_c |
| slope | rho_w - rho_c | 1 - rho_p2    |

S and D can each be written as a scaled correlation matrix:

```
S = diag(sqrt(a)) * R_S * diag(sqrt(a)),  where a_i = 1 + rho_p_i
D = diag(sqrt(b)) * R_D * diag(sqrt(b)),  where b_i = 1 - rho_p_i
```

where R_S and R_D are the corresponding k by k correlation matrices. Because R_S and R_D are ordinary k by k correlation matrices, they should be able to be parameterized using the same Cholesky parameterization already used inside `UNSTRUCTURED_CORR_t` in TMB. I believe that this guarantees PD by construction because the optimizer works in a fully unconstrained space and can never produce a non-PD matrix.

Once R_S and R_D are chosen, recovering rho_w and rho_c is straightforward:

```
rho_w = (S_off + D_off) / 2
rho_c = (S_off - D_off) / 2
```

where S_off and D_off are the off-diagonal entries of S and D respectively.

The total parameter count is unchanged, the formula syntax for users is unchanged (though I can try implementing the suggestion by @Pascal-Kueng if this approach reasonable), and missing data is still handled correctly since the sum/difference decomposition is internal to the parameterization.

Does this approach seem reasonable, and is this the direction you both had in mind? I wanted to check before trying to implement.

Best,

Rob

Pascal-Kueng:
I think that sounds very sensible! 

Since the sum and difference blocks are just two smaller unstructured blocks, the unstructured-correlation part can probably be handled by the existing `UNSTRUCTURED_CORR_t` implementation, right?

As far as I can tell, the main extra constraint specific to this parameterization seems to be the partner-correlation/scaling terms. Each `rho_p_i` would need to be mapped into `(-1, 1)`. One possible mapping, consistent with the style used elsewhere in the covariance code, might be:

```cpp
rho_p_i = Type(2) * invlogit(theta_i) - Type(1);
```

Then those terms could define the diagonal/scaling parts of the sum and difference blocks, while UNSTRUCTURED_CORR_t handles the valid correlation matrices for the two blocks.

---

I think it might be worth it to open a separate issue about the formula-syntax question, since that is relevant to both this proposal and #1285? I could get around to that potentially next week. 

As a current workaround,  I am encoding the richer formula syntax into a form where the needed metadata is carried as a single extra argument (via a list). For indisting(), it may be worth checking whether the simpler syntax indisting(1 + x1 + x2 | group, exch = memberID) already gets far enough through reformulas::splitForm() to be useful.

RobAckerman:
So I have tried implementing some of the changes discussed above (again, please note that I used assistance from Claude).

1. To guarantee PD, I used the sum-and-difference approach to parameterization. In particular, I parameterize the sum block S and the difference block D using `UNSTRUCTURED_CORR_t`, which guarantees PD by construction via its Cholesky parameterization. Partner correlations use the invlogit mapping suggested by @Pascal-Kueng:

`rho_p_i = 2 * invlogit(theta_i) - 1`

which ensures both `1 + rho_p_i > 0` and `1 - rho_p_i > 0` (positive diagonal variances in S and D). The back-transform is:

```
rho_w = (S_off + D_off) / 2
rho_c = (S_off - D_off) / 2
```

The total parameter count k*(k+1) is unchanged.

2. Although I really like @Pascal-Kueng's suggestion for the new formula syntax, when I tried implementing it, I had to use some preprocessing steps that I began getting nervous about. But I agree that the current set-up I had requiring users to create P1/P2 dummy codes and require consistent person ordering is not ideal. To try to reduce some of these problems, the new implementation replaces the dummy code formulation with a factor variable approach:

```
glmmTMB(outcome ~ c_time +
          indisting(0 + member + member:c_time | dyadID) +
          homcs(0 + member | dyadID:Time),
        dispformula = ~0, REML = TRUE)
```

where `member` is a  factor with exactly two levels. This is a better approach than the previous one because the ordering of the person variables within each variable type is guaranteed by R's factor expansion. I still include informative errors to ensure that the formula syntax is written correctly. In particular, a validation function is run that checks formula in `indisting()` does not include an intercept, that the first term is a factor variable only representing dyad member, that the factor variable has exactly two levels, and that subsequent slope terms interact with this factor variable.

The specific changes are documented inline in the attached patch script `install_indisting_pd`, which modifies three files: `R/enum.R`, `R/glmmTMB.R`, and `src/glmmTMB.cpp`.

I should note that i reran my previous analyses and found that the results with this updated `indisting()` still arrive at the correct solution.

3. I agree with @Pascal-Kueng that a cleaner syntax like `indisting(1 + x | exch(coupleID, memberID))` would be preferable long-term. Thank you, @Pascal-Kueng, for looking into this!.

Please let me know if you both have any further thoughts on how to improve this further, or if you think I am going in the wrong direction.

Thanks!

Best,

Rob

[install_indisting_pd.txt](https://github.com/user-attachments/files/28929202/install_indisting_pd.txt)

RobAckerman:
Bump. Just wanted to let you know that I believe I've addressed your positive definiteness concerns with the updated implementation. The optimizer searches over an unconstrained parameter space, and the mapping to the covariance matrix is constructed so that positive definiteness and the indistinguishability constraints are automatically satisfied for any values of the unconstrained parameters. I'm happy to post more details if needed, or to update the PR whenever you think it's ready for review.

Pascal-Kueng:
From my perspective, this parameterization makes sense conceptually. So I think this now needs feedback from a maintainer/contributor like @bbolker on whether this is the right strategy and what the next steps should be.

bbolker:
I'm not looking **very** closely, but this looks sensible to me so far. I agree that extending/generalizing the formula interface is ultimately a good idea, but for now I'm OK with putting the onus of providing the extra structure on the user (good examples will  help -- these could be added as a new section in the 'covariance structures' vignette). Please go ahead and update the PR with your current implementation so we can take a look ...

RobAckerman:
Great, thanks so much! I will get started on this soon.

RobAckerman:
Hello. I've now updated the PR with the current implementation of `indisting()`. The updated parameterization guarantees positive definiteness via a sum-and-difference block decomposition, following @Pascal-Kueng's suggestion. The formula syntax now also requires that users provide a factor variable for the member identification variable, which I think makes it less likely for users to make errors (though I still include checks and informative error messages). I also added a vignette section for `indisting()` to help users. As always, please let me know if there is anything you would like me to change. Thanks!

bbolker:
I haven't been back to this in a while, hope to look at it soon ...

RobAckerman:
No worries. I figured you were busy. Thanks!

Issue #1265, opened 2026-03-07
profile.glmmTMB(... cl = <custom>): object 'new_cl' not found


HenrikBengtsson:
glmmTMB 1.1.14 produces:

```r
library(glmmTMB)
m <- glmmTMB(count~ mined + (1|site), zi=~mined, family=poisson, data=Salamanders)
workers <- parallel::makeCluster(2)
p <- profile(m, parallel = "snow", ncpus = length(workers), cl = workers)
#> Error in profile.glmmTMB(m, parallel = "snow", ncpus = length(workers),  : 
#>  object 'new_cl' not found
```

This is from:

https://github.com/glmmTMB/glmmTMB/blob/fc14888a24e24a529a8e89159fe82ece4069cb23/glmmTMB/R/profile.R#L135-L145

HenrikBengtsson:
I've created PR #1266 to fix this.

PR #1292, opened 2026-06-21
fix diagnose() scale report and check_zstats documentation

SchmidtPaul:
Three small fixes in `diagnose()`, all R-level / documentation (no C++).

### 1. `check_scales` reports the wrong threshold (and repeats the header)

The "unusually large or small standard deviations" header interpolates the
*vector of flagged SDs* into the `%g` slot instead of the threshold
`big_sd_log10`, so the printed threshold is wrong and the header is repeated
once per flagged predictor.

```r
library(glmmTMB)
set.seed(101); n <- 300
dd <- data.frame(y       = rnorm(n),
                 x_big   = rnorm(n, sd = 1e4),
                 x_small = rnorm(n, sd = 1e-4))
fm <- glmmTMB(y ~ x_big + x_small, dd)
diagnose(fm, explain = FALSE,
         check_coefs = FALSE, check_zstats = FALSE, check_hessian = FALSE)

#> before:
#> predictors with unusually large or small standard deviations (|log10(sd)|>9761.51):
#> predictors with unusually large or small standard deviations (|log10(sd)|>9.1882e-05):
#>        x_big      x_small
#> 9.761511e+03 9.188200e-05

#> after:
#> predictors with unusually large or small standard deviations (|log10(sd)|>3):
#>        x_big      x_small
#> 9.761511e+03 9.188200e-05
```

The other checks already interpolate their tolerance (`big_coef`, `big_zstat`);
this aligns `check_scales` with them.

### 2. `check_zstats` documentation has the ratio inverted

`?diagnose` describes the Z-statistic as "(ratio of standard error to mean)".
The function computes `estimate / std.error` (and prints "Large Z-statistics
(estimate/std err) ..."), so the `@param` is corrected to
"(ratio of estimate to standard error)".

### 3. `is.complex()` on the `eigen()` list in the NPD-Hessian branch

In the NA-Hessian fallback, complexity is tested on the `eigen()` *list*
(`is.complex(eigs)`, always `FALSE`) instead of `is.complex(eigs$values)` as the
sibling branch already does. No observable effect today (that fallback
eigendecomposes the real symmetric `cov.fixed`), but it is an obvious
copy-paste slip, fixed for consistency.

### Note, not fixed here

In that same NA-Hessian branch `bad <- NA` has length 1, so the
`length(bad) == 0` "Hessian seems OK" bail-out never triggers and the
subsequent `for (b in bad)` then iterates over `NA`. I left this alone because
the path is hard to reproduce (it needs `NA` entries in the
Richardson-extrapolated Hessian) and the intended behaviour there looks like a
design decision. Happy to follow up with a separate issue/PR if you'd like a
particular fix.

A regression test for (1) is added in `tests/testthat/test-diagnose.R`.

---
*This change was prepared with the assistance of an LLM.*

### DIFF
--- a/glmmTMB/R/diagnose.R
+++ b/glmmTMB/R/diagnose.R
@@ -24,7 +24,7 @@
 ##' @param big_zstat numeric tolerance for Z-statistic
 ##' @param check_coefs identify large-magnitude coefficients? (Only checks conditional-model parameters if a (log, logit, cloglog, probit) link is used. Always checks zero-inflation, dispersion, and random-effects parameters. May produce false positives if predictor variables have extremely large scales.)
 ##' @param check_hessian identify non-positive-definite Hessian components?
-##' @param check_zstats identify parameters with unusually large Z-statistics (ratio of standard error to mean)? Identifies likely failures of Wald confidence intervals/p-values.
+##' @param check_zstats identify parameters with unusually large Z-statistics (ratio of estimate to standard error)? Identifies likely failures of Wald confidence intervals/p-values.
 ##' @param check_scales identify predictors with unusually small or large scales?
 ##' @param explain provide detailed explanation of each test?
 ##' @return a logical value based on whether anything questionable was found
@@ -105,7 +105,7 @@ diagnose <- function(fit,
         sdvec <- sdvec[sdvec>0 & abs(log10(sdvec))>big_sd_log10]
         if (length(sdvec)>0) {
             model_OK <- FALSE
-            cat(sprintf("\npredictors with unusually large or small standard deviations (|log10(sd)|>%g):\n\n",sdvec))
+            cat(sprintf("\npredictors with unusually large or small standard deviations (|log10(sd)|>%g):\n\n",big_sd_log10))
             print(sdvec)
             prt_explain("Predictor variables with very narrow or wide ranges generally give rise to parameters with very large or",
                         "small magnitudes, which can sometimes exacerbate numerical instability, and may also be appear",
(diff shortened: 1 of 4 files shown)


bbolker:
Thanks for indicating that this was prepared with LLM assistance (to be honest, it smelled a little bit like it ...). Can I ask what prompted the fix? Did you notice problems during your use of `diagnose()` and then fire up a code assistant to fix the problem(s)?

SchmidtPaul:
Yes, feel free to completely ignore it, if not useful. I am in the process of building a "bug hunt" skill/workflow including multiple steps defined in a .md file. Thus, I did not actually have a problem, but basically told the LLM to screen the entire source for potential bugs. This package is one I use frequently and I am hoping that these "bug hunts" lead to helpful PRs, but let me know if it does not.

bbolker:
could you please add a short NEWS item saying that minor bugs in `diagnose()` were fixed?

SchmidtPaul:
Apologies if I'm misunderstanding something, but I believe the NEWS item is already part of the PR - it's in `glmmTMB/inst/NEWS.Rd`, under the `CHANGES IN VERSION 1.1.15` → `BUG FIXES` section (right after the #1265 / #1276 entries):

> `diagnose` now prints the intended threshold value (rather than the flagged standard deviations themselves) in its "unusually large or small standard deviations" report; the `check_zstats` argument documentation was corrected to describe the Z-statistic as estimate/std.error (GH #1292; Paul Schmidt)

Happy to shorten or reword it if you'd prefer something more concise.

bbolker:
oops, thanks.

Issue #982, opened 2024-01-24
Family extensions: ordinal response using cumulative link models and other variants


DrJerryTAO:
As an extension to binary regression, cumulative link models allow ordinal response using the same distribution families. There are also adjacent category, continuing ratio, and stopping ratio models that are closely related but use different probability mass function. Will @mebrooks and other developers consider extending the families to include those for an ordinal response? In these models, zero inflation shouldn't be an issue, and a constant `ziformula = ~0` should be reasonable. However, there will be a threshold-invariant-coefficient assumption to consider, which is commonly known as proportional odds, parallel line, or parallel regression assumption. Christensen (2022) relaxed this assumption and allowed threshold-variant coefficients called "nominal effects". I like the potential to model scale effects through `dispformula = ~ ...` and random effects. 

A useful reference: Christensen, R. H. B. (2022). Cumulative link models for ordinal regression with the R package ordinal. https://cran.r-project.org/web/packages/ordinal/vignettes/clm_article.pdf which allows only cumulative link models but not adjacent category, continuing ratio, and stopping ratio models. 
Yee, T. W. (2010). The VGAM package for categorical data analysis. Journal of Statistical Software, 32(1), Article 1. https://doi.org/10.18637/jss.v032.i10 which does not allow dispersion (i.e., scale) nor random effects.

bbolker:
Hi @MrJerryTAO,
 
 The answer to this feature request is pretty much the same as that for [your other feature request](https://github.com/glmmTMB/glmmTMB/issues/981#issuecomment-1908260729) - "sounds fine, but we don't necessarily have any resources for that".

We should probably add a general FAQ for how we consider feature requests. Speaking just for myself (not *necessarily* on behalf of the other developers, feature requests are more likely to be honored if

* a clean pull request is provided 
* it's an **easy** extension of functionality that's already in the package
* it doesn't require additions to the package API (i.e., other arguments being specified/passed to the TMB code) &mdash; this is particularly paralyzing because API decisions should ideally be future-proof
* it's useful to lots of people
* it's useful to the developers and their friends/colleagues
* it's not already available in another R package

bbolker:
I'm closing this as a duplicate (nearly) of #514

PR #1298, opened 2026-07-16
add ordinal (cumulative link / proportional odds) family (closes #514)

### LINKED ISSUE #514: Implement ordinal regression structures
It would be very useful if glmmTMB supported different types of ordinal regression, e.g. the proportional odds model (cumulative link logit model).

jmgirard:
Implements the `ordinal` family (cumulative link / proportional odds) discussed in #514, following the design sketched there (thresholds in `psi`) and incorporating the feedback on my implementation comment. Closes #514.

## What's included

- `ordinal(link = "logit"|"probit"|"cloglog")` family for ordered-factor responses (1-based integer codes also accepted), supporting the full range of glmmTMB fixed/random-effect structures
- `family_params()` returns the K−1 thresholds named by adjacent level pairs (`"Low|Medium"`, ...)
- `predict(type = "probs")`: n × K per-category probability matrix, with `se.fit` support (ADREPORT)
- `type = "response"` = expected category index E[Y]; `type = "link"` = latent-scale linear predictor
- `simulate()` returns ordered factors on the original response levels
- Dunn-Smyth residuals via the discrete PIT of the cumulative-link CDF
- Tests/validation (against `MASS::polr` and `ordinal::clm`/`clmm`, all three links): coefficients, thresholds, logLik, **and standard errors** (coefficient, delta-method threshold, and predicted-probability SEs) agree to ~1e-5; random slopes + RE correlations match `clmm`; K=2 reduces exactly to binomial-logit; AIC/df bookkeeping matches `clm`/`clmm` exactly; offsets, NA handling, save/reload, K=9 with sparse extreme categories, and complete-separation behavior (diverges gracefully, like `clm`) all covered, plus prediction identities, simulate/refit round trips, error paths, and downstream (emmeans/car) regression tests

## Responses to the review points raised on #514

**1. `length(psi) > 2` and downstream packages.** I ran a downstream audit on an ordinal fit (wine data, `rating ~ temp + contact + (1|judge)`) and a tweedie control (length-1 psi): emmeans (EMMs, contrasts, `type="response"`), `car::Anova`, `broom.mixed::tidy/glance`, `insight::get_parameters/find_parameters/get_predicted/model_info`, `parameters::model_parameters`, `performance::r2/check_convergence`, `confint`, `drop1`, `diagnose` — all pass; the tweedie control is unaffected. (Two of these needed the fix described under point 3.) The easystats folks have signalled on #514 that any remaining special-casing on their side is easy.

**2. Threshold parameterization.** I took the softmax suggestion (Koslik et al 2025) seriously and ran a controlled comparison — same likelihood, same optimizer (nlminb), starts identical on the threshold scale, 50 random starts × 8 conditions (n ∈ {200, 2000} × K ∈ {4, 7} × balanced/sparse-extreme-categories):

| parameterization | found optimum | mean iterations |
|---|---|---|
| softmax (Koslik-style) | 100% | 24.5 |
| raw + Inf-on-violation (`ordinal`-style) | 100% | 27.0 |
| cumsum-exp (`polr`-style) | 100% | 36.1 |

All three are equally *robust*, but softmax is the most efficient, decisively so when extreme categories are sparse (~19–29 vs ~64–72 iterations for cumsum-exp at n=200). **This PR therefore uses the softmax parameterization**: `psi` holds log-weights of the K baseline category probabilities (last fixed to 0) and `theta = qlogis(cumsum(softmax(c(psi, 0))))`. Two side benefits: the equiprobable default start is exactly `psi = 0`, and `psi` has a direct interpretation (baseline log probability ratios vs. the last category). The comparison script is available on request.

**3. Internal `map` for the intercept.** Verified that `modelInfo$map` records only the *user's* map (`NULL` for a plain ordinal fit); the internal map lives only in `obj$env$map`, and user-supplied psi maps compose correctly with it (tested). The audit surfaced that emmeans and `car::Anova` mishandle map-fixed coefficients — for *any* mapped model, not just ordinal (`vcov` drops/NA-fills mapped entries while `fixef` keeps them, so emmeans hits a dimension mismatch and Anova's hypothesis algebra propagates `0*NA = NA` into blank chisq values). Added a small `pad_mapped_vcov()` helper that treats map-fixed coefficients as known constants (zero variance), used by `emm_basis.glmmTMB` and `Anova.glmmTMB`. Regression tests assert EMM contrasts equal the fixed-effect coefficients and Anova chisq equals z². The cosmetic `(Intercept) 0 NA` summary row is now suppressed (only when the map was internal).

**4. E[Y] as the reported mean.** Kept, with the reasoning documented: some scalar per-observation summary is needed for `fitted()`/response residuals/`mu_predict`, and E[Y] is the standard choice (it's what DHARMa-style diagnostics want as `fittedPredictedResponse`, and the alternatives — modal or median category — discard more information). The "real" predictions are `type = "probs"`; docs point users there.

## Not included — proposed sequencing for follow-ups

- **Scale effects** (heteroscedastic latent scale, à la `clm(scale = ~...)`; requested by @mmrabe above): planned as the *next* PR. Implementation is small — divide `(theta − eta)` by `exp(eta_disp)` in the likelihood and let the existing `dispformula` machinery drive it — but it changes the semantics of `sigma()`/`predict(type="disp")` for this family and, more importantly, the headline combination (scale effects **with** random effects) has no frequentist reference implementation to validate against (`clm` has scale but no REs; `clmm` has REs but no scale), so it needs simulation-based validation rather than riding along here.
- **Nominal (category-specific) effects** (requested by @qdread above): deferred pending an API design decision — these need a new user-facing formula argument (`nominal = ~ ...` or similar), and per the maintainers' criteria on #982, API additions deserve their own future-proofing discussion. I'd propose opening a follow-up issue with a design sketch once this PR settles, rather than baking in an interface unilaterally.
- **Automatic DHARMa support**: blocked on a small DHARMa-side fix, not anything here — DHARMa's `getObservedResponse()` applies its binomial factor convention (`as.numeric(f) − 1`, 0-based) while simulations arrive as 1-based category codes, an off-by-one that makes `simulateResiduals()` degenerate; it also doesn't know to treat the ordinal family as an integer response. I'll submit a patch to DHARMa once this merges. Meanwhile the manual route works correctly (KS uniformity p ≈ 0.86 on a well-specified model):
  <details><summary>working <code>createDHARMa()</code> recipe</summary>

  ```r
  sr <- DHARMa::createDHARMa(
      simulatedResponse = sapply(simulate(fit, 250), as.numeric),
      observedResponse  = as.numeric(model.response(model.frame(fit))),
      fittedPredictedResponse = predict(fit, type = "response"),
      integerResponse = TRUE)
  ```
  </details>
- **OSA residuals**: feasible in principle via TMB's discrete one-step-prediction machinery (the `keep` indicator is already in place for this family), but Dunn-Smyth residuals and the DHARMa route above already cover the residual-diagnostic need, so I'd treat this as demand-driven rather than planned.

## Notes for review

- Numerical robustness: cumulative log-probabilities go through `logit_inverse_linkfun()` (accurate tails for probit/cloglog via `logit_pnorm`/`logit_invcloglog`) and the middle-category likelihood uses the branch-free identity `log(plogis(s1)−plogis(s2)) = logspace_sub(s1,s2) − log1pexp(s1) − log1pexp(s2)`, so probit/cloglog fits keep finite NLL/gradients at extreme η; thresholds are computed by exact prefix/suffix logsumexp, so no overflow when one category weight dominates (both tested)
- `confint()` reports delta-method Wald CIs for thresholds (analytic Jacobian of the softmax transform); `car::Anova` type III gives an NA row for the fixed intercept rather than an error; integer-coded responses warn that K is inferred as `max(code)`
- REML runs and behaves sanely (RE SD increases as expected) but note the thresholds (`psi`) are *not* integrated out alongside `beta` — happy to take direction on whether that's the definition you want, or whether REML should warn/error for this family
- No new TMB DATA entries and no changes to existing families' parameter structures, so no `up2date()` burden
- `enum.R` regenerated via `make enum-update`; docs via roxygen (kept at RoxygenNote 7.3.3 formatting where possible)
- Full test suite + validation results in the PR checks; happy to walk through any part of the diff

*Disclosure, as on #514: developed with substantial help from an AI assistant (Claude), with design directed by the #514 discussion and every numeric claim validated against `MASS::polr` / `ordinal::clm` / `clmm`.*


bbolker:
This looks good and super-thorough. A couple of quick comments from skimming:

* I'm happy that softmax appears to perform well, as that's also what I would have preferred on theoretical/aesthetic grounds
* would you be willing to post a separate PR for your `pad_mapped_vcov`/emmeans/Anova fixes and tests? That has value and would be worth incorporating immediately, while we discuss any other issues with this one ...
* what tolerance parameter did you use in softmax? (I know, I could look at the code ...)

jmgirard:
Thanks! Both done:

* Split PR: [#1299](https://github.com/glmmTMB/glmmTMB/pull/1299) has the `pad_mapped_vcov`/emmeans/Anova fixes + tests, standalone against master. Splitting it out surfaced (and fixed) one more case our ordinal tests missed: with a general user-mapped coefficient, Anova type II printed blank statistics via `0*NA = NA`, and terms containing only mapped coefficients needed the same NA-instead-of-error handling as type III. Once that merges I'll rebase this PR and its copy of those fixes drops out of the diff.
* Softmax tolerance: none — there's no tolerance/smoothing parameter. Koslik et al. enforce their (unimodality) constraints via a smoothed penalty with a parameter ρ; here I used only the softmax reparameterization idea, exactly: theta_j = logsumexp(psi[1..j]) − logsumexp(c(psi[(j+1)..], 0)) (≡ qlogis(cumsum(softmax(c(psi, 0)))), computed in prefix/suffix-logsumexp form so it's overflow-proof). Monotonicity holds by construction (strictly, for all finite psi), so no constraint handling is needed. One honest caveat: in double precision, thresholds can tie if a category's fitted probability underflows past machine epsilon (adjacent log-weights differing by ≳36); the likelihood then correctly returns `Inf` for observations in that category (cleanly — the branch-free logit-scale form can't produce NaN there) and step-halving recovers, i.e. the same graceful behavior as ordinal's Inf-on-violation, but only reachable when a category is essentially unobserved. Glad the choice matches your aesthetics — the iteration-count experiment agreed with you too.

jmgirard:
Rebased onto current master, so this is mergeable again. Now that #1299 is merged, its `pad_mapped_vcov`/emmeans fixes have dropped out of this diff as expected, leaving just the ordinal family.

The one conflict was with #1303: `dunnsmyth_resids()` gained a `size` argument for binomial-type responses, while this branch had factored the randomized-quantile step into `pit_norm_resids()` for the ordinal PIT path. I kept the ordinal branch, threaded `size = wts` through the non-ordinal branch, and left `pit_norm_resids()` as the shared tail. Existing tests cover both paths. Family enum values don't collide (`combinomial = 102`, `ordinal = 1100`). Full suite passes locally against TMB 1.9.23.

One open question from the original post that I don't think I should decide unilaterally: REML runs and behaves sanely, but the thresholds (`psi`) are not integrated out alongside `beta`. That's defensible since they aren't regression coefficients, though it isn't REML in the strict sense either. Options: leave as-is and document, integrate `psi` out along with `beta`, or warn/error for this family until someone needs it. The first and third are both a few lines. Let me know which you prefer.

bbolker:
I would say add a warning for now (option 3).

Glad this is getting close!

jmgirard:
Added the warning (option 3) in 847d6ba5: `REML = TRUE` with the ordinal family now warns that the fixed effects are integrated out but the thresholds are not, and the fit proceeds. Documented in the family entry and NEWS, with a test.

Reviewing that turned up a related bug, fixed in ea6c7f70. Because the ordinal intercept is always map-fixed, a null model has nothing left to integrate out, so `sdreport()` returns no joint precision matrix and `vcov()` failed with "attempt to set an attribute on NULL", taking `summary()`, `confint()` and `car::Anova()` with it:

```r
glmmTMB(Sat ~ 1, weights = Freq, data = MASS::housing,
        family = ordinal(), REML = TRUE)
```

The bug is general rather than ordinal-specific (`y ~ 0` does the same for any family), so I've submitted it separately as #1316, same as we did with `pad_mapped_vcov`. This branch carries a copy so it stands alone; the hunk is byte-identical to the one in #1316, so it will drop out cleanly when that merges.

PR #1299, opened 2026-07-16
fix emmeans and car::Anova for models with map-fixed coefficients

jmgirard:
Split out of #1298 at @bbolker's request — these fixes are independent of the ordinal family and apply to any model with fixed-effect coefficients fixed via `map`.

**The bugs** (for any model with `map = list(beta = factor(...))` containing `NA`):
- `emmeans` errored with a dimension mismatch (`(subscript) logical subscript too long`): `emm_basis.glmmTMB` pairs the full-length `fixef()` vector with the *reduced* `vcov(include_nonest = FALSE)` matrix.
- `car::Anova` type II silently printed **blank test statistics**: the full-size vcov carries `NA` rows/columns for mapped coefficients, and `0 * NA = NA` propagates through the hypothesis-matrix algebra.
- `car::Anova` type III hard-errored (`Lapack dgesv: system is exactly singular`) when a term contained only mapped coefficients.

**The fix**: a small `pad_mapped_vcov()` helper that treats map-fixed coefficients as known constants with zero sampling variance — padding the reduced vcov back to full size, or zeroing the NA rows/columns of the full-size one. It is applied only to the *default* vcov in `emm_basis.glmmTMB` and `Anova.glmmTMB`; a user-supplied `vcov.` is used verbatim. Wald hypothesis rows involving only zero-variance coefficients are dropped, so those terms report `NA` rather than erroring (both type II and type III).

**Tests**: the helper's three branches (reduced, full-size, no-op), emmeans SEs and contrasts against the freely-estimated model, Anova II/III with mapped intercepts and mapped non-intercept coefficients, and verbatim use of user-supplied `vcov.`.

Notes for review:
- `pad_mapped_vcov` reads the *internal* map (`obj$env$map`), so it also covers maps applied internally by the package (the ordinal family in #1298 fixes its intercept this way, which is how these surfaced).
- I deliberately did **not** change `vcov()` itself: its NA convention distinguishes "non-estimable" from "estimated," and changing that contract has a much larger blast radius. If you'd rather model "fixed-known vs non-estimable" inside `vcov()`, happy to discuss — this PR keeps the change minimal at the consumer level.
- #1298 will be rebased on this once merged (it currently contains an equivalent version of these fixes; the diff there will shrink accordingly).

### DIFF
--- a/glmmTMB/R/Anova.R
+++ b/glmmTMB/R/Anova.R
@@ -145,7 +145,9 @@ Anova.II.glmmTMB <- function(mod, vcov., singular.ok=TRUE, test="Chisq",
         ## hypothesis rows involving only map-fixed coefficients (known
         ## constants, zero variance) are untestable; drop them so the
         ## term gets an NA row instead of a singular-matrix error
-        zv <- diag(vcov.) == 0
+        ## (guard against NA variances from user-supplied vcov, e.g.
+        ## include_nonest=TRUE, which would make any(zv) return NA)
+        zv <- !is.na(diag(vcov.)) & diag(vcov.) == 0
         if (any(zv) && nrow(hyp.matrix.term) > 0) {
             hyp.matrix.term <- hyp.matrix.term[!apply(hyp.matrix.term, 1,
                                           function(x) all(x[!zv] == 0)), ,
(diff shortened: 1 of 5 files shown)


jmgirard:
The NA-diagonal guard here is right, but accepting the full autofix also deleted the map-fixed row-dropping block — so zv is now computed but never used (the `if (any(zv) ...)` that consumed it is gone), and `Anova.III` no longer drops hypothesis rows involving only map-fixed coefficients. That's the behavior 8ce77019 added to give untestable terms an NA row instead of a singular-matrix error, and `Anova.II` still has it, so the two paths have diverged.

If the intent was just to fix the NA-diagonal crash, the minimal change is to keep the block and only harden the zv line:

```r
zv <- !is.na(diag(vcov.)) & diag(vcov.) == 0
if (any(zv) && nrow(hyp.matrix) > 0) {
    hyp.matrix <- hyp.matrix[!apply(hyp.matrix, 1, function(x) all(x[!zv] == 0)), , drop=FALSE]
}
```

The dead `zv` makes me think the autofix deleted more than intended — was dropping the row-filter deliberate, or a side effect of the suggested diff?

bbolker:
ooops, sorry for being sloppy. Are you willing to edit this PR according to your suggestion (and add a test) ?  (I could also do it but am juggling a couple of other things.)

Issue #1300, opened 2026-07-20
Adding Conway-Maxwell-Binomial regression to glmmTMB


jbogomolovas2:
Hi,

First of all great tool: love it, use it, and reccomend to others.
I am wondering would you be interested in adding CMB to the arsenal of distributions families.
I already did some prepwork: key solvers for CMB are in TMB already, regression itself lives on my fork: https://github.com/jbogomolovas2/glmmTMB/tree/combinom-family
It has been tested, connected to emmeans and DHARMa as well.

Regards,
Julius

bbolker:
in principle this sounds fine -- I'd probably like to move to 'squash merges', so that all commits get squashed into a single commit upon merging the PR.

In any case, I think you should go ahead and do a PR. (Does it work with `performance::check_model()` ? Are Dunn-Smyth residuals available?)

jbogomolovas2:
Will do! Thank you

PR #1303, opened 2026-07-23
Add Conway-Maxwell-Binomial (combinomial) family

jbogomolovas2:
# Add Conway-Maxwell-Binomial (`combinomial`) family

As discussed in #1300 — here it is.

Adds `combinomial()` (mean-parameterized CMB, modeled on `compois`) for
under-/over-dispersed binomial-type responses. `allow_negative_nu = TRUE`
switches the dispersion model to an identity link so ν can go negative
(super-dispersion).

**Heads up:** needs TMB newer than 1.9.21 — the `dcombinom2` atomic
([kaskr/adcomp#415](https://github.com/kaskr/adcomp/pull/415)) isn't on CRAN
yet, so CI will fail to compile against CRAN TMB. Same situation `compois` was
in back in the day. `remotes::install_github("kaskr/adcomp/TMB")` to build.
CI/timing is your call.

To your two questions:

- **`check_model()`**: works (see below) — simulation is implemented, so
  all the DHARMa-based panels are backed by real CMB draws.
check_model_cmb_after
- **Dunn-Smyth**: yes — via DHARMa, and also wired into the analytic
  `type = "dunn-smyth"`. Doing that meant passing `size` through
  `dunnsmyth_resids()`, which incidentally fixes the existing analytic path
  for binomial `cbind(k, n-k)` responses with n > 1 (it was using
  `size = 1`); regression test included. Pearson residuals work too.

Tests included; `devtools::check()` clean locally (modulo the TMB version
thing). One loose end elsewhere: DHARMa's family allow-list emits a cosmetic
`nKcase` warning for two-column non-binomial responses — I'll send them a
small PR adding `combinomial`.

Refs: Shmueli et al. 2005 (JRSS C); Kadane 2016 (Bayesian Analysis);
Huang 2017 (Stat. Modelling) for the mean parameterization.


bbolker:
Am I correct that we could depend on TMB >= 1.9.23 (on CRAN now)?

Out of curiosity - do you know if the apparent heteroscedasticity detected by the upper right panel of `check_model()` is real? (Maybe it comes from our failure to compute *standardized* residuals, which is a big can of worms about getting leverage calculations right ... ???)

jbogomolovas2:
> Am I correct that we could depend on TMB >= 1.9.23 (on CRAN now)?
Yes. I checked and it is in it.
> 
> Out of curiosity - do you know if the apparent heteroscedasticity detected by the upper right panel of `check_model()` is real? (Maybe it comes from our failure to compute _standardized_ residuals, which is a big can of worms about getting leverage calculations right ... ???)

I had same worry:
homogeneity_explained
the residuals in that panel aren't scaled by the family variance function V(mu) at all, only by a constant sigma-hat. For any binomial-type family the response-scale spread is proportional to sqrt(p(1-p)), maximal near p = 0.5, so the panel has to bow even for a perfectly specified model. If I Pearson standardize it flattens out.

bbolker:
That sounds like a bug worth reporting to the `performance` folks ... ?

jbogomolovas2:
> That sounds like a bug worth reporting to the `performance` folks ... ?

done:
https://github.com/easystats/performance/issues/926
I nibbled in a bit more and that is a bit more serious issue on their end.

jbogomolovas2:
> > That sounds like a bug worth reporting to the `performance` folks ... ?
> 
> done: [easystats/performance#926](https://github.com/easystats/performance/issues/926) I nibbled in a bit more and that is a bit more serious issue on their end.

Made a PR
https://github.com/easystats/performance/pull/927

PR #1302, opened 2026-07-22
Anova ddf

bbolker:
this adds a `ddf` argument to be able to do F tests in `anova()`. It's Claude-generated (under close supervision).

PR #1308, opened 2026-07-28
Speed up dof_satt() via caching and cheaper finite differencing

bbolker:
## Summary
- `dof_satt()` recomputed an expensive Hessian/Jacobian (involving repeated `TMB::sdreport()` calls) from scratch on every call; now cached on `model$obj$env` and reused across repeated calls on the same fit.
- Switched `numDeriv::jacobian()` from the default Richardson differencing to `method = "simple"`, since the resulting ddf are approximate anyway.
- Benchmarked ~5.8x faster per call, ~36x faster on repeated calls to the same model; ddf values agree with the prior implementation to ~0.03%.
- Added a Satterthwaite-vs-lmerTest regression test to `test-ddf.R` (hard-coded reference values, no new test dependency) — the existing suite only self-consistency-checked `dof_satt()` against `summary()`'s ddf column, with no independent external reference the way the Kenward-Roger tests check against `pbkrtest`.

Full investigation, benchmarks, and comparison with lmerTest's implementation (which motivated this change) are in `misc/README_satterthwaite.md`.

Branched off `master`, independent of PR #1302 (`anova_ddf`) — see the README's "Future work" section for how the two should be combined later.

## Test plan
- [x] `tests/testthat/test-ddf.R`: 6/6 pass (4 existing + 2 new)
- [x] Verified `dof_KR()`/Kenward-Roger unaffected
- [x] Verified `summary(fit, ddf = "satterthwaite")` p-values match end-to-end (4+ decimal places) against the pre-change implementation

### DIFF
--- a/glmmTMB/R/denom_df.R
+++ b/glmmTMB/R/denom_df.R
@@ -323,22 +323,61 @@ dof_KR <- function(model) {
     lengths(lapply(glmmTMB::ranef(model)$cond, colnames))
 }
 
+## Precompute (and cache) the pieces needed for Satterthwaite denominator-df
+## calculations that depend only on the fitted model, not on the specific
+## contrast(s) being tested: the (inverse) Hessian of the negative
+## log-likelihood with respect to the variance/dispersion parameters
+## ("kappa"), and the Jacobian of cov(beta) with respect to those same
+## parameters.
+##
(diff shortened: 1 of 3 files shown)


bbolker:
Closing for now -- planning to merge this together with the other two of #1302/#1306/#1308 (plus additional ddf-consistency fixes) into a single combined PR. Branch is not being deleted.

Issue #1310, opened 2026-07-30
Cannot reproduce behavior of older version of glmmTMB degrees of freedom from Gaussian LMM


qdread:
Hello, I had to go back and reproduce some glmmTMB + emmeans results that I created about 1 year ago, in June 2025. Not quite sure what version of glmmTMB I was using at the time. I am now using glmmTMB version 1.1.14. At the time, I fit a Gaussian LMM with some random effects and then used `emmeans()` to get estimated marginal means for each level of a fixed effect. It used some kind of denominator degrees of freedom approximation which were used to get t-statistics for the pairwise mean differences and their associated p-values.

Now I went back to run the code, now using the more recent version of glmmTMB, and I see that it is now setting the denominator degrees of freedom to infinity and giving me asymptotic confidence intervals and z-statistics for the pairwise differences. I get that, but I need to reproduce the behavior of the older version so that I can recreate a figure for a manuscript revision without having to redo the whole analysis. Is there an easy way for me to reproduce the older behavior of Gaussian LMM in glmmTMB without having to roll back the glmmTMB version? Or if you can just tell me what the default denominator degrees of freedom approximation was, I can apply it.

(By the way I tried to find the relevant change in the NEWS file but could not pinpoint what may have caused this change.)

thanks so much!!!

bbolker:
Can you give us a minimal reproducible example please? The default was probably Satterthwaite. Are you also using the same version of `emmeans` as before?  (I will dig into this a bit, there have been some recent bug fixes in this area that are still sitting in pull requests.  To be honest I'm a bit surprised if/how you managed to get ddf in older versions ...)

This mostly works but reveals at least one more bug ...

```r
set.seed(101)
remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref="emmeans_fix")
library(glmmTMB)
library(emmeans)
dd <- expand.grid(f = factor(LETTERS[1:4]), g = factor(1:10), rep = 1:10)
form <- y ~ 1 + f + (1|g)
dd$y <- simulate_new( form[-2],
                     newdata = dd,
                     family = gaussian,
                     newparams = list(beta = 1:4, betadisp = -1, theta = 0))[[1]]
m <- glmmTMB(form, data = dd, family = gaussian, REML = TRUE)
pairs(emmeans(m, "f", ddf = "asymptotic"))
pairs(emmeans(m, "f", ddf = "satterthwaite")) ## slow
[...]

Issue #1304, opened 2026-07-23
Error using emmeans with satterthwaite df


aosmith16:
I was trying to use emmeans.glmmTMB with satterthwaite degrees of freedom. I get an error about not being able to find function `dof_satt()`, although that does get loaded along with the rest of the functions in glmmTMB.

A reprex using a built-in model. The reprex is a GLMM but I see the same with LMMs:
``` r
library(glmmTMB) # v. 1.1.14
#> Warning: package 'glmmTMB' was built under R version 4.4.3
library(emmeans) # v. 2.0.4
#> Welcome to emmeans.
#> Caution: You lose important information if you filter this package's results.
#> See '? untidy'
salamander1 <- up2date(readRDS(system.file("example_files", "salamander1.rds", package="glmmTMB")))
emmeans(salamander1, ~ mined, type="response", ddf = "satterthwaite")
#> performance of Kenward-Roger and Satterthwaite approx for GLMMs is poorly understood
#> Error in `dof_satt()`:
#> ! could not find function "dof_satt"
```

<sup>Created on 2026-07-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

bbolker:
Can you try this with `remotes::install_github("glmmTMB/glmmTMB/glmmTMB", ref = "emmeans_fix")` and see whether that resolves it?

aosmith16:
Yep, that resolves it! Thanks for looking into this, much appreciated.

PR #1311, opened 2026-07-31
Emmeans fix

bbolker:
This is a new PR cleaning up a variety of emmeans-related stuff. From the NEWS:

**New features**
- `anova.glmmTMB` gains a `ddf` argument, matching `summary.glmmTMB`: the default `"asymptotic"` gives the usual likelihood ratio test, while `"kenward-roger"` or `"satterthwaite"` instead give an F-ratio test between each pair of nested models
- `car::Anova` (Type II/III) also gains a `ddf` argument: `"kenward-roger"` or `"satterthwaite"` give an F-ratio table instead of the default Wald chi-squared table, with each term's denominator df computed via the same machinery as `anova.glmmTMB`/`summary.glmmTMB`

**Bug fixes**
- `emmeans()` with `ddf = "satterthwaite"` failed with "could not find function `dof_satt`" because emmeans strips the enclosing environment of the degrees-of-freedom function before calling it (GH #1304, Ariel Muldoon)
- relatedly, `emmeans()` with `ddf = "satterthwaite"` would have failed for any request involving more than one contrast (e.g. pairwise comparisons) once the issue above was fixed, because the Satterthwaite degrees-of-freedom function was being called with a single contrast vector rather than the one-row matrix it expected (GH #1304)
- `emmeans` and `car::Anova` now work for models with fixed-effect coefficients fixed via `map` (previously a dimension mismatch in `emmeans`, and a singular-matrix error or `NA` test statistic in `Anova`); such coefficients are treated as known constants with zero variance

**User-visible changes**
- Satterthwaite ddf approximation now uses a faster, slightly less accurate approximation method; results may change slightly
- fixed garbled warning from Kenward-Roger correction (GH #1310)
- `ddf = "kenward-roger"` on a non-REML fit now throws an error consistently across `summary`, `anova`, `Anova`, and `emmeans` (previously `emmeans` silently substituted `"satterthwaite"` with a warning instead); `"kenward-roger"` and `"satterthwaite"` on a non-Gaussian family now consistently emit a warning (rather than a message, or nothing) across all four


bbolker:
@mebrooks , I think this is ready to merge ...

glmmTMB/R/denom_df.R
```diff
@@ -323,22 +323,71 @@ dof_KR <- function(model) {
...
+    kappa_opt <- model$fit$par
+    h_kappa <- numDeriv::jacobian(func = model$obj$gr, x = kappa_opt, method = "simple")
+    ## one-sided ("simple") finite differences are not guaranteed to give a
+    ## numerically symmetric matrix; eigen(symmetric=TRUE) would silently use
+    ## only the lower triangle in that case, distorting the result, so
+    ## symmetrize explicitly first
+    h_kappa <- (h_kappa + t(h_kappa)) / 2
+    eig_h_kappa <- eigen(h_kappa, symmetric = TRUE)
+    ## diag(x) for a length-1 numeric x builds an x-by-x identity matrix
+    ## rather than a 1x1 matrix containing x -- an easy footgun when there's
```
mebrooks, line 368: No footguns :)

PR #1316, opened 2026-08-09
vcov() fails for REML fits with no free fixed effects

jmgirard:
A `REML=TRUE` fit with nothing left to integrate out returns an object whose `vcov()` errors:

```r
d <- data.frame(y = rnorm(50))
m <- glmmTMB(y ~ 0, data = d, REML = TRUE)
vcov(m)
## Error in dimnames(Q) <- list(...) : attempt to set an attribute on NULL
```

`summary()`, `confint()` and `car::Anova()` go the same way, since they all route through `vcov()`.

The trigger is a model with no random effects *and* no free fixed effects: `randomArg` gets `"beta"` with zero free elements, so nothing is random, `sdreport()` returns no `jointPrecision`, and the REML branch of `vcov.glmmTMB` does `dimnames(Q) <- ...` on a `NULL`. It also happens when the whole `beta` vector is fixed via `map`.

Such a fit is numerically the ML fit (identical objective and parameter vector), so the fix is to take the ordinary covariance path when there is no joint precision matrix rather than special-casing it.

Tests cover both routes into the state (`y ~ 0` and a fully `map`-fixed `beta`), and assert that a genuine REML fit is unaffected, i.e. still has `beta` in `par.random`. Full suite passes.

I hit this via the ordinal family in #1298, where it is reachable from an ordinary null model because that family always map-fixes its intercept. Splitting it out since it is general and stands alone; #1298 carries a copy that will drop out once this merges.

### DIFF
--- a/glmmTMB/R/methods.R
+++ b/glmmTMB/R/methods.R
@@ -420,7 +420,12 @@ vcov.glmmTMB <- function(object, full = FALSE, include_nonest = TRUE,
     warning("Calculating sdreport. Use se=TRUE in glmmTMB to avoid repetitive calculation of sdreport")
     sdr <- sdreport(object$obj, getJointPrecision=REML)
   }
-  if (REML) {
+  ## a REML fit with no free fixed effects to integrate out (e.g. y ~ 0, or
+  ## any model whose whole 'beta' vector is map-fixed) leaves sdreport()
+  ## without a joint precision matrix; such a fit is numerically identical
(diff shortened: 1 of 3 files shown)


jmgirard:
For completeness: the same NULL-`jointPrecision` pattern also breaks `glmmTMBControl(profile = TRUE)` at R/glmmTMB.R:2001, for ML as well as REML. Filed separately as #1317 rather than folded in here, since the right behaviour there needs a decision about what `h` should be when there is nothing to profile.

Issue #1039, opened 2024-05-28
calculation of df.resid


xyzyc:
I'm notice that the df.resid is not correct when option `REML = TRUE`.
i've made up a very simple example. 

```
## RCBD to check glmmTMB df.redis

dat <- data.frame(
         trt = factor(c(rep("A", 16), rep("B", 16), rep("C", 16))),
         block = factor(rep(c(rep(1, 4), rep(2, 4), rep(3, 4), rep(4, 4)),3)),
         rep = factor(rep(c(1,2,3,4), 12))
)
mu <- c(rep(0, 16), rep(1, 16), rep(-1, 16))
b <- rnorm(4, 0, 0.1)
y <- mu + rep(c(rep(b[1], 4), rep(b[2], 4), rep(b[3], 4), rep(b[4], 4)),3) + rnorm(48, 0, sd = 0.1)

dat$y <- y

ml <- lme4::lmer(y ~ trt + (1|block) , data = dat)
df.residual(ml) 
## this is 43, as we estimate 3 fixed effect and 2 variance of random effects

glmm1 <- glmmTMB(y ~ trt + (1|block), data = dat)
df.residual(glmm1)
## this is 43

glmm1$fit$par # 3 beta, betad and theta
glmm1$fit$parfull # 3 beta, betad and theta


glmm2 <- glmmTMB(y ~ trt + (1|block), data = dat, REML = TRUE)
df.residual(glmm2)
## 46, not correct

glmm2$fit$par ## this one only inlcudes betad and theta
glmm2$fit$parfull ## 3 beta, 4 b, betad and theta
```

I'm not so clear about the meaning of `betad` and `theta`. I kind of understands `beta ` is for fixed effect, `betad` is for dispersion, and `theta` is for random effect. I don't know what `b` is. 

My question is:
If I want to calculate the residual degree of freedom correctly for REML, should I count all the `beta`, `betad` and `theta` and ignore all the `b`?

bbolker:
This is a good point. I think the short answer to your last question is "yes".  **However**, it's also worth considering what you're actually using `df.residual()` for, as a lot of possible applications (e.g. computing a finite-size corrected AIC [AICc], or estimating overdispersion ...) are using `df.residual` as at best a crude approximation &mdash; once you leave the world of strictly linear models, counting each observation as 1df and each parameter as 1 df may only be a crude guess of the amount of residual information ...

At present `glmmTMB:::df.residual.glmmTMB` is just defined as

```r
[...]

Issue #893, opened 2023-02-22
Degrees of freedom used by emmeans after fitting a model in glmmTMB


davidatkaritane:
I recently noticed that emmeans uses the residual degrees of freedom when calculating a confidence interval after fitting a glm in glmmTMB, even when there is no dispersion parameter (and hence only fixed effects).

This differs from standard practice, where the degress of freedom would be set to Inf, as happens when emmeans is used after fitting such a model with glm(.).

I can understand why one might want to do this (a bit more conservatism in the interval, to allow for the fact that df=Inf is justified by asymptotics), but I thought I should check what the glmmTMB team had in mind when they chose this option.

I contacted Russell Lenth, who maintains emmeans, and he suggested I discuss it with the glmmTMB team.

bbolker:
Can you provide a reproducible example? The behaviour you're describing isn't immediately obvious, at least with the current development/master branch on GitHub.

```
library(glmmTMB)
library(emmeans)
salamander1 <- up2date(readRDS(system.file("example_files","salamander1.rds",package="glmmTMB")))
confint(emmeans(salamander1, ~ mined))
```
```
 mined emmean     SE  df asymp.LCL asymp.UCL
 yes   0.0879 0.2329 Inf    -0.369     0.544
 no    1.2298 0.0909 Inf     1.052     1.408
```
```
> packageVersion("emmeans")
[1] ‘1.8.4.1’
> packageVersion("glmmTMB")
[1] ‘1.1.5.9000’
> 
```

davidatkaritane:
My apologies - I should have done that first time...

```
library(emmeans)
library(glmmTMB)

n<-10
a<-1
b<-1

x<-1:n
mu<-exp(a+b*x)
y<-rpois(n,mu)
d<-data.frame(y=y,x=x)

m1<-glm(y~x,family=poisson,data=d)
emmeans(m1,~x)
```

```
   x emmean      SE  df asymp.LCL asymp.UCL
 5.5    6.5 0.01369 Inf     6.473     6.526

Results are given on the log (not the response) scale. 
Confidence level used: 0.95 
```

Thus df=Inf after using glm

```
m2<-glmmTMB(y~x,family=poisson,data=d)
emmeans(m2,~x)
```

```
   x emmean      SE df lower.CL upper.CL
 5.5    6.5 0.01369  8    6.468    6.531

Results are given on the log (not the response) scale. 
Confidence level used: 0.95 
```
[...]

PR #1327, opened 2026-09-01
emmeans: fix ddf handling for models without random effects

SchmidtPaul:
**This is the leftover of a case you already decided.** In #893 someone reported
that `emmeans()` used residual degrees of freedom for a glmmTMB Poisson fit
without random effects, where `glm()` reports `Inf`. Their reason was "even
when there is no dispersion parameter (and hence only fixed effects)". That was
resolved by short-circuiting every non-Gaussian family to `"asymptotic"`. A
Gaussian fit whose dispersion parameter is held fixed via `map` is the same
situation, since no variance parameter is estimated there either, but the family
test cannot see it: that takes looking at the parameter vector, which is what the
new `estDisp()` does. On master `emmeans()` reports `t` for such a fit while
`summary()` reports `z` on the same model. So the principle is not new here, only
the last case that fell through it.

**Scope.** Four `ddf` defects in the no-random-effects branch of
`emm_basis.glmmTMB()`, all Gaussian-only except the last. Throughout, *master*
means the behaviour as it stands today (`20232272`) and *this PR* the patched
behaviour:

* The fit above is the only one whose **default** output changes.
* Two more defects bite on master only when `ddf` is passed **explicitly**, on
  ordinary fits with an estimated dispersion: the request is either ignored
  without warning, or it errors out inside `dof_satt()`.
* The fourth is the missing `match.arg()`, and repairing it is **the only change
  in this PR with a footprint outside this corner**: an unrecognized `ddf`
  becomes an error for every model and component, and partial matches start
  working. Details in [What else changes](#what-else-changes).

## How I ran into it

I hit this in a meta-analysis, where the observation variances are known a priori
(`weights = 1/vi`, dispersion pinned via `map`), while cross-checking the glmmTMB
fit against `metafor::rma()`, which fits the same thing:

```r
library(glmmTMB); library(emmeans); library(metafor)
set.seed(42)
d <- expand.grid(geno = factor(c("A", "B", "C")), block = 1:4)
d$vi <- round(runif(nrow(d), 0.02, 0.30), 3)
d$yi <- rnorm(nrow(d), c(A = 4, B = 4.6, C = 5.1)[d$geno], sqrt(d$vi))

m_tmb <- glmmTMB(yi ~ geno, weights = 1/vi, data = d,
                 start = list(betadisp = 0), map = list(betadisp = factor(NA)))
m_rma <- rma(yi, vi, mods = ~ geno, data = d, method = "FE")

max(abs(fixef(m_tmb)$cond - coef(m_rma)))         #> 6.2e-08
max(abs(sqrt(diag(vcov(m_tmb)$cond)) - m_rma$se)) #> 5.5e-14   same SEs to every digit
```

The two fits agree to machine precision, so any difference downstream is a
reference-distribution difference and nothing else. `rma()` reports `z`, and so
does `summary()` (on master and unchanged by this PR):

```r
m_rma$test                          #> "z"
summary(m_tmb)$coefficients$cond
#>             Estimate Std. Error z value Pr(>|z|)
#> (Intercept)   4.0720     0.2454 16.5951   0.0000
```

`emmeans()` is the only one of the three that reports `t`, on 9 residual df:

```r
emmeans(m_tmb, ~ geno)
#> master:
#>  geno emmean    SE df lower.CL upper.CL
#>  A      4.07 0.245  9     3.52     4.63
#>  B      4.63 0.175  9     4.23     5.02
#>  C      5.34 0.198  9     4.89     5.79
#>
#> this PR:
#>  geno emmean    SE  df asymp.LCL asymp.UCL
#>  A      4.07 0.245 Inf      3.59      4.55
#>  B      4.63 0.175 Inf      4.28      4.97
#>  C      5.34 0.198 Inf      4.95      5.73

predict(m_rma, newmods = rbind(c(0, 0), c(1, 0), c(0, 1)))   ## same model, for reference
#>     pred     se  ci.lb  ci.ub
#> 1 4.0720 0.2454 3.5911 4.5529
#> 2 4.6259 0.1753 4.2823 4.9695
#> 3 5.3380 0.1981 4.9498 5.7262
```

Same estimate, same SE, different interval: on master the multiplier is
`qt(0.975, 9)` = 2.262 instead of `qnorm(0.975)` = 1.960, so every interval is 15%
too wide and the nominal 95% interval covers 97.6%. With this PR `emmeans()`
matches `summary()` and `rma()` to the printed digit.

**Why infinite df is the right answer here.** With the dispersion pinned there is
no variance parameter left to estimate. `vcov()` is then the known-weight WLS
covariance rather than an estimate carrying its own sampling variability, so
`estimate/SE` is exactly standard normal and there are no degrees of freedom to
spend. (Whether the supplied variances are themselves estimates elsewhere is a
separate question, and one that residual df do not address either, since they
scale with `n` rather than with the precision of the weights.)

## What master does today

On master, `emmeans()` reaches its own opinion instead of deferring to the shared
`check_ddf()` that `summary()`/`anova()`/`Anova()` have used since #1311. This is
the whole no-random-effects branch of `get_ddf()` in `emm_basis.glmmTMB()`, as it
stands before this PR:

```r
## master, R/emmeans.R
if (!hasRandom(object)) {
    if (fam != "gaussian") return(ddf_set("asymptotic"))
    if (trivialDisp(object)) return("df.residual")  ## don't want to warn here
    if (ddf == "kenward-roger") return(ddf_set("satterthwaite"))
    return(ddf)
}
```

Three of the four defects sit on the `trivialDisp()` line, one on the two below
it. Continuing on the data above, measured against source builds of both:

```r
m0 <- glmmTMB(yi ~ geno, weights = 1/vi, data = d)        ## dispersion estimated
d$grp <- factor(rep(1:2, 6))
mhd <- glmmTMB(yi ~ geno, dispformula = ~ grp, data = d)  ## non-trivial dispformula
dfs <- function(...) unique(summary(emmeans(...))$df)

dfs(m0, ~ geno, ddf = "asymptotic")   #> master: 8, silently  this PR: Inf
dfs(m0, ~ geno)                       #> master: 8            this PR: 8, default kept
dfs(m0, ~ geno, ddf = "asymptotick")  #> master: 8            this PR: error from match.arg
dfs(mhd, ~ geno, ddf = "satterthwaite")
#> master:  Error in seq_len(nrow(Q)) : argument must be coercible to non-negative integer
#>          (preceded by a 'length.out' warning from the same call)
#> this PR: 7   (== summary(mhd, ddf = "satterthwaite"), with the same message)
```

The four defects, all of them describing master:

1. **Residual df although no variance parameter was estimated** (the case above).
   `trivialDisp()` inspects the *structure* of `dispformula`; it cannot see the
   pin, because pinning leaves `dispformula` at `~1`.
2. **An explicit `ddf` is ignored without a warning.** The `"df.residual"` return
   deliberately bypasses `ddf_set()` so the *default* does not warn, but that
   swallows an explicit request too.
3. **An unrecognized `ddf` silently becomes residual df** in this branch, while
   models with random effects error and the `zi`/`disp` components warn.
4. **`"satterthwaite"`/`"kenward-roger"` error out** for a non-trivial
   `dispformula`. The request reaches `dof_satt()` and fails in `GMRFmarginal()`
   (`utils.R:23`), because `sdreport(getJointPrecision = TRUE)` returns no
   `jointPrecision` without random effects. `summary()` intercepts this in
   `check_ddf()` and falls back with a message, so the two interfaces disagree.
   (`dof_KR()` is never reached: `"kenward-roger"` is downgraded to
   `"satterthwaite"` one line earlier.)

## What this PR changes it to

The same branch with the patch applied, plus the two lines above it. Comments
abridged:

```r
## this PR, R/emmeans.R
## NB has to be evaluated before the match.arg() below, which would make
## missing(ddf) FALSE. A ddf from getOption("glmmTMB.df") is a default, not a
## request: summary()/anova()/Anova() don't read that option at all
ddf_explicit <- !missing(ddf)
if (!identical(ddf, "df.residual")) {
    ddf <- match.arg(ddf, c("asymptotic", "kenward-roger", "satterthwaite"))
}
## ...
if (!hasRandom(object)) {
    if (fam != "gaussian") return(ddf_set("asymptotic"))
    if (!ddf_explicit) {
        ## residual df for a plain LM-like fit, but *not* when no dispersion
        ## parameter is estimated at all
        if (trivialDisp(object) && estDisp(object)) {
            return("df.residual")  ## don't want to warn here
        }
        return("asymptotic")
    }
    ## explicit request: hand the hard cases to the same check_ddf() that
    ## summary()/anova()/Anova() use, so emmeans gives the same df and the
    ## same message instead of ignoring the request or erroring inside
    ## GMRFmarginal()
    if (ddf %in% c("asymptotic", "df.residual")) return(ddf)
    check_ddf(object, ddf)
    return("df.residual")
}
```

The patched branch splits on whether the caller asked for anything, which
separates "quietly pick a sensible default" from "honour a request or explain why
you cannot". The new `estDisp()` in `methods.R` (next to
`trivialDisp()`/`zeroDisp()`) asks the post-`map` parameter vector whether a
dispersion parameter is estimated at all; `map` pinning to a shared level rather
than to `NA` counts as estimated.

One thing I decided rather than assumed, and would happily reverse: a `ddf` found
in `getOption("glmmTMB.df")` is treated as a *default*, not as a request, so it
does not reach the explicit-request branch. Treating it as a request was my first
version and it is worse: `options(glmmTMB.df = "asymptotic")`, i.e. setting the
option to the value the signature already documents as its default, would flip
every plain Gaussian fixed-effects fit from `n - p` to `Inf`, and a standing
`"kenward-roger"` would print the "no random effects in model" message on every
`emmeans()` call. Since `summary()`/`anova()`/`Anova()` don't read the option at
all, either behaviour would be a new divergence between the entry points. As
submitted, the option behaves on master and with this PR alike.

## What else changes

Only the `match.arg()` reaches beyond Gaussian no-RE models, since it runs before
the component and family branches:

* With this PR an unrecognized `ddf` is an error everywhere. On master it depends
  on where you are: an error for models with random effects (different wording,
  from `check_ddf()`), a warning plus `asymptotic` for the `zi`/`disp`
  components, and silence plus residual df in the branch this PR is about.
* With this PR partial matches such as `ddf = "satt"` work for models with random
  effects, where master errors. That is convergence rather than collateral
  damage: `summary(m, ddf = "satt")` works on master already, because
  `summary.glmmTMB()` has had `match.arg()` all along.

For the pinned-dispersion fit, the switch to infinite df changes the *shape* of
some emmeans output, not just the numbers. Compared with master: confidence
limits are named `asymp.LCL`/`asymp.UCL` instead of `lower.CL`/`upper.CL` (so
`s$lower.CL` returns `NULL` rather than erroring), tests report `z.ratio` instead
of `t.ratio`, and `joint_tests()` gains a `Chisq` column next to `F.ratio` and
reports `df2 = Inf`. That is inherent to asymptotic inference, but worth stating.

Also worth stating: nobody is stuck on master. `emmeans(m_tmb, ~ geno, df = Inf)`
gives the right answer there today. This PR is about the default, not about
reachability.

## Left alone on purpose

* With this PR, a pinned model asked explicitly for `ddf = "satterthwaite"` still
  gets residual df, as on master, so `emmeans()` and `summary()` give the same
  answer to the same request even though the new `Inf` default argues the other
  way. Making both report `Inf` means teaching
  `check_ddf()`/`summary.glmmTMB()` about `estDisp()`, which changes `summary()`;
  happy to follow up separately.
* `ddf = "df.residual"` is exempt from the new `match.arg()`, so it keeps working
  exactly as on master. It is what `get_ddf()` returns internally and this method
  has always accepted it, but the other entry points reject it, so adding it to
  the advertised choices would re-open the divergence you closed in `a419d0c8`.
  Keeping it working is the smallest-footprint option, not necessarily the right
  one: say the word and I will drop the carve-out and let `match.arg()` be the
  single source of truth.
* The Rd now states that the residual df here count the dispersion parameter, so
  they are one lower than `lm()` reports. That only describes what
  `df.residual.glmmTMB()` does today; whether it is the right count is open in
  #1039, where you agreed `beta` should be included and pointed at the `TODO`
  above the function. I did not want to touch the counting rule in a PR about
  which distribution to use, but say the word if you would rather have the Rd
  stay silent about it.
* Two disagreements with `summary()` that exist on master and that this PR leaves
  exactly as they are, rather than changing them under cover of a bug fix. For a
  plain `glmmTMB(y ~ x)`, `summary()` reports `z` while `emmeans()` reports `t`
  on `n - p`, which is the commonest no-RE case of all. And for a non-Gaussian
  model without random effects, an explicit `"satterthwaite"` gives `summary()`
  residual df with a message but `emmeans()` `Inf` with a warning, because the
  non-Gaussian branch short-circuits one line before the one this PR touches.
  Both are worth fixing; both change more output than this PR does, so I would
  rather do them separately if you want them.

## Tests, NEWS, docs

Tests are in `test-ddf.R`; those covering changed behaviour fail on master, and
the blocks pinning down behaviour that must *not* change are labelled as such.
They check the confidence-interval multiplier (`qnorm(0.975)` versus
`qt(0.975, df)`) against degrees of freedom derived from the design rather than
read back from `df.residual()`, so the oracle is independent of the code under
test. `NEWS.Rd` and the `ddf` section of `?downstream_methods` are updated; the
behaviour-changing items are under USER-VISIBLE CHANGES. Happy to add the GH
reference to the NEWS items once this has a number.

Every `master:`/`this PR:` figure above was measured against two source builds,
one of `20232272` and one with this patch applied.

*This contribution was prepared with the assistance of an AI coding tool.*

### DIFF
--- a/glmmTMB/R/emmeans.R
+++ b/glmmTMB/R/emmeans.R
@@ -45,6 +45,26 @@
 ##' \code{binomial} or \code{poisson} that lack one. For families other than \code{gaussian},
 ##' \code{"kenward-roger"} and \code{"satterthwaite"} are allowed but emit a warning, because
 ##' their performance (and theoretical justification) for GLMMs is poorly understood.
+##'
+##' For Gaussian models \emph{without} random effects, \code{emmeans()} defaults
+##' to the residual degrees of freedom for a plain fit, i.e. one with
+##' \code{dispformula = ~1} and an estimated dispersion parameter. Any other such
+##' fit defaults to \code{"asymptotic"} (infinite df): a non-trivial
+##' \code{dispformula}, \code{dispformula = ~0}, or a dispersion parameter held
+##' fixed via the \code{map} argument to \code{\link{glmmTMB}}. In the last case
+##' there is no variance parameter left to estimate, so the residual variance is
+##' known and the Wald statistics are exactly standard normal. (As elsewhere in
+##' \pkg{glmmTMB}, the residual degrees of freedom count the dispersion
+##' parameter, so they are one lower than \code{lm()} reports for the same
+##' fixed-effect model.) Those are defaults, as is a value taken from
+##' \code{getOption("glmmTMB.df")}; a \code{ddf} passed in the call itself is
+##' respected where possible. \code{"kenward-roger"} and \code{"satterthwaite"}
+##' need random effects, so for models without them they fall back to the
+##' residual degrees of freedom with a message, exactly as in
+##' \code{\link{summary.glmmTMB}}. That fallback also applies to a model whose
+##' dispersion parameter is fixed, where it overrides the infinite-df default
+##' described above, so that \code{emmeans()} and \code{summary()} give the same
+##' answer to the same request.
 ##' @param mod a glmmTMB model
 ##' @param component which component of the model to test/analyze ("cond", "zi", or "disp")
 ##'     or, in \pkg{emmeans} only, "response" or "cmean" as described in Details.
@@ -117,6 +137,23 @@ emm_basis.glmmTMB <- function (object, trms, xlev, grid, component = c("cond", "
     ## 1. no random effects
     fam <- family(object)$family
 
+    ## did the caller actually ask for a particular ddf in this call, or are we
+    ## falling back on the default? only the latter may be silently overridden
+    ## below. A ddf coming from getOption("glmmTMB.df") is a default, not a
+    ## request: summary()/anova()/Anova() don't read that option at all, so
+    ## treating it as a request here would make emmeans() disagree with them
+    ## whenever it is set. NB has to be evaluated before the match.arg() below,
+    ## which would make missing(ddf) FALSE
+    ddf_explicit <- !missing(ddf)
+    ## same choices as summary()/anova()/Anova(); without this an unrecognized
+    ## string silently ended up as residual df. "df.residual" is what get_ddf()
+    ## returns internally and has always been accepted here too, so keep it
+    ## working, but don't offer it as a choice: the other entry points don't
+    ## take it
+    if (!identical(ddf, "df.residual")) {
+        ddf <- match.arg(ddf, c("asymptotic", "kenward-roger", "satterthwaite"))
+    }
+
     ddf_set <- function(used, requested = ddf) {
         if (requested != used) {
             warning(gettextf("ddf '%s' specified, using ddf '%s' instead", requested, used))
@@ -129,9 +166,30 @@ emm_basis.glmmTMB <- function (object, trms, xlev, grid, component = c("cond", "
 
         if (!hasRandom(object)) {
             if (fam != "gaussian") return(ddf_set("asymptotic"))
-            if (trivialDisp(object)) return("df.residual")  ## don't want to warn here
-            if (ddf == "kenward-roger") return(ddf_set("satterthwaite"))
-            return(ddf)
+            if (!ddf_explicit) {
+                ## default: residual df for a plain LM-like fit (nobs - npar,
+                ## which counts the dispersion parameter, so one fewer than
+                ## lm() would report) -- but *not* when no dispersion
+                ## parameter is estimated at all (e.g. pinned via 'map'). The
+                ## residual variance is known then, the Wald statistics are
+                ## exactly normal, and residual df would only make the
+                ## intervals spuriously wide
+                if (trivialDisp(object) && estDisp(object)) {
+                    return("df.residual")  ## don't want to warn here
+                }
+                return("asymptotic")
+            }
+            ## an explicit request is honoured where it can be, and otherwise
+            ## downgraded through the same check_ddf() that
+            ## summary()/anova()/Anova() use, so emmeans gives the same
+            ## message and the same df instead of silently ignoring the
+            ## request (trivial dispformula) or erroring inside
+            ## GMRFmarginal() on its way through dof_satt() (non-trivial
+            ## one), which needs the joint precision matrix of a model with
+            ## random effects
+            if (ddf %in% c("asymptotic", "df.residual")) return(ddf)
+            check_ddf(object, ddf)
+            return("df.residual")
         }
 
         ## hard error (not a silent downgrade) for KR + non-REML, matching
--- a/glmmTMB/R/methods.R
+++ b/glmmTMB/R/methods.R
@@ -82,6 +82,19 @@ zeroDisp <- function(object) {
     formComp(object, "dispformula", ~0)
 }
 
+## TRUE if a dispersion parameter is actually estimated: 'betadisp' survives
+## into the fitted (post-'map') parameter vector. This is about the parameters,
+## whereas trivialDisp()/zeroDisp() are about the structure of dispformula;
+## 'map' pinning to a shared level still estimates one parameter, so it's TRUE
+estDisp <- function(object) {
+    pnames <- names(object$obj$env$par)
+    ## unusable or pre-'betadisp' parameter vector (see up2date()): assume it
+    ## was estimated, the historical assumption, rather than guessing otherwise
+    if (length(pnames) == 0 ||
+        !"betadisp" %in% names(object$obj$env$parameters)) return(TRUE)
+    "betadisp" %in% pnames
+}
+
 noZI <- function(object) {
   formComp(object, "ziformula", ~0)
 }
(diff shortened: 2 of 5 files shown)


bbolker:
This is a good PR but the issues are complex &mdash; I'm going to have to find a small chunk of time to think over it calmly (maybe with a flip chart)

image

SchmidtPaul:
Yeah sorry. Maybe this helps a little?

image

bbolker:
I'm probably going to go ahead and accept this as "definitely better than what we have now".  I'm happy to go back and see this short-circuit:

```r
if (fam != "gaussian") return(ddf_set("asymptotic"))
```
because I'm worried about what to do with non-Gaussian GL[M]Ms with estimated scale or dispersion parameters (negative binomial, Gamma, beta, ...). At some point we'll have to come back to this ...

PR #1334, opened 2026-09-15
error early when profile=TRUE has no free fixed effects; make profile=TRUE work with REML=TRUE

### LINKED ISSUE #1317: profile=TRUE errors with 'attempt to set an attribute on NULL' when there are no free fixed effects
`glmmTMBControl(profile = TRUE)` errors for any model with no free fixed effects:

```r
library(glmmTMB)
set.seed(101)
d <- data.frame(y = rnorm(50))

glmmTMB(y ~ 0, data = d, control = glmmTMBControl(profile = TRUE))
## Error in dimnames(Q) <- list(parnames, parnames) :
##   attempt to set an attribute on NULL
```

Same for a model whose whole `beta` vector is fixed via `map`:

```r
d2 <- data.frame(y = rnorm(50), x = rnorm(50))
glmmTMB(y ~ x, data = d2, control = glmmTMBControl(profile = TRUE),
        start = list(beta = c(0, 0)), map = list(beta = factor(c(NA, NA))))
## same error
```

It affects ML and `REML = TRUE` equally. Ordinary models with `profile = TRUE` are fine.

Cause: with `profile = "beta"` and no free `beta` elements there is nothing random left, so `sdreport()` returns no `jointPrecision`, and `fitTMB()` at R/glmmTMB.R:2001 does `dimnames(Q) <- list(parnames, parnames)` on a `NULL`.

Reproduced on current master (4391170f), R 4.6.1, TMB 1.9.23.

Related: #1316 fixes the same NULL-`jointPrecision` pattern in `vcov.glmmTMB`, but deliberately does not touch this one, since it isn't clear what `h` (the Hessian of all non-random parameters, used just below) should be when there is nothing to profile. Happy to submit a patch if you have a preference for the right behaviour here: erroring early with a clear message, silently disabling `profile` for such models, or something else.

I ran into this while auditing the other `jointPrecision` call sites for #1316. The two remaining sites look fine: the one in `coef.glmmTMB` is unreachable (after an unconditional `stop()`), and the Kenward-Roger path in `denom_df.R` tested clean.

jmgirard:
`glmmTMBControl(profile = TRUE)` failed with "attempt to set an attribute on NULL" for any model with no free fixed-effect parameter (`y ~ 0`, or `beta` fully fixed via `map`), because `sdreport()` returns no `jointPrecision` when nothing is profiled. As discussed on #1317, this PR stops early instead: the guard sits at the top of the profile branch of `fitTMB()`, before the first `MakeADFun()`, and the message names `profile`, the two user-level causes, and the intercept-only `ordinal` case (whose intercept glmmTMB fixes internally).

While adding the tests I found that `profile = TRUE` with `REML = TRUE` failed for every model: `whichNotRandom` counted `beta` as non-random, so the marginal Hessian `h` was too large for the gradient in the Newton step (`solve.default(h, g)`: 'b' (1 x 1) must be compatible with 'a' (3 x 3)). Changes:

- `whichNotRandom` excludes every name in `TMBStruc$randomArg` instead of the hard-coded `c("b", "bzi", "bdisp")`.
- the final `sdreport()` on the profile branch requests `getJointPrecision = TMBStruc$REML`, as the non-profile branch already does, so `vcov(., full = TRUE)` has no `NA` under REML.
- the profile `MakeADFun()` passes `random = setdiff(randomArg, "beta")`: TMB adds the profiled parameters to `random` itself and printed "Duplicates in 'random' - will be removed" under REML.
- the Newton refinement is skipped when the rebuilt objective has no parameters (a Poisson model with no random effects under REML integrates `beta` out and estimates nothing else), which previously produced a spurious `max()` warning.

Under REML, profile and non-profile fits agree for `y ~ x` (Gaussian), `count ~ mined + (1|site)` and `count ~ mined * spp + (1|site), zi = ~ (1|spp)` (Salamanders, Poisson): max |estimate difference| 0 / 1.5e-8 / 7.0e-6, max |SE difference| 1.4e-10 / 3.8e-4 / 1.1e-4, |logLik difference| below 3e-9.

Tests in `test-control.R` cover the four error cases (ML and REML, `~ 0` and fully mapped `beta`), a partially mapped `beta` under ML, a fully mapped `betazi` that must not trip the guard, the three REML comparisons, and the empty-parameter REML case. The `profile` roxygen entry documents the requirement, and NEWS has two items.

Two small pre-existing things I noticed but left alone, in case you want them: `is.null(mapArg$beta)` in the ordinal intercept-mapping block partial-matches a user `map` that names only `betazi`, and the profile branch of `fitTMB()` does not honor `doOptim = FALSE`.

Closes #1317


bbolker:
this is a little tricky; in order to make sure we're doing the right thing we need to know exactly what we intend `whichNotRandom` to mean, because there are two potential meanings:

* which parameters are *actually* not random, i.e. don't use `random=TRUE` on the (R)TMB backend?
* which parameters should be treated as non-random for the purposes of summary outputs etc.? (i.e., we typically want REML output to look approximately the same as if the parameters were estimated with REML=FALSE ...)

jmgirard:
I may well be missing something here, but my understanding is that the two meanings show up at two different places, and I tried to keep them separate in the PR.

In `fitTMB()`, as far as I can tell, `whichNotRandom` is used to build the Hessian that is handed to the rebuilt objective as `hessian.fixed`. Since that objective only has parameters that are not `random`, it seems the vector has to mean "actually not random" there (`beta` included under REML). That size mismatch looks like the reason REML + profile was failing before, but I would be glad to hear if there is a better way to handle it.

The "what users should see" meaning seems to live in `vcov.glmmTMB()`, which keeps `beta` and marginalizes only the random-effect blocks, so REML output looks like ML output. The PR's other change (asking the final `sdreport()` for the joint precision under REML, as the non-profile path already does) is what lets REML profile fits go through `vcov()` by that same path. In the tests, REML profile fits match non-profile REML fits to within 4e-4 on fixed-effect SEs and 7e-6 on estimates, which made me think this was working as intended, but I am not certain I have the whole picture.

If you think the intent should be spelled out, I am happy to add a comment at the `fitTMB()` site, or to change the approach if you had something different in mind.

bbolker:
Sorry this keeps being slow (and I hope we can finish this in time to fold it in with an imminent release):

* I agree that we should be careful about whether we really mean to match *all* fixed effects (conditional/Z-I/dispersion), in which case we want `$beta` (e.g. meaning "actually not random" as in the comment above), or only conditional fixed effects (`[["beta"]]`, to avoid partial matching).  I would change this but don't want to be in too much of a hurry. It would probably be best not to rely on partial matching at all, i.e. to use `whichNotRandom` or something like it ...

I can see, illustrating your point above, that `whichNotRandom` is defined differently for the purposes of profiling vs constructing `vcov`:

```r
glmmTMB.R:        whichNotRandom <- which( ! parnames %in% TMBStruc$randomArg )
glmmTMB.R:        Qm <- GMRFmarginal(Q, whichNotRandom)
methods.R:      whichNotRandom <- which( !rownames(Q)  %in% c("b", "bzi", "bdisp") )
methods.R:      Qm <- GMRFmarginal(Q, whichNotRandom)
```
 
Looking over the code some more, I think it would be more transparent if we just had

```r
whichNotRandom <- function(namevec, include_beta = FALSE) {
    random_args <- c("b", "bzi", "bdisp")
    if (include_beta) random_args <- c(random_args, "beta")
    which(!namevec %in% random_args)
}
```

which we used in both places, with the appropriate flag. (This is slightly redundant with the construction of `randomArgs` in `mkTMBStruc`, but I think it's worth it -- we should include a comment referencing `mkTMBStruc/randomArgs`)

jmgirard:
Done as you proposed: `whichNotRandom(namevec, include_beta = FALSE)` now lives in `R/utils.R` with a comment pointing at mkTMBStruc/randomArg, the profile branch calls it with `include_beta = REML`, and `vcov.glmmTMB()` calls it with the default so REML fixed-effect output is unchanged. I kept `[["beta"]]` in the guard, since `profile = "beta"` only profiles the conditional fixed effects, and added a comment saying so. The branch is rebased on master (the emmeans ddf and ordinal PRs), full suite and R CMD check clean locally.

PR #1335, opened 2026-09-16
Rtmb port

bbolker:
This is complete except for a NEWS entry (which might have to be fairly substantial ...)


bbolker:
Proposed NEWS entry: @kaskr, @mebrooks  does this look OK?

>   \item the package now includes an experimental (for now)
\pkg{RTMB} back-end, disabled by default,
      which can be turned on for a specific model fit
      by \code{control = glmmTMBControl(use_rtmb = TRUE)}, or
      persistently via \code{useRTMB(TRUE)}. Runs using the RTMB
      back-end should match TMB back-end results up to floating point
      error, and should be as fast as old-style TMB fits.
      Going forward we expect to switch development to the RTMB
      back-end, which allows more flexibility without a performance
      cost; the TMB back-end will be maintained for at least the
      next few releases, although new features may be RTMB-only.

mebrooks:
Yes, I agree with this also (I see kaskr gave a thumbs up).

mebrooks:
After running `make all`, my enum.R has a problem. 
```
## Auto generated - do not edit by hand
.valid_link <- c(
  log                 = 0,
  logit               = 1,
  probit              = 2,
  inverse             = 3,
  cloglog             = 4,
  identity            = 5,
  sqrt                = 6,
  lambertW            = 7
  if (combinom_disp == 1) phi = etadisp;
          if (combinom_disp == 0) {
)
```
I think this line of the Makefile 
`grep "_link[ ]*=" $(PACKAGE)/src/glmmTMB.cpp | sed s/_link//g >> $@`

is interpreting `combinom_disp_link` as a new link.

bbolker:
this should be fixed now; I changed the internal name to `combinom_disp_Link` (slightly awkward; in the long run we might want to implement flexible links for the dispersion model in general, although in most cases it would be a bad idea ...)

